PR #36, opened 2022-11-25
feat: switchboard for active device tracking

ChewingGlass:


Anchor.toml
```diff
@@ -58,6 +52,24 @@ program = "./deps/solana-program-library/account-compression/target/deploy/spl_n
...
+url = "https://api.mainnet-beta.solana.com"
+
+[[test.validator.clone]]
+address = "JBu1AL4obBcCMqKBBxhpWCNUt136ijcuMZLFvTP7iWdB"
+
+
+[[test.validator.clone]]
+address = "SW1TCH7qEPTdLsDHRgPuMQjbQxKdH2aBStViMFnt64f"
+
+[[test.validator.clone]]
```
ThornM9, line 65: Add a comment on these clones saying what they're for

packages/helium-cli/src/bootstrap.ts
```diff
@@ -1,481 +0,0 @@
-import {
```
ThornM9, line 1: Why's this file being deleted?

packages/helium-cli/src/utils.ts
```diff
@@ -0,0 +1,118 @@
...
+import fetch from "node-fetch";
+
+export async function exists(
+  connection: Connection,
+  account: PublicKey
+): Promise<boolean> {
+  return Boolean(await connection.getAccountInfo(account));
+}
+
+export async function createAndMint({
```
ThornM9, line 27: This function is duplicating a lot of the logic that already exists in spl-utils

ThornM9 (changes requested):

packages/helium-cli/src/bootstrap.ts
ChewingGlass: Replaced with create-dao and create-subdao

packages/helium-cli/src/utils.ts
ChewingGlass: Not duplicating, it's using the functions from spl utils.

ThornM9 (approved):

PR #52, opened 2023-01-05
Feature/lazy tx upgrade

ChewingGlass:


packages/migration-service/src/gen-transactions.ts
```diff
@@ -12,6 +12,7 @@ import {
   registrarKey,
   PROGRAM_ID as VSR_PROGRAM_ID,
 } from "@helium/voter-stake-registry-sdk";
+import * as Collections from "typescript-collections";
```
bryzettler, line 15: No preference but https://www.npmjs.com/package/mnemonist looks more maintained?

bryzettler (approved):

ThornM9 (approved):

packages/migration-service/src/gen-transactions.ts
ChewingGlass: Oh yeah this was just already there in the merkle lib

PR #56, opened 2023-01-13
Bugfixes to enable onboarding

ChewingGlass:


packages/helium-entity-manager-sdk/src/pdas.ts
```diff
@@ -5,6 +5,16 @@ import crypto from "crypto";
 // @ts-ignore
 import bs58 from "bs58";
 
+async function hashAnyhere(input: Buffer) {
```
ThornM9, line 8: Helper function probably doesn't belong in the pdas file

ThornM9 (approved):

PR #62, opened 2023-01-18
Refactor info structs to use entity key, add end epoch scripts, update config scripts

ChewingGlass:


programs/helium-entity-manager/src/instructions/genesis_issue_hotspot_v0.rs
```diff
@@ -56,7 +56,7 @@ pub struct GenesisIssueHotspotV0<'info> {
   #[account(
     init,
     payer = lazy_signer,
-    space = 8 + std::mem::size_of::<KeyToAssetV0>(),
+    space = 8 + std::mem::size_of::<KeyToAssetV0>() + 8 * args.entity_key.len(),
```
ThornM9, line 59: Why is 8*entity_key.len() space required?

ChewingGlass: size of just gets the theoretical size if the vec is empty. Need to know how long the vec is

bryzettler (approved):

PR #65, opened 2023-01-20
Fix veHNT delegation for corner cases

ChewingGlass:


bryzettler (approved):
Would wait for @ThornM9 to eye this also before merging

bryzettler (approved):

PR #69, opened 2023-01-27
Feature/hot utils

matthewcarlreetz:
Adds various utilities for hotspots, currencies, wallets, and transactions.


packages/currency-utils/src/pyth.ts
```diff
@@ -0,0 +1,42 @@
...
+  PriceStatus,
+  PythHttpClient,
+} from "@pythnetwork/client";
+
+export const getOraclePrice = async ({
+  tokenType,
+  cluster,
+  connection,
+}: {
+  tokenType: "HNT";
```
matthewcarlreetz, line 13: Can add more tokens when they're ready.

packages/hotspot-utils/src/mplAssetAPI.ts
```diff
@@ -124,20 +124,19 @@ export async function getAssetProof(
...
   before?: string;
   after?: string;
-}
+};
 
 export async function getAssetsByOwner(
   url: string,
   wallet: string,
   {
-    sortBy = "created",
```
matthewcarlreetz, line 140: `"Created"` wasn't working. I chatted with Jon Wong from metaplex, turns out this is now an object.

packages/currency-utils/src/currencyUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+export const getBalance = async ({
+  pubKey,
+  mint,
+  connection,
+}: {
+  connection: Connection;
+  pubKey: PublicKey;
+  mint: string;
+}) => {
+  const tokenAccounts = await connection.getTokenAccountsByOwner(pubKey, {
```
ChewingGlass, line 13: This is suboptimal. You should be able to use `getAssociatedTokenAddressSync` to get the address of the token account. Then fetch and decode that particular account with AccountLayout

packages/currency-utils/src/currencyUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+  const tokenAcct = tokenAccounts.value.find((ta) => {
+    const accountData = AccountLayout.decode(ta.account.data);
+    return accountData.mint.toBase58() === mint;
+  });
+  if (!tokenAcct) return;
+
+  return Number(AccountLayout.decode(tokenAcct.account.data).amount);
+};
+
+export const getSolBalance = async ({
```
ChewingGlass, line 26: Is this needed? Method already exists on connection

packages/distributor-oracle/src/server.ts
```diff
@@ -59,19 +64,22 @@ export class DatabaseMock implements Database {
...
-    }
-  };
+    };
+  }
 
   async getCurrentRewards(assetId: PublicKey) {
     // @ts-ignore
-    const asset = await this.getAssetFn(this.issuanceProgram.provider.connection._rpcEndpoint, assetId);
+    const asset = await this.getAssetFn(
+      this.issuanceProgram.provider.connection._rpcEndpoint,
```
ChewingGlass, line 73: Your auto format messed up a ts ignore I think

packages/hotspot-utils/src/hotspotUtils.ts
```diff
@@ -0,0 +1,162 @@
...
+} from "@metaplex-foundation/mpl-bubblegum";
+import {
+  ConcurrentMerkleTreeAccount,
+  SPL_ACCOUNT_COMPRESSION_PROGRAM_ID,
+  SPL_NOOP_PROGRAM_ID,
+} from "@solana/spl-account-compression";
+
+export type SolHotspot = Awaited<ReturnType<typeof getSolHotspotInfo>>;
+
+export const createHeliumEntityManagerProgram = async ({
```
ChewingGlass, line 32: This is dangerous. The program will not be able to `rpc` transactions because `Wallet` cannot sign.

packages/hotspot-utils/src/hotspotUtils.ts
```diff
@@ -0,0 +1,162 @@
...
+  createTransferInstruction,
+} from "@metaplex-foundation/mpl-bubblegum";
+import {
+  ConcurrentMerkleTreeAccount,
+  SPL_ACCOUNT_COMPRESSION_PROGRAM_ID,
+  SPL_NOOP_PROGRAM_ID,
+} from "@solana/spl-account-compression";
+
+export type SolHotspot = Awaited<ReturnType<typeof getSolHotspotInfo>>;
+
```
ChewingGlass, line 31: We should add a function that does search assets so that we can get just the helium NFTs:

```bash
curl --request POST \
     --url https://rpc-devnet.aws.metaplex.com/ \
     --header 'Content-Type: application/json' \
     --data '{
      "jsonrpc": "2.0",
      "method": "search_assets",
      "id": "get-assets-op-1",
      "params": {
        "ownerAddress": "BcJzP2hEYgzjUwpHEtS6RhuqGfEJVx8Rq3MejujAAWrR",
        "page": 0,
        "creatorAddress": "9XtEZRnuWizL8vGoNwRX636guuSXTjQvsf3wAXmH4C1b",
        "creatorVerified": true,
        "sortBy": {"sortBy": "created", "sortDirection": "asc"}
      }
    }'

```

Where you get creatorAddress from 

```js
import { entityCreatorKey } from "@helium/helium-entity-manager-sdk";

const hntMint = new PublicKey("...hnt mint...")
const creator = entityCreatorKey(hntMint)[0]
```

ChewingGlass (changes requested):

packages/hotspot-utils/src/hotspotUtils.ts
matthewcarlreetz: @ChewingGlass All I really want to do is determine what the fees are and who's paying them when a hotspot is being updated. I know I can pull the `feePayer` from the `update-meta` transaction, but what I'm not sure about is how to determine the staking fees (e.g. If a hotspot the location doesn't change, there's no staking fee.) Can that be pulled from the transaction as well?

matthewcarlreetz: or would it be better to pull the asset from the mpl asset api and compare the new location to the previous?

packages/currency-utils/src/currencyUtils.ts
matthewcarlreetz: Definitely not needed, I'll remove

packages/hotspot-utils/src/hotspotUtils.ts
matthewcarlreetz: I removed this. Just going to simulate the transaction to determine fees.

packages/currency-utils/src/currencyUtils.ts
matthewcarlreetz: updated 🏁

packages/spl-utils/src/mplAssetAPI.ts
```diff
@@ -164,3 +163,42 @@ export async function getAssetsByOwner(
   }
 }
 
+export type SearchAssetsOpts = {
+  sortBy?: { sortBy: "created"; sortDirection: "asc" | "desc" };
+  page?: number;
+};
+
+export async function searchAssets(
```
matthewcarlreetz, line 171: is this what you were thinking @ChewingGlass ?

ChewingGlass: I would allow people to pass the params directly. Given they may be different for each call.

ChewingGlass (changes requested):

ChewingGlass (approved):

PR #81, opened 2023-02-09
Update hotspot utils deps

matthewcarlreetz:


packages/hotspot-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@helium/hotspot-utils",
-  "version": "0.0.29",
+  "version": "0.0.26",
```
ChewingGlass, line 3: This should be the same version as all other packages

ChewingGlass (changes requested):

PR #108, opened 2023-03-15
Price oracle for IOT and MOBILE

ThornM9:


packages/helium-cli/src/submit-price.ts
```diff
@@ -0,0 +1,95 @@
...
+      type: "number",
+      required: false,
+      describe: "The price to submit. Required if not using default pricing"
+    },
+    priceOracle: {
+      type: "string",
+      required: true,
+      describe: "Address of the price oracle to submit a price to"
+    },
+    useBinancePrice: {
```
abhay, line 48: ideally we add a useCoingeckoPrice as well. Current implementation: https://github.com/helium/helium-wallet-rs/blob/a7e9c0721e6ab6511b928745aed0564e1ffe4f6c/src/cmd/oracle.rs

ThornM9: Sounds good. Bilaxy too?

abhay: negative. Just Coingecko

ChewingGlass (approved):

bryzettler (approved):

PR #114, opened 2023-03-19
Create ecc signature verifiaction service and issue entity signer

ChewingGlass:


tests/helium-entity-manager.ts
```diff
@@ -262,6 +275,57 @@ describe("helium-entity-manager", () => {
       };
     });
 
+    // Only uncomment this when you want to debug the sig verifier service locally.
+    // You can run it like:
+    // env ANCHOR_WALLET=path/to//helium-program-library/tests/verifier-test.json cargo run
+    // it("properly uses the sig verifier service", async () => {
```
ThornM9, line 281: Uncomment this test but make it xit() instead of it(), it will skip the test but allow linting to keep out some code rot

bryzettler (approved):

ThornM9 (approved):

PR #126, opened 2023-03-29
add monitoring for sol balances

ThornM9:


packages/monitor-service/src/metrics.ts
```diff
@@ -15,12 +15,19 @@ export const supplyGauge = new client.Gauge({
 });
 register.registerMetric(supplyGauge);
 
-export const balanceGauge = new client.Gauge({
-  name: "solana_account_balance",
-  help: "Balance of this account",
+export const tokenBalanceGauge = new client.Gauge({
+  name: "solana_account_token_balance",
```
ChewingGlass, line 19: This will break the existing dashboard. I think calling it solana_account_balance and putting both sol and regular balances in it is fine. The label tells you what you need to know.

ChewingGlass (changes requested):

packages/monitor-service/src/metrics.ts
```diff
@@ -17,7 +17,7 @@ register.registerMetric(supplyGauge);
 
 export const balanceGauge = new client.Gauge({
   name: "solana_account_balance",
-  help: "Balance of this account",
+  help: "Balance of this token account",
```
ChewingGlass, line 20: not always token

PR #127, opened 2023-03-29
Rewards oracle wrapper contract

ThornM9:


programs/rewards-oracle/src/instructions/set_current_rewards_wrapper_v0.rs
```diff
@@ -0,0 +1,73 @@
...
+  pub entity_key: Vec<u8>,
+  pub oracle_index: u16,
+  pub current_rewards: u64,
+}
+
+#[derive(Accounts)]
+#[instruction(args: SetCurrentRewardsWrapperArgsV0)]
+pub struct SetCurrentRewardsWrapperV0<'info> {
+  #[account(mut)]
+  pub payer: Signer<'info>,
```
ChewingGlass, line 20: Isn't actually payer, I'd call this oracle.

programs/lazy-distributor/src/lib.rs
```diff
@@ -11,6 +11,15 @@ pub mod token_metadata;
 pub use instructions::*;
 pub use state::*;
 
+#[derive(Clone)]
+pub struct LazyDistributor;
+
+impl anchor_lang::Id for LazyDistributor {
```
ChewingGlass, line 17: Shouldn't be needed, anchor exports this doesn't it?

programs/rewards-oracle/src/state.rs
```diff
@@ -0,0 +1,18 @@
+use anchor_lang::prelude::*;
+
+#[account]
+#[derive(Default)]
+pub struct PriceOracleV0 {
```
ChewingGlass, line 5: This shouldn't be needed or used right?

programs/lazy-distributor/src/lib.rs
ThornM9: Anchor doesn't export it, this code block is in all our programs that are CPI'd into

programs/rewards-oracle/src/state.rs
ThornM9: Yep, updated

programs/rewards-oracle/src/instructions/set_current_rewards_wrapper_v0.rs
ThornM9: True

ChewingGlass (approved):

PR #138, opened 2023-04-06
replace pyth with price oracle contract

ThornM9:


ChewingGlass (changes requested):
just hardcode the price oracle and change only the mint data credits endpoint

The idea here is that the hardcoded oracle gets resolved by resolvers. And then when we remove that account and switch back to pyth, nobody has to upgrade their clients





[9:09](https://heliumfoundation.slack.com/archives/D048GGWGRQT/p1680790140331749)
The idea here is that the hardcoded oracle gets resolved by resolvers. And then when we remove that account and switch back to pyth, nobody has to upgrade their clients

ChewingGlass (approved):

bryzettler (approved):

PR #157, opened 2023-04-14
Fix router keys

ChewingGlass:


packages/migration-service/src/gen-transactions.ts
```diff
@@ -736,43 +743,43 @@ async function run() {
...
-            lazySigner,
-            circuitBreaker: dcCircuitBreaker,
-            dao,
-            subDao,
-          })
-          .instruction();
+        for (const router of routers) {
+          const instruction = await dcProgram.methods
+            .genesisIssueDelegatedDataCreditsV0({
+              amount: dcBal.div(new BN(routers.length)),
```
ThornM9, line 761: Is dividing evenly correct?

ThornM9 (approved):

bryzettler (approved):

packages/migration-service/src/gen-transactions.ts
ChewingGlass: Yeah, realistically this shouldn't even happen.

PR #174, opened 2023-04-19
Add endpoint to correct treasuries because of incorrect active devices

ChewingGlass:
https://explorer.solana.com/tx/5baokyfrMCyqsEHdbzUKNbyzNZY2giDGAt7YYeZXq1HhXe9uiMSHYZp65SmBPjmTZYQtVsot2PBwLYmKrtfVhcAL?cluster=devnet

^ A devnet test

PR #175, opened 2023-04-19
Remove correct treasuries

ChewingGlass:


abhay (approved):
Removed temporary endpoint added in https://github.com/helium/helium-program-library/pull/174

ThornM9 (approved):

PR #176, opened 2023-04-19
Fix router dc

ChewingGlass:


abhay (approved):
This script corrects an issue with the import of Data Credits and Router delegation to Solana. Instead of assigning the Router (an account on chain) Data Credits to the Routers directly, the script instead assigned the Owner (another account) Data Credits to the Router. This script fixes this issue on chain and will be removed after it is run. 

The impact of this is limited to just the Data Transfer infrastructure used to pay for LoRaWAN (and soon 5G) data on chain so it's important that we fix this tonight.

PR #273, opened 2023-06-01
Bulk claim rewards cli

ThornM9:


packages/helium-admin-cli/src/create-dao.ts
```diff
@@ -182,6 +183,7 @@ export async function run(args: any = process.argv) {
   const heliumSubDaosProgram = await initDao(provider);
   const heliumVsrProgram = await initVsr(provider);
 
+
```
ChewingGlass, line 186: nit: why?

packages/helium-admin-cli/src/issue-test-hotspot.ts
```diff
@@ -0,0 +1,92 @@
...
+  rewardableEntityConfigKey,
+} from "@helium/helium-entity-manager-sdk";
+import { daoKey, subDaoKey } from "@helium/helium-sub-daos-sdk";
+import { getAsset, getAssetProof, HNT_MINT, IOT_MINT, MOBILE_MINT, proofArgsAndAccounts } from "@helium/spl-utils";
+import { PublicKey } from "@solana/web3.js";
+import os from "os";
+import yargs from "yargs/yargs";
+import { exists } from "./utils";
+import { Keypair as HeliumKeypair } from "@helium/crypto";
+// @ts-ignore
```
ChewingGlass, line 16: why?

packages/helium-admin-cli/src/issue-test-hotspot.ts
```diff
@@ -0,0 +1,92 @@
...
+      alias: "k",
+      describe: "Anchor wallet keypair",
+      default: `${os.homedir()}/.config/solana/id.json`,
+    },
+    url: {
+      alias: "u",
+      default: "http://127.0.0.1:8899",
+      describe: "The solana url",
+    },
+    testingEccVerifierKeypair: {
```
ChewingGlass, line 34: testing vs test below?

utils/bulk-claim-rewards/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+sha2 = "0"
+clap = { version = "4", features = ["derive"]}
+helium-crypto = { version = "0.6.3", features = ["solana"] }
+reqwest = "0"
+helium-entity-manager = { path = "../../programs/helium-entity-manager", features = [ "no-entrypoint" ] }
+circuit-breaker = { path = "../../programs/circuit-breaker", features = [ "no-entrypoint" ] }
+lazy-distributor = { path = "../../programs/lazy-distributor", features = [ "no-entrypoint" ] }
+rewards-oracle = { path = "../../programs/rewards-oracle", features = [ "no-entrypoint" ] }
+helium-sub-daos = { path = "../../programs/helium-sub-daos", features = [ "no-entrypoint" ] }
+data-credits = { path = "../../programs/data-credits", features = [ "no-entrypoint" ] }
```
ChewingGlass, line 29: Why does this have so many dependencies? It doesn't seem to need half of these

utils/bulk-claim-rewards/src/claim_rewards.rs
```diff
@@ -0,0 +1,818 @@
...
+            .unwrap()
+            .get("json_uri")
+            .unwrap()
+            .as_str()
+            .unwrap()
+            .split("/")
+            .collect::<Vec<_>>()
+            .last()
+            .unwrap()
+            .to_string(),
```
ChewingGlass, line 220: There has got to be a better, cleaner way with serdes

ChewingGlass (changes requested):

packages/distributor-oracle/src/server.ts
```diff
@@ -424,17 +408,13 @@ export class OracleServer {
         ? await this.db.getCurrentRewardsByEntity(entityKey)
         : await this.db.getCurrentRewards(mint);
       if (proposedCurrentRewards.toNumber() > currentRewards) {
-        res.status(400).send({ error: "Invalid amount" });
-        return;
+        return {success: false, message: "Invalid amount"};
```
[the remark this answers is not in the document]
ThornM9: Yes, for a single `signTransaction` it sends a 400 with the error message, for a `signBulkTransactions` it returns the error message in place of a transaction

packages/distributor-oracle/src/server.ts
```diff
@@ -444,8 +424,44 @@ export class OracleServer {
...
+  ) {
+    if (!req.body.transactions) {
+      res.status(400).send({ error: "No transactions field" });
+      return;
+    }
+
+    let _this = this;
+    const serializedTxs = await Promise.all(req.body.transactions.map(async (txData) => {
+      const result = await _this.signTransaction(txData);
+      return result.success ? result.transaction : result.message;
```
ChewingGlass, line 442: I don't love this mechanism of handling failure. How can I tell on the client side what did succeed vs what didnt? Do I just Transaction.from and catch when it explodes?

packages/distributor-oracle/src/client.ts
```diff
@@ -115,6 +116,179 @@ export async function getPendingRewards(
...
+  assetEndpoint,
+  getAssetFn = getAsset,
+  getAssetProofFn = getAssetProof,
+}: {
+  program: Program<LazyDistributor>;
+  rewardsOracleProgram?: Program<RewardsOracle>;
+  provider: AnchorProvider;
+  rewards: BulkRewards[]; // array of bulk rewards fetched from the oracle. Maps entityKey to reward
+  dao?: PublicKey; // Optional override to the dao, defaults to HNT
+  // Must either provider assets or entityKeys. entityKeys are legacy.
```
ChewingGlass, line 141: If entityKeys are legacy, this is a new function. Remove them.

packages/distributor-oracle/src/client.ts
```diff
@@ -115,6 +116,179 @@ export async function getPendingRewards(
...
+  lazyDistributorAcc,
+  wallet = provider.wallet.publicKey,
+  skipOracleSign = false,
+  assetEndpoint,
+  getAssetFn = getAsset,
+  getAssetProofFn = getAssetProof,
+}: {
+  program: Program<LazyDistributor>;
+  rewardsOracleProgram?: Program<RewardsOracle>;
+  provider: AnchorProvider;
```
ChewingGlass, line 138: Can't you get provider from program?

ChewingGlass (changes requested):

packages/distributor-oracle/src/server.ts
ThornM9: Yes that was the original idea, so successful transactions in the batch can be executed. I've changed it to return an error from the endpoint now if any of the transactions failed to be signed.

packages/distributor-oracle/src/server.ts
```diff
@@ -437,10 +437,18 @@ export class OracleServer {
...
     let _this = this;
-    const serializedTxs = await Promise.all(req.body.transactions.map(async (txData) => {
+    const serializedTxs = await Promise.all(req.body.transactions.map(async (txData, idx) => {
       const result = await _this.signTransaction(txData);
-      return result.success ? result.transaction : result.message;
+      if (!result.success) {
+        res
+          .status(400)
+          .send({ error: result.message ? `${result.message}\n\nTransaction index: ${idx}` : `Error signing transaction index: ${idx}` });
+        return null;
```
ChewingGlass, line 446: I don't think this will do what you think it will. This return is inside the map, so the map just continues. Meanwhile you've already sent a 400. I'm not even sure what happens if you double res.send. Would prefer to not have that happen at all

ChewingGlass (changes requested):

ChewingGlass (approved):

PR #288, opened 2023-06-19
Allow updating net emission cap (fixes #286)

### LINKED ISSUE #286: HNT DAO has incorrect net emissions cap, support updating the cap on the smart contract, then change it
It's set to 34.24 HNT per epoch, but that was meant to be for when epochs were 30 minutes
The spec is 50k HNT month, so here's where 32.42 comes from:
50000 HNT / 30.42 avg days in a month / 48 epochs per day
So I guess you can just fire off an update to the DaoV0 account setting it to 1666.6666667, the correct amount

ChewingGlass:


bryzettler (approved):
LGTM

PR #313, opened 2023-07-11
Active hotspot oracle

ThornM9:


utils/active-hotspot-oracle/src/main.rs
```diff
@@ -0,0 +1,489 @@
...
+    CommitmentConfig::confirmed(),
+  );
+  let helium_entity_program = anchor_client.program(helium_entity_manager::id());
+  let tpu_client = TpuClient::new(
+    helium_entity_program.rpc().into(),
+    &helium_entity_program
+      .rpc()
+      .url()
+      .replace("https", "wss")
+      .replace("http", "ws"),
```
ChewingGlass, line 109: Dup of above?

utils/active-hotspot-oracle/src/main.rs
```diff
@@ -0,0 +1,489 @@
...
+    TpuClientConfig::default(),
+  )
+  .unwrap();
+
+  // set up delta-rs connection
+  let table_path = format!("s3://{}/{}", s3_bucket, s3_table);
+  let mut s3_config = HashMap::from([("aws_default_region".to_string(), s3_region.clone())]);
+  s3_config.insert("bucket_name".to_string(), s3_bucket);
+  s3_config.insert("aws_secret_access_key".to_string(), secret_access_key);
+  s3_config.insert("allow_http".to_string(), "true".to_string());
```
ChewingGlass, line 119: These configs should prob be different in production. Make it conditional on the secret/access keys being present? Look at how delta lake sink does this

utils/active-hotspot-oracle/src/main.rs
```diff
@@ -0,0 +1,489 @@
...
+    mark_inactive_mobile_ixs,
+    &kp,
+  )?;
+
+  // write to new checkpoint file and upload to s3
+  println!("Uploading new save files to s3");
+  last_inactive_pub_keys.extend(mark_inactive_diff.iter().cloned());
+  let active_pub_keys = HashSet::new();
+  save_checkpoint(active_pub_keys, s3_config.clone(), "active_save.json").await?;
+  save_checkpoint(
```
ChewingGlass, line 247: Is there a race condition here if saving a checkpoint fails? Now we have on-chain not matching the checkpoint of actives/inactives.

ChewingGlass (changes requested):

utils/active-hotspot-oracle/src/main.rs
ThornM9: If saving the checkpoint fails then the next time the job runs it will run from the previous checkpoint. In this application we need to guarantee a tx will run at least once, if a tx gets run twice it's better than skipping some. 

Should be using the fetched infos to check if a tx even needs to be run though.

ChewingGlass: Couldn't the following scenario break us:

  Last checkpoint says hotspot A is active
  Currently A is not active
  Update A to not active
  Save checkpoint fails
  
  Next run:
  Last checkpoint says A is active
  Currently A is active (it was turned back on again)
  A matches last status, so no change. (BUG)

ThornM9: It doesn't "break", that scenario doesn't cause any unintended consequences and I wrote it with that possibility in mind. But yes, as I said in my last comment I will change it to use the fetched info account to determine if the tx needs to be run or not to avoid wasting a tx.

ChewingGlass (approved):

utils/hpl-utils/src/token.rs
```diff
@@ -24,11 +24,11 @@ pub enum TokenError {
 )]
 #[serde(rename_all = "lowercase")]
 pub enum Token {
-  Sol,
   Hnt,
   Mobile,
   Iot,
   Dc,
+  Sol,
```
bryzettler, line 31: Moving this enum wont cause any issues?

bryzettler (approved):

PR #314, opened 2023-07-12
add endpoint to pay for mobile onboarding fees

ThornM9:


programs/helium-entity-manager/src/instructions/pay_mobile_onboarding_fee_v0.rs
```diff
@@ -0,0 +1,110 @@
...
+  pub data_credits_program: Program<'info, DataCredits>,
+  pub token_program: Program<'info, Token>,
+  pub associated_token_program: Program<'info, AssociatedToken>,
+  pub system_program: Program<'info, System>,
+  pub helium_sub_daos_program: Program<'info, HeliumSubDaos>,
+}
+
+pub fn handler(ctx: Context<PayMobileOnboardingFeeV0>) -> Result<()> {
+  let dc_fee = ctx.accounts.sub_dao.onboarding_dc_fee;
+  assert!(dc_fee == 4000000);
```
abhay, line 67: should this just be `> 0`?

ThornM9: That works too, I just preferred more well defined code path. It's not a permanent endpoint too, so it should be removed when or before the onboarding fee changes

abhay: roger. i like the specificity. thx.

programs/helium-entity-manager/src/instructions/temp_pay_mobile_onboarding_fee_v0.rs
```diff
@@ -0,0 +1,110 @@
...
+      "dc".as_bytes(),
+      dc_mint.key().as_ref()
+    ],
+    seeds::program = data_credits_program.key(),
+    bump = dc.data_credits_bump,
+    has_one = dc_mint
+  )]
+  pub dc: Account<'info, DataCreditsV0>,
+  #[account(mut)]
+  pub mobile_info: Box<Account<'info, MobileHotspotInfoV0>>,
```
ChewingGlass, line 56: Missing a seeds check to ensure that this is part of the correct rewardable entity config. Otherwise you can just make a new dao with a madeup data credits and "pay" on a real helium hotspot

programs/helium-entity-manager/src/instructions/temp_pay_mobile_onboarding_fee_v0.rs
```diff
@@ -0,0 +1,110 @@
...
+  pub token_program: Program<'info, Token>,
+  pub associated_token_program: Program<'info, AssociatedToken>,
+  pub system_program: Program<'info, System>,
+  pub helium_sub_daos_program: Program<'info, HeliumSubDaos>,
+}
+
+pub fn handler(ctx: Context<TempPayMobileOnboardingFeeV0>) -> Result<()> {
+  let dc_fee = ctx.accounts.sub_dao.onboarding_dc_fee;
+  assert!(dc_fee == 4000000);
+  assert!(ctx.accounts.mobile_info.dc_onboarding_fee_paid == 0);
```
ChewingGlass, line 68: require_eq!

ChewingGlass (changes requested):

ChewingGlass (approved):

PR #315, opened 2023-07-12
Improve react hooks

ChewingGlass:


packages/helium-admin-cli/src/update-rewardable-entity-config.ts
```diff
@@ -88,19 +86,19 @@ export async function run(args: any = process.argv) {
   } else {
     settings = {
       mobileConfig: {
-        fullLocationStakingFee: toBN(0, 0),
-        dataonlyLocationStakingFee: toBN(0, 0),
+        fullLocationStakingFee: toBN(1000000, 0),
```
ThornM9, line 89: Might as well make these args with default values

ThornM9 (changes requested):

ThornM9 (approved):

PR #316, opened 2023-07-12
Upgrade helium-crypto

madninja:
Upgrade helium-crypto to 0.7.1 which adds RSA key support

Note:

* utils/vehnt does not build for other errors so was not upgraded
* utils/generate-test-gateway-txn was not upgraded because of other errors

### DIFF
--- a/utils/ecc-sig-verifier/Cargo.toml
+++ b/utils/ecc-sig-verifier/Cargo.toml
@@ -11,7 +11,7 @@ overflow-checks = true
 [dependencies]
 solana-sdk = "1.14.11"
 rocket = { version = "0.5.0-rc.2", features = ["json"] }
-helium-crypto = "0.7.1"
+helium-crypto = "0.7.3"
 hex = "0.4.3"
 serde = "1.0.157"
 bincode = "1.3.3"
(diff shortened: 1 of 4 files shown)


utils/ecc-sig-verifier/Cargo.toml
```diff
@@ -11,9 +11,9 @@ overflow-checks = true
 [dependencies]
 solana-sdk = "1.14.11"
 rocket = { version = "0.5.0-rc.2", features = ["json"] }
-helium-crypto = "0.6.8"
+helium-crypto = "0.7.1"
 hex = "0.4.3"
 serde = "1.0.157"
 bincode = "1.3.3"
-anchor-lang = { workspace = true }
+anchor-lang = { version = "0.26.0", features = ["init-if-needed"] }
```
ThornM9, line 18: `init-if-needed` feature is only for programs, not needed client side

ThornM9 (changes requested):

utils/ecc-sig-verifier/Cargo.toml
madninja: Fixed in [0e38d1f](https://github.com/helium/helium-program-library/pull/316/commits/0e38d1f6ba3521225e8444d72c1a108238c5acc3)

ChewingGlass (approved):

PR #337, opened 2023-07-25
Enable large RSA keys with keyToAsset creator on all new assets, and v1 endpoint.

ChewingGlass:


bryzettler (approved):

packages/helium-entity-manager-sdk/src/pdas.ts
```diff
@@ -112,6 +112,21 @@ export function decodeEntityKey(
   }
 }
 
+export const keyToAssetKeyRaw = (
```
ThornM9, line 115: Pointless function? Not used anywhere in this PR

packages/metadata-service/src/index.ts
```diff
@@ -1,7 +1,17 @@
 import cors from "@fastify/cors";
 // @ts-ignore
 import animalHash from "angry-purple-tiger";
+import { decodeEntityKey, init, keyToAssetKeyRaw } from "@helium/helium-entity-manager-sdk"
```
ThornM9, line 4: Unused import keyToAssetKeyRaw

programs/helium-entity-manager/src/instructions/issue_entity_v0.rs
```diff
@@ -146,16 +158,23 @@ pub fn handler(ctx: Context<IssueEntityV0>, args: IssueEntityArgsV0) -> Result<(
...
-  let asset_id = get_asset_id(
-    &ctx.accounts.merkle_tree.key(),
-    ctx.accounts.tree_authority.num_minted,
-  );
 
   let name = animal_name.to_string();
+  let mut uri = format!("{}/{}", ENTITY_METADATA_URL, key_str);
+
+  // HACK: Handle wifi hotspots with v1 uri. TODO: Find a way to do this for all hotspots after migration
+  // https://docs.google.com/document/d/1Q6VdcDA3McehQYb1MzWL8Iy421D84T7UrmJE7Z9Vc_Y/edit?usp=sharing
```
ThornM9, line 166: If the wifi hotspots have a key size of 1024 bytes and they are included in the args, then won't this instruction fail due to transaction size?

programs/rewards-oracle/src/instructions/set_current_rewards_wrapper_v1.rs
```diff
@@ -0,0 +1,77 @@
...
+  #[account(
+    mut,
+    has_one = lazy_distributor
+  )]
+  pub recipient: Box<Account<'info, RecipientV0>>,
+
+  // We assume that the oracle verified that the key to asset entity key went with these rewards.
+  // This endpoint then verifies the connection of that key_to_asset to the recipient asset.
+  #[account(
+    constraint = key_to_asset.asset == recipient.asset,
```
ThornM9, line 32: You can verify the hash of the entity key to harden against malicious oracles. i.e. pass in `entity_key_hash` in args, verify that `args.entity_key_hash == hash(key_to_asset.entity_key)`

ThornM9 (approved):

ThornM9 (changes requested):

packages/distributor-oracle/src/client.ts
```diff
@@ -225,24 +224,27 @@ export async function formBulkTransactions({
   // construct the set and distribute ixs
   let setAndDistributeIxs = await Promise.all(
     compressionAssetAccs.map(async (assetAcc, idx) => {
-      const entityKey = assetAcc.content.json_uri.split("/").slice(-1)[0];
-      const keyToAsset = keyToAssetKey(dao, entityKey, encoding)[0];
+      const keyToAssetK = assetAcc.creators[1].address
```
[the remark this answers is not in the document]
ChewingGlass: good catch

programs/rewards-oracle/src/instructions/set_current_rewards_wrapper_v1.rs
ChewingGlass: We can't really prevent malicious oracles though, they have the key to set rewards for whatever hotspot they want

programs/helium-entity-manager/src/instructions/issue_entity_v0.rs
madninja: Yeah, iirc the base58(check) encoding of a RSA key is like 338 bytes.. add the URL overhead to that I guess.. 

Example: https://github.com/helium/helium-crypto-rs/blob/main/src/rsa/mod.rs#L238

packages/helium-entity-manager-sdk/src/pdas.ts
ChewingGlass: useful to have, we're going to need it later

packages/metadata-service/src/index.ts
ChewingGlass: fixed

PR #347, opened 2023-07-29
Fix emissions with another days of subscribers, missed past multisig …

ChewingGlass:
…window


abhay (approved):
This is necessary because the multisig wasn’t able to execute in time today. Will try again tomorrow.

PR #348, opened 2023-07-31
Start halvening 1 second after midnight to apply only to the first full day

ChewingGlass:


bryzettler (approved):

abhay (approved):
This is correct because rewards are issued in arrears (for the prior period).

PR #365, opened 2023-08-11
Fix Rewards Oracle Package.json

KnightAR:
Package.json was set to use a very old version of @helium/idls, updated to make sure that the @helium/idls match the rest of the packages


ChewingGlass:
There's a ton of merge conflicts here, and large # of file changes. Not sure what happened.

bryzettler:
This should be a non issue anymore as I ran `lerna-update-wizard` and consolidated the deps across the repo.

PR #382, opened 2023-08-30
Fix subscriber metadata

Perronef5:


packages/metadata-service/src/index.ts
```diff
@@ -67,24 +70,59 @@ server.get<{ Params: { keyToAssetKey: string } }>(
...
+          daoKey(HNT_MINT)[0],
+          eccCompact,
+          "b58"
+        );
+
+        const keyToAsset = await program.account.keyToAssetV0.fetchNullable(
+          keyToAssetK
+        );
+
+        const asset = await getAsset(
```
ChewingGlass, line 95: Why is any of this needed? Should just be a passthrough right?

Perronef5: Ahh yeah good point

ChewingGlass (approved):

PR #409, opened 2023-09-19
make hnt-to-rent-service work with jupiter

bryzettler:


packages/hnt-to-rent-service/src/jupiter.ts
```diff
@@ -0,0 +1,184 @@
...
+  const borrowIx = SystemProgram.transfer({
+    fromPubkey: platformWallet.publicKey,
+    toPubkey: userWallet,
+    lamports: ataRent,
+  });
+
+  const repayIx = SystemProgram.transfer({
+    fromPubkey: userWallet,
+    toPubkey: platformWallet.publicKey,
+    lamports: ataRent,
```
ChewingGlass, line 159:
```suggestion
    lamports: ataRent + (10000),
```

^Force the user to also pay the transaction fee back into this wallet.

ChewingGlass (changes requested):

packages/hnt-to-rent-service/src/jupiter.ts
bryzettler: ahh, good call

ChewingGlass (approved):

PR #434, opened 2023-09-29
Add reverse geocoding to account postgres sink

ChewingGlass:


packages/account-postgres-sink-service/program_account_configs_example.json
bryzettler, line 1: would probably just revert this and add the plugin config to the existing example.json

bryzettler (approved):

bryzettler (changes requested):
Needs some code added to `handleAccountWebhook` that ties into the plugin

packages/account-postgres-sink-service/src/plugins/extractHexLocation.ts
```diff
@@ -0,0 +1,154 @@
...
+
+      return {
+        ...account,
+        city: null,
+        state: null,
+        country: null,
+        streetAddress: null,
+        lat: null,
+        long: null,
+        ..._omit(reverseGeod?.dataValues || {}, ["createdAt", "updatedAt"]),
```
bryzettler, line 139: I think "updatedAt" => "refreshed_at" its ugly I know

bryzettler (approved):

PR #437, opened 2023-10-02
Update bulk send to support versioned txns

Perronef5:


ChewingGlass:
Closing because this would require oracle support as well

PR #446, opened 2023-10-13
Repair marker sizes and add close expired vote

ChewingGlass:


bryzettler (approved):

packages/crons/package.json
```diff
@@ -38,15 +38,19 @@
     "@helium/helium-sub-daos-sdk": "^0.4.0",
     "@helium/lazy-distributor-sdk": "^0.4.0",
     "@helium/mobile-entity-manager-sdk": "^0.4.0",
+    "@helium/organization-sdk": "^0.0.5",
```
bryzettler, line 41: Do you need to add these three to the tsconfig refrences file too?

bryzettler (approved):

PR #454, opened 2023-10-19
Update metadata-service to use read-replica RDS

deasydoesit:


deasydoesit:
Closed in lieu of #455

PR #500, opened 2023-12-05
Cloudflare

deasydoesit:
Updates `metadata-service` and `entity-invalidator` for alignment with Cloudflare

### DIFF
--- a/packages/entity-invalidator/src/index.ts
+++ b/packages/entity-invalidator/src/index.ts
@@ -1,11 +1,9 @@
 import { decodeEntityKey } from "@helium/helium-entity-manager-sdk";
-import AWS from "aws-sdk";
 import { Op } from "sequelize";
 import { IotHotspotInfo, KeyToAsset, MobileHotspotInfo } from "./model";
 // @ts-ignore
 import {
-  CLOUDFLARE_EMAIL,
-  CLOUDFLARE_API_KEY,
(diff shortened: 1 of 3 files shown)


packages/entity-invalidator/src/index.ts
```diff
@@ -149,72 +149,30 @@ function delay(ms: number) {
...
+    const response = await fetch(CLOUDFLARE_API_URL, { method: 'POST', headers: CLOUDFLARE_HEADERS, body: body });
+    const data = await response.json();
 
-function getPaths(entity: KeyToAsset): string[] {
-  const v1 = `/v1/${entity.address}`;
-  const v2 = `/v2/hotspot/${entity.address}`;
-  if ((entity.entityKeyStr?.length || 0) >= 200) {
-    return [v1, v2];
-  }
+    console.log(data);
```
bryzettler, line 169: rm?

bryzettler (approved):

packages/entity-invalidator/src/index.ts
deasydoesit: The console.log is for us track invalidation responses from the Cloudflare API but I can remove it if need be

bryzettler: All good, just though it was leftover and didn't serve a purpose

PR #562, opened 2024-01-30
ci: push IDL to Github release tag

lthiery:
When a program is tagged, the IDL will be pushed to Github.


lthiery:
Here's an example of what the tag looked like: https://github.com/lthiery/helium-program-library/releases/tag/program-helium-entity-manager-0.2.9

ChewingGlass (approved):

lthiery:
https://github.com/helium/helium-anchor-gen now watches for these json files and will make PRs to itself when it detects an update (even if the version update does not change the IDL)

PR #569, opened 2024-02-05
Add send with squads to utils

Perronef5:


ChewingGlass (changes requested):
Don't you need to also remove it from helium-admin-cli?

Perronef5:
> Don't you need to also remove it from helium-admin-cli?

Sigh

ChewingGlass:
@Perronef5 builds still failing. 

Also thinking we may want to make this its own package. Or at least put it in its own file and export explicitly by name from index.ts so it can be tree shaken. Basically every package we have imports spl-utils, so they'll all get a needless squads sdk in their bundle

PR #688, opened 2024-08-07
Feature/close expired proxies cron

bryzettler:


packages/crons/src/close-governance.ts
```diff
@@ -114,6 +128,65 @@ import pLimit from "p-limit";
...
+                  proxyAssignment: new PublicKey(proxy.publicKey),
+                })
+                .instruction();
+            })
+          )
+        ).filter(truthy);
+      })
+    );
+
+    await batchSequentialParallelInstructions({
```
ChewingGlass, line 185: Don't we have a helper to batch instructions to txs packing as tightly as possible?

ChewingGlass (changes requested):

packages/crons/src/close-governance.ts
bryzettler: does that then run them in sequential order or parallel?

ChewingGlass (approved):

PR #750, opened 2024-12-07
Fix net emissions not happening due to extra emitted HNT for HIP-138

ChewingGlass:


abhay (approved):
look at you spending all those CUs

bryzettler (approved):

PR #774, opened 2025-01-23
add temp endpoint and backfill script for mobile deployment infos

bryzettler:
#808


riobah:
Does this script fires instructions for all currently known radios, or only the missing ones? 

We're parsing the HEM instructions to keep the current state, that's why I'm trying to understand how to handle this new temporary ix.

ChewingGlass (approved):

bryzettler:
@riobah 

Updates the on chain structs for ones that have missing/wrong deployment info or ones that were onboarded with missing antenna data

packages/helium-admin-cli/src/backfill-mobile-deployment-infos.ts
```diff
@@ -0,0 +1,262 @@
...
+        }
+      })
+    )
+  ).filter(truthy);
+
+  console.log(`Total corrections needed: ${ixs.length}`);
+  if (commit) {
+    try {
+      for (const c of chunks(ixs, 250)) {
+        await batchParallelInstructionsWithPriorityFee(provider, c);
```
ChewingGlass, line 255: Instead of doing this in chunks, you can do this:

```
const transactions = await batchInstructionsToTxsWithPriorityFee(
    provider,
    instructions,
    {
      useFirstEstimateForAll: true,
    }
  );

  await bulkSendTransactions(provider, transactions, console.log);
  ```

It'll make it use waaaay less RPC since it won't simulate every tx to get its compute, and get an individual priority fee.

ChewingGlass (changes requested):

ChewingGlass (approved):

PR #804, opened 2025-01-31
Lazy distributor using ed25519 instruction instead of tx signer

ChewingGlass:


lerna.json
```diff
@@ -4,5 +4,5 @@
     "packages/*"
   ],
   "useWorkspaces": true,
-  "version": "0.9.23"
+  "version": "0.9.24-alpha.0"
```
bryzettler, line 7: ?

packages/distributor-oracle/src/server.ts
```diff
@@ -606,14 +634,176 @@ export class OracleServer {
...
+          })
+        ).instruction();
+      }
+
+      const entityKey = decodeEntityKey(
+        keyToAsset.entityKey,
+        keyToAsset.keySerialization
+      )!;
+
+      // Transfer some tokens from lazy signer to me
```
bryzettler, line 752: ?

bryzettler (approved):

lerna.json
ChewingGlass: Had to be done to build the server to use in devnet.

PR #809, opened 2025-01-31
Add temporary endpoint to fix rewards destinations

ChewingGlass:


packages/spl-utils/src/transaction.ts
```diff
@@ -472,7 +472,7 @@ export type Status = {
   currentBatchProgress: number;
   currentBatchSize: number;
 };
-const TX_BATCH_SIZE = 200;
+const TX_BATCH_SIZE = 100;
```
abhay, line 475: intentional?

abhay (approved):

packages/spl-utils/src/transaction.ts
ChewingGlass: Yeah. Helius gets mad at 200

PR #819, opened 2025-02-08
Net Emissions account for no emit

bryzettler:


packages/helium-admin-cli/src/end-epoch.ts
```diff
@@ -1,111 +1,416 @@
...
+            }
+          }
+        }
+      }
+
+      targetTs = targetTs.add(new BN(EPOCH_LENGTH));
+    }
+
+    // distribute hst
+    const hydraProgram = await initHydra(provider);
```
ChewingGlass, line 211: You shouldn't need this anymore, 0 HST

programs/helium-sub-daos/src/instructions/calculate_utility_score_v0.rs
```diff
@@ -87,16 +100,31 @@ pub fn handler(
...
+      .accounts
+      .not_emitted_counter
+      .to_account_info()
+      .data_is_empty()
+  {
+    let info: Account<NotEmittedCounterV0> = Account::try_from(&ctx.accounts.not_emitted_counter)?;
+    not_emitted = info.amount_not_emitted.saturating_sub(prev_not_emitted);
+  };
+
+  ctx.accounts.dao_epoch_info.num_not_emitted = not_emitted;
```
ChewingGlass, line 117: Shouldn't this keep track of the `total` not emitted. Right now it's only keeping track of what was not emitted last epoch. I don't think this calculation will work with taking the info.amount_not_emitted and subbing prev_not_emitted. You'd need to have prev_not_emittebe the previous total.

ChewingGlass (changes requested):

ChewingGlass (approved):

PR #942, opened 2025-04-29
filter via truthy and add guard

bryzettler:


packages/spl-utils/src/transaction.ts
```diff
@@ -828,6 +828,7 @@ export async function batchInstructionsToTxsWithPriorityFee(
 
   let firstTxComputeAndPrio: TransactionInstruction[] | null = null;
   for (const instruction of instructions) {
+    if (!instruction) continue;
```
abhay, line 831: Triggers me every time

abhay (approved):

ChewingGlass (approved):

PR #947, opened 2025-05-02
Feat/improved delegation

ChewingGlass:


abhay (approved):

packages/helium-vote-service/src/index.ts
```diff
@@ -2,6 +2,7 @@ import cors from "@fastify/cors";
 import fastifyStatic from "@fastify/static";
 import { organizationKey } from "@helium/organization-sdk";
 import { createMemoInstruction } from "@solana/spl-memo";
+import { DuneClient } from "@duneanalytics/client-sdk";
```
abhay, line 5: I wonder if it makes more sense to sync this into pg sink as we do other derived data.

ChewingGlass: We'd have to do a transaction-based sink. So the existing account sink code wouldn't work. Not worth the effort imo unless we start doing more analytics

abhay: yeah that's fair. this is a good start though.

PR #962, opened 2025-05-14
Feat/asset ownership

bryzettler:


packages/account-postgres-sink-service/src/utils/database.ts
```diff
@@ -58,6 +58,35 @@ export const database = new Sequelize({
       }
       config.password = password;
     },
+    afterConnect: async (connection: any) => {
```
bryzettler, line 61: `cursors` table is used by multiple services now and they are identified by a unique `service` column. Ie account_sink, asset_ownership. Need to update pkey from cursor to service

packages/asset-ownership-service/src/services/substream.ts
```diff
@@ -0,0 +1,537 @@
...
+  const registry = createRegistry(substream);
+  const transport = createGrpcTransport({
+    baseUrl: SUBSTREAM_URL!,
+    httpVersion: "2",
+    interceptors: [createAuthInterceptor(token)],
+    useBinaryFormat: true,
+    jsonOptions: { typeRegistry: registry },
+  });
+
+  let merkleTrees: Set<string> = new Set(
```
bryzettler, line 175: setting merkleTrees to in memory set here.

packages/asset-ownership-service/src/services/substream.ts
```diff
@@ -0,0 +1,537 @@
...
+
+                      const argMap = Object.fromEntries(
+                        (formattedInstruction.args || []).map((arg) => [
+                          arg.name,
+                          arg,
+                        ])
+                      );
+
+                      switch (decodedInstruction.name) {
+                        case "update_maker_tree_v0":
```
bryzettler, line 388: since the create_tree is done via CPI it will be nested in the `update_maker_tree_v0` and we will need to use the hem coder to decode it. This is the same for `MintToCollectionV` and `issue_entity_v0`

packages/asset-ownership-service/src/utils/database.ts
```diff
@@ -0,0 +1,146 @@
...
+        config.dialectOptions = {
+          ssl: {
+            require: false,
+            rejectUnauthorized: false,
+          },
+        };
+      }
+      config.password = password;
+    },
+    afterConnect: async (connection: any) => {
```
bryzettler, line 61: same logic for switching the pkey

packages/asset-ownership-service/src/services/substream.ts
```diff
@@ -0,0 +1,566 @@
...
+  const registry = createRegistry(substream);
+  const transport = createGrpcTransport({
+    baseUrl: SUBSTREAM_URL!,
+    httpVersion: "2",
+    interceptors: [createAuthInterceptor(token)],
+    useBinaryFormat: true,
+    jsonOptions: { typeRegistry: registry },
+  });
+
+  let makerMerkleTrees: Set<string> = new Set(
```
ChewingGlass, line 177: Missing data only merkle trees

packages/asset-ownership-service/src/services/substream.ts
```diff
@@ -0,0 +1,566 @@
...
+                    const { message } = transaction;
+                    const { addressTableLookups } = message;
+                    const addressLookupTableAccounts: AddressLookupTableAccount[] =
+                      [];
+
+                    for (const addressTableLookup of addressTableLookups) {
+                      const key = addressTableLookup.accountKey.toBase58();
+                      if (!lookupTableCache[key]) {
+                        const lookupTableResult =
+                          await provider.connection?.getAddressLookupTable(
```
ChewingGlass, line 336: Doesn't this call already happen in convertSubstreamTransaction? And it's not cached there. Though the account fetch cache kind of does it

PR #972, opened 2025-06-03
Feat/rewards by wallet endpoint

bryzettler:
adds two endpoints
`/rewards?owner=`
`/rewards?destination=`

### DIFF
--- a/packages/distributor-oracle/src/pgDatabase.ts
+++ b/packages/distributor-oracle/src/pgDatabase.ts
@@ -11,7 +11,6 @@ import { recipientKey } from "@helium/lazy-distributor-sdk";
 import {
   Asset,
   getAsset,
-  HNT_MINT,
   searchAssets,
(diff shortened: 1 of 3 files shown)


packages/distributor-oracle/src/pgDatabase.ts
```diff
@@ -253,4 +260,125 @@ export class PgDatabase implements Database {
...
+            model: KeyToAsset,
+            required: true,
+            attributes: [],
+            include: [
+              {
+                model: Recipient,
+                required: true,
+                attributes: [],
+                where: {
+                  lazyDistributor: HNT_LAZY_DISTRIBUTOR.toBase58(),
```
ChewingGlass, line 278: Isn't there already a this.lazyDistributor?

ChewingGlass (changes requested):

PR #982, opened 2025-06-11
Feat/swap distributor logic

bryzettler:


packages/distributor-oracle/src/pgDatabase.ts
```diff
@@ -111,66 +47,51 @@ export class PgDatabase implements Database {
...
+        },
+      ],
+      limit,
+      offset,
+    });
+
+    const entityKeys = keyToAssets
+      .map((kta) => kta.encodedEntityKey)
+      .filter(truthy);
+    const lifetimeRewards = await this.getBulkRewards(entityKeys);
```
ChewingGlass, line 72: Can't this all be done via one big sql query? Why pull the key to assets first?

ChewingGlass (changes requested):

PR #1057, opened 2025-09-09
Feat/atomic data publisher

bryzettler:


ChewingGlass:
`handleAccountWebhook` should be taking lastBlockHeight as an arg, not fetching it from RPC. Since the account data is from a different block than whatever the RPC is on

ChewingGlass (changes requested):
`handleAccountWebhook` should take `lastBlockHeight` as an arg, since the block height is associated with the piece of data being inserted, not with whatever block height is currently available on the rpc.

utils/atomic-data-publisher/src/main.rs
```diff
@@ -0,0 +1,145 @@
...
+  let settings = Settings::new().context("Failed to load configuration")?;
+
+  initialize_logging(&settings.logging)?;
+
+  info!("Starting Atomic Data Publisher");
+  info!("Configuration loaded successfully");
+
+  validate_config(&settings).context("Configuration validation failed")?;
+
+  let service = Arc::new(
```
ChewingGlass, line 32: Does this need to be an Arc? It's just being run here and not referenced anywhere else

utils/atomic-data-publisher/src/service.rs
```diff
@@ -0,0 +1,513 @@
...
+          return Err(AtomicDataError::NetworkError(e.to_string()));
+        }
+        Ok(())
+      })
+    };
+    handles.push(metrics_handle);
+    info!("Metrics server started on {}", metrics_bind_addr);
+
+    let polling_handle = {
+      let service = self.clone();
```
ChewingGlass, line 173: Feels a bit weird for the service to clone itself to start a polling loop. Should the background tasks be another trait/service?

utils/atomic-data-publisher/src/queries.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+  }
+
+  /// Parameters:
+  /// - $1: hotspot_type - Either 'mobile' or 'iot' to filter for specific hotspot type
+  /// - $2: last_processed_block - The last block that was already processed
+  /// - $3: max_block - The maximum block number to process (exclusive)
+  ///
+  /// Returns: job_name, solana_address, asset, block, atomic_data (JSON)
+  pub const CONSTRUCT_ATOMIC_HOTSPOTS: &'static str = r#"
```
ChewingGlass, line 52: It'd be great if queries were all stored in sql files in a directory, so you get some syntax highlighting.

utils/atomic-data-publisher/src/database.rs
```diff
@@ -0,0 +1,797 @@
...
+
+    Ok(rows)
+  }
+
+  fn process_query_results(&self, rows: &[sqlx::postgres::PgRow], job_name: &str) -> Vec<ChangeRecord> {
+    let mut changes = Vec::with_capacity(rows.len());
+
+    for row in rows {
+      let mut atomic_data_array = Vec::with_capacity(1);
+      atomic_data_array.push(row.get::<serde_json::Value, _>("atomic_data"));
```
ChewingGlass, line 399: why does this need to be an array if it's always length one? This function feels like one ChangeRecord should include all rows?

utils/atomic-data-publisher/src/metrics.rs
```diff
@@ -0,0 +1,134 @@
...
+
+  pub fn observe_database_query_duration(&self, duration: f64) {
+    self.database_query_duration.observe(duration);
+  }
+
+  pub fn observe_publish_duration(&self, duration: f64) {
+    self.publish_duration.observe(duration);
+  }
+
+  pub async fn serve_metrics(self: Arc<Self>, bind_address: &str, mut shutdown_rx: tokio::sync::watch::Receiver<bool>) -> anyhow::Result<()> {
```
michaeldjeffrey, line 64: Using `metrics-exporter-prometheus` handles this.
https://github.com/helium/oracles/blob/main/metrics/src/lib.rs#L23

bryzettler: nice, will refactor

utils/atomic-data-publisher/src/database.rs
```diff
@@ -0,0 +1,1070 @@
...
+const MIN_CHUNK_SIZE: u64 = 50_000;
+const MAX_CHUNK_SIZE: u64 = 500_000;
+const DEFAULT_CHUNK_PERCENTAGE: f64 = 0.25;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ChangeRecord {
+  pub job_name: String,
+  pub query_name: String,
+  pub target_block: u64,
+  pub atomic_data: serde_json::Value,
```
michaeldjeffrey, line 24: There seems to be a lot of validation that this value is an array. 
Could we reduce almost all of that by changing this field to a `Vec<serde_json::Value>`?

utils/atomic-data-publisher/src/errors.rs
```diff
@@ -0,0 +1,58 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum AtomicDataError {
+  #[error("Database error: {0}")]
+  DatabaseError(String),
```
michaeldjeffrey, line 6: `thiserror` provides `#[from]` attribute that auto implements of much of the stuff you have below.

```rs
  #[error("Database error: {0}")]
  DatabaseError(#[from] sqlx::Error)
```

This also means we keep all the context of the error stack, rather than what we get from `.to_string()`.

utils/atomic-data-publisher/src/service.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    let publisher = Arc::new(Publisher::new(
+      config.service.polling_jobs.clone(),
+      keypair,
+      config.service.clone(),
+      config.ingestor.clone(),
+      metrics.clone(),
+    ).await?);
+
+    let (shutdown_sender, shutdown_signal) = tokio::sync::watch::channel(false);
```
michaeldjeffrey, line 90: There's a crate called `triggered` that is a nicer interface for shutdown signals than a `watch::channel()`

utils/atomic-data-publisher/src/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+    // Initialize all metrics with zero values so they appear in /metrics endpoint
+    // This must be called AFTER the Prometheus exporter is installed
+    counter!("atomic_data_publisher_errors_total").absolute(0);
+    counter!("atomic_data_publisher_changes_published_total").absolute(0);
+    counter!("atomic_data_publisher_ingestor_connection_failures_total").absolute(0);
+    counter!("atomic_data_publisher_ingestor_retry_attempts_total").absolute(0);
+    counter!("atomic_data_publisher_ingestor_publish_failures_total").absolute(0);
+
+    // Initialize histograms (they'll show up after first recording)
+    histogram!("atomic_data_publisher_database_query_duration_seconds");
```
michaeldjeffrey, line 26: you can `describe_histogram!` to initialize with no values.

utils/atomic-data-publisher/src/polling_service.rs
```diff
@@ -0,0 +1,304 @@
...
+    let changes = vec![record.clone()];
+
+    let (total_published, total_failed, should_break) = self.process_batches(changes, record.target_block).await?;
+    self.database.finalize_job(&record, total_failed).await?;
+    Ok((total_published, total_failed, should_break))
+  }
+
+  async fn process_batches(
+    &self,
+    changes: Vec<ChangeRecord>,
```
michaeldjeffrey, line 156: This function is only every called with a Vec with 1 ChangeRecord.

PR #1191, opened 2026-05-27
Feat/proxy assign queue relinquish

bryzettler:
When a wallet assigns a delegated position to a proxy that has already voted on an active proposal, the proxy's original sweep couldn't have included that position, so the assign flow propagates the vote onto it — but the resulting
  vote marker is never scheduled for relinquish, so it lingers after voting ends.

  This replaces that with a queueProxyVoteV0 task per active proposal the proxy voted on, queued only when a delegated position is in the assignment. queueProxyVoteV0 takes voter as a plain account (not a signer), so the assigning
  wallet can queue it. The vote-service cron then counts the votes for the newly delegated positions and requeues their relinquish via the task-queue PDA. Applied in both assign.ts and useAssignProxies.ts.

  Vote propagation becomes asynchronous — handled by the cron shortly after assignment, not inline in the assign tx.

  Tests: e2e coverage in governance.test.ts (with a createHeliumOrgVotingProposal helper that seizes the forked Helium org to create a discoverable voting proposal): a proxy casts a real proxiedVoteV1, a wallet position is delegated,
  and assignProxies is asserted to emit queueProxyVoteV0. Still needs a live surfpool run to validate.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,6 @@
+# helium-program-library — Claude rules
+
+## Housekeeping
+
+- **Never commit plans or specs to this repo.** Brainstorming output, implementation plans, design specs, and similar working documents go in `~/.claude/plans/`, not in `docs/` or anywhere else in the project tree. This includes anything written by the `superpowers:brainstorming`, `superpowers:writing-plans`, or related skills — those skills' default paths (`docs/superpowers/...`) are overridden by this rule.
+- If a plan or spec is already staged or committed, move it to `~/.claude/plans/` and drop it from the commit before pushing.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,6 +1,0 @@
-# helium-program-library — Claude rules
-
-## Housekeeping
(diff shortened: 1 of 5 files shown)


ChewingGlass (approved):

CLAUDE.md
abhay, line 1: fine to keep this in if you want but your choice. it's just a little light :)

abhay:
Probably need to add some coverage for delegated proxy assignment creating both countProxyVoteV0 and queued cleanup.

Also some specific feedback here: 

https://github.com/helium/helium-program-library/pull/1192

CLAUDE.md
bryzettler: yeah I figure this will grow as we go, I have a full local one and got sick of `/docs` getting added every pr

bryzettler: Or I can adjust my local and forego this fully

bryzettler:
Included in https://github.com/helium/helium-program-library/pull/1196

PR #1195, opened 2026-06-26
fix(blockchain-api): batch governance vote validation to fix 504 for large proxy voters

madninja:
## Problem

The governance `vote` and `relinquishVote` handlers can't build a transaction for wallets with many delegated positions — the request returns **504 Gateway Time-out** and the vote never goes through. The largest proxy voter (~1,094 delegated positions) hits this every time; small wallets are unaffected.

Reproduced directly against production (the build endpoint needs no signature):

| positionMints | result |
|---|---|
| 1 | HTTP 200 in **1.6s** |
| 1094 | HTTP **504** in **60.7s** (nginx gateway timeout) |

## Root cause

Both handlers validate each position mint with **two sequential RPC calls** inside a loop:
- `validatePositionOwnership` → `getAccountInfo` on the ATA, and
- `proxyAssignmentV0.fetchNullable` for non-owned positions.

That's ~2 round-trips per position, run sequentially. At 1,094 positions that's **~2,200 sequential RPCs** in one request handler, which blows past the gateway timeout. All of that work collapses into a single `proxiedVoteV1` instruction.

## Fix

Batch the lookups up front, then classify in memory:
- `validatePositionOwnershipBatch` — one `getMultipleAccounts` per 100 ATAs instead of one `getAccountInfo` per position.
- one `fetchMultiple` for the distinct registrars (→ `proxyConfig`).
- one `fetchMultiple` for the proxy assignments.

The instruction-building path (single `proxiedVoteV1` for the proxy + one `voteV0` per owned position, plus the queue tasks) is **unchanged**, so the produced transactions are identical. Same change applied to `vote.ts` and `relinquish-vote.ts`.

## Verification

Replayed against mainnet for the 1,094-position wallet:
- classification identical to the old path: **1,094 proxied, `hasProxies=true`, 0 missing, 0 owned**
- **~45 RPC calls** instead of ~2,200
- **~6s** instead of a 60s gateway timeout

The existing `tests/e2e/governance.test.ts` covers `vote`/`relinquishVote` for single/multi positions; this change preserves that behavior. There is no existing test at the large-proxy scale that triggers the timeout — a regression test that votes with a large proxied set would guard against this returning.

### DIFF
--- a/.github/workflows/blockchain-api-e2e.yml
+++ b/.github/workflows/blockchain-api-e2e.yml
@@ -46,6 +46,7 @@ jobs:
     env:
       NODE_ENV: test
       NEXT_PORT: 3000
+      SOLANA_CLI_VERSION: 2.1.6
       SURFPOOL_RPC_URL: http://127.0.0.1:8899
       ASSET_ENDPOINT: ${{ secrets.ASSET_ENDPOINT }}
       NEXT_PUBLIC_SOLANA_CLUSTER: mainnet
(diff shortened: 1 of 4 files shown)


bryzettler:
included in https://github.com/helium/helium-program-library/pull/1196

PR #1197, opened 2026-06-30
HIP 149: on-chain implementation (Decisions 1, 2, 4)

madninja:
Implements the complete on-chain (`helium-program-library`) surface of HIP 149: the
Decision 1 deployer earnings backstop, the Decision 2 operations and growth
supplement, and the Decision 4 Council compensation carve-out. Decision 3 (PoC
retirement) is verifier-oracle work and lands separately; the Decision 4 curtailment
and vacancy provisions are governance/program-upgrade actions with no on-chain code.

## Decision 1 — deployer earnings backstop

- **Floor (top-up):** when the Mobile data deployer baseline falls below half the
  carrier-paid USD for the epoch, mint a top-up so deployers reach that target.
  Bounded by `max(0, smoothed_hnt_burned - net_emissions_cap)` so the top-up and the
  existing HIP 20 net-emissions re-emit share one burn budget and never re-mint the
  same destroyed HNT twice.
- **Cap (overflow to stakers):** when the data bucket would pay deployers more than
  3x the carrier-paid USD, redirect the excess from the rewards escrow to the
  delegator pool. `total_rewards` unchanged; net-supply-neutral.
- HNT/USD from the maintained on-chain Pyth push account (`HNT_PYTH_PRICE_FEED`), the
  same feed the price oracle already uses to price rewards.
- **Fail-safe:** a missing, stale, malformed, or wrong-feed price — or absent Mobile
  epoch-info accounts — makes the backstop dormant for that epoch (baseline emission
  + existing re-emit, no top-up/cap) rather than failing the instruction.

## Decision 2 — operations and growth supplement

- A per-epoch HNT mint bounded by two hardcoded timestamps: a flat first window then
  a linear taper to zero, both self-terminating (no vote or manual halt needed to end
  them). `issue_rewards_v0` mints the per-sub-DAO slice while the window is active.
- `calculate_utility_score_v0` records the same total in `current_hnt_supply` so the
  mint is not misread as burned HNT. Applied there only; adding it to
  `total_hnt_burned` as well would double-count. Verified on a live validator: the
  supply correction equals the minted total exactly, so `smoothed_hnt_burned` reflects
  only carrier burns.
- Ships **INACTIVE by default** (far-future start) so the program is safe to deploy
  and review before the patch-time constants are finalized.

## Decision 4 — Council compensation

- The supplement mint is split: 1.25% to a Council fanout (a `mini_fanout` that
  distributes pro-rata among the seated community members) and the remainder to the
  Receiving Entity vault. The two destinations sum to the per-sub-DAO supplement, so
  total minted and the supply correction are unchanged.
- The Council destination is pinned to a patch-time constant
  (`COUNCIL_FANOUT_TOKEN_ACCOUNT`); the fanout's owner is the governance multisig,
  never the Receiving Entity. `create-council-fanout` (admin-CLI) creates it.

## Commits (organized for security review)

- **Contract** — all Solana program changes (`helium-sub-daos` + `hpl-crons`) in a
  single commit; each program file shows its complete final diff.
- **Tooling** — admin-CLI + spl-utils (price/account wiring, council fanout creation).
- **Tests** — backstop floor/cap/dormant integration tests, the optional-account call
  sites, and a CU-limit bump that stabilizes the world-setup Metaplex CPI. Full
  `helium-sub-daos` suite passes 89/89.

## Deploy-time steps (not code)

- Set the three patch-time constants: `INFLATION_START`, `SUPPLEMENT_VAULT_OWNER`,
  `COUNCIL_FANOUT_TOKEN_ACCOUNT`.
- Run `create-council-fanout` at council-seating, then hardcode its printed token
  account into `COUNCIL_FANOUT_TOKEN_ACCOUNT`.
- Bump the HNT circuit breaker (multisig `update_mint_windowed_breaker_v0`) to cover
  the supplement, and point the cron's `supplement_vault` / `council_vault` at the
  real accounts.

## Not in this PR (tracked separately)

- The `mobile_verifier` change to size the Mobile data pool from the actual escrow so
  the Decision 1 cap's redirect is reflected off-chain, and the Decision 3 PoC
  retirement (Mobile + IoT verifier oracles) — both in the `oracles` repo. The floor
  needs no oracle change; the cap is dormant-correct until that ships.

### DIFF
--- a/packages/spl-utils/src/constants.ts
+++ b/packages/spl-utils/src/constants.ts
@@ -12,7 +12,8 @@ export const MOBILE_MINT = new PublicKey(
 
 export const IOT_MINT = new PublicKey("iotEVVZLEywoTn1QdwNPddxPWszn3zFhEot3MfL9fns");
 
-// TODO: Replace with actual HNT feed
+// Mainnet HNT/USD Pyth pull-oracle push account (shard 0 for HNT_PRICE_FEED_ID),
+// kept continuously fresh on-chain.
 export const HNT_PYTH_PRICE_FEED = new PublicKey(
   "4DdmDswskDxXGpwHrXUfn2CNUm9rt21ac79GHNTN3J33"
 );
--- a/programs/helium-sub-daos/src/backstop.rs
+++ b/programs/helium-sub-daos/src/backstop.rs
@@ -30,6 +30,19 @@ use anchor_lang::prelude::*;
 /// The backstop keys off the Mobile sub-DAO's `dc_burned` and percent share.
 pub const MOBILE_SUB_DAO: Pubkey = pubkey!("Gm9xDCJawDEKDrrQW6haw94gABaYzQwCq4ZQU8h8bd22");
 
+/// The canonical mainnet HNT/USD Pyth push account (shard 0 for `HNT_PRICE_FEED_ID`),
+/// kept continuously fresh on-chain and mirrored in `spl-utils`' `HNT_PYTH_PRICE_FEED`.
+/// `calculate_utility_score_v0` pins the supplied price account to this key so a caller
+/// cannot substitute a different (stale-but-valid) HNT price update to steer the backstop.
+///
+/// Operational constraint: this constant and the account the end-epoch cron forwards must
+/// move together. A closed or stale canonical account fails safe (the content checks in
+/// `read_hnt_price` make the epoch dormant, issuance continues). But repointing the cron
+/// at a *different* account without also upgrading this constant hard-fails the pin
+/// (`InvalidPriceOracle`) and halts issuance. To rotate the Pyth feed account, upgrade this
+/// constant and update the cron in the same rollout.
+pub const HNT_PYTH_PRICE_FEED: Pubkey = pubkey!("4DdmDswskDxXGpwHrXUfn2CNUm9rt21ac79GHNTN3J33");
+
 /// Mobile data bucket fraction (HIP 149 Decision 3), as a percentage. Hardcoded.
 pub const MOBILE_DATA_BUCKET_PERCENT: u128 = 70;
 
@@ -49,9 +62,17 @@ pub struct BackstopInput {
   /// `10^(hnt_decimals - pyth_exponent - 5)`, the DC->HNT scale factor (mirrors
   /// `mint_data_credits_v0`). With 8 HNT decimals and a -8 exponent this is `10^11`.
   pub decimals_factor: u128,
-  /// Confidence-adjusted Pyth HNT price (`ema_price - 2 * ema_conf`), guaranteed `> 0`
-  /// by the caller. Same convention as `mint_data_credits_v0`.
-  pub hnt_price_with_conf: u64,
+  /// Lower-bound Pyth HNT price (`ema_price - 2 * ema_conf`), guaranteed `> 0` by the
+  /// caller. Used for the floor/top-up: a lower price converts the USD target into more
+  /// HNT, which is the conservative direction for a minimum. Same convention as
+  /// `mint_data_credits_v0`.
+  pub hnt_price_floor: u64,
+  /// EMA (point-estimate) Pyth HNT price, no confidence adjustment. Used for the earnings
+  /// cap so it binds at exactly `3.0 x carrier_paid_USD` at the point price. The floor uses
+  /// the lower bound (never underpay a minimum); the cap uses the point price (land on the
+  /// HIP's 3.0x rather than biasing the ceiling in either direction with the confidence
+  /// width) (I-01).
+  pub hnt_price_cap: u64,
 }
 
 /// Result of the backstop computation for one epoch.
@@ -90,11 +111,12 @@ pub fn compute_backstop(input: &BackstopInput) -> BackstopOutput {
 
   // carrier_paid_USD = dc_burned x 1e-5. The cap is 3.0x of it (dc_burned*3 in DC units);
   // the floor target (0.5x) is computed below, after the share guard, since it's only
-  // used on the top-up path.
+  // used on the top-up path. The cap uses the EMA point price so it binds at exactly 3.0x;
+  // the floor uses the lower (confidence-adjusted) price so a minimum is never underpaid.
   let deployer_cap_hnt = scale_dc_to_hnt(
     (input.mobile_dc_burned as u128).saturating_mul(3),
     input.decimals_factor,
-    input.hnt_price_with_conf,
+    input.hnt_price_cap,
   );
 
   // Total emission unaffected by the backstop: schedule + the existing re-emit.
@@ -115,7 +137,7 @@ pub fn compute_backstop(input: &BackstopInput) -> BackstopOutput {
   let target_hnt = scale_dc_to_hnt(
     (input.mobile_dc_burned / 2) as u128,
     input.decimals_factor,
-    input.hnt_price_with_conf,
+    input.hnt_price_floor,
   );
 
   // deployer_baseline = (emission + existing_re_emit) x mobile_share x 0.70.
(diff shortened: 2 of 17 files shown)


programs/helium-sub-daos/src/instructions/issue_rewards_v0.rs
```diff
@@ -194,28 +206,51 @@ pub fn handler(ctx: Context<IssueRewardsV0>, args: IssueRewardsArgsV0) -> Result
...
+  // claims already key off the DAO-level delegation_rewards_issued. A zero ceiling
+  // (no carrier burn this epoch, or no price oracle) disables the redirect.
+  let is_mobile = TESTING || ctx.accounts.sub_dao.key() == crate::backstop::MOBILE_SUB_DAO;
+  let staker_overflow: u64 = if is_mobile {
+    crate::backstop::staker_overflow(rewards_amount, ctx.accounts.dao_epoch_info.deployer_cap_hnt)
+  } else {
+    0
+  };
+
+  let delegation_mint = delegation_rewards_amount
```
bryzettler: This naming of `_mint` confused me

programs/helium-sub-daos/src/state.rs
```diff
@@ -176,11 +176,21 @@ pub struct DaoEpochInfoV0 {
...
+  // shared delegator pool.
+  pub deployer_cap_hnt: u64,
 }
 
 impl DaoEpochInfoV0 {
   pub fn size() -> usize {
-    60 + 8 + std::mem::size_of::<DaoEpochInfoV0>()
+    // -16 for the two HIP 149 backstop u64s (hnt_price_used, deployer_cap_hnt),
+    // which consume reserved padding rather than growing the 356-byte account.
+    60 + 8 + std::mem::size_of::<DaoEpochInfoV0>() - 16
```
bryzettler: we're sure theres padding on the acct?

programs/helium-sub-daos/src/instructions/issue_rewards_v0.rs
madninja: Good catch. Renamed to `delegation_pool_amount`: it is the amount minted to the delegator pool (the 6% delegator slice plus any HIP 149 earnings-cap overflow), not a mint account. Fixed in 04b6eb580.

programs/helium-sub-daos/src/state.rs
madninja: Yes: the base `size()` is `60 + 8 + size_of`, so there is a fixed 60-byte reserve over discriminator + struct, and `size()` is only used to init the fresh per-epoch `dao_epoch_info` (no realloc of existing accounts), so consuming reserve is safe. Separately, I dropped the `hnt_price_used` field entirely (it was written but never read on-chain; the price is recoverable from `deployer_cap_hnt` + `dc_burned`, or the public Pyth feed). So the reserve consumption is now 8 bytes rather than 16 (52 left). Fixed in 04b6eb580.

programs/helium-sub-daos/src/state.rs
```diff
@@ -176,11 +176,20 @@ pub struct DaoEpochInfoV0 {
...
+  // deployer_cap_hnt, or from the public Pyth feed at the epoch timestamp.
+  pub deployer_cap_hnt: u64,
 }
 
 impl DaoEpochInfoV0 {
   pub fn size() -> usize {
-    60 + 8 + std::mem::size_of::<DaoEpochInfoV0>()
+    // -8 for the HIP 149 backstop u64 (deployer_cap_hnt), which consumes reserved
+    // padding rather than growing the account.
+    60 + 8 + std::mem::size_of::<DaoEpochInfoV0>() - 8
```
michaeldjeffrey: I don't quite understand this one.

This calculation was added in https://github.com/helium/helium-program-library/pull/733/changes#diff-b1ba70a1e6abebb20ab6878a5d4b46414e8a0bf6377f5d12a43f9eff10d2427dR168

We've added 5 `u64` fields since then without changing the size calculation. Why now?

Is `60` also reserved space?

bryzettler: 60 is padding, 8 is the discriminator, and then it adds the size of the struct. I think we can go without the - 8 all together

madninja: Agreed, the `- 8` was wrong. `std::mem::size_of::<DaoEpochInfoV0>()` already accounts for `deployer_cap_hnt`, which is exactly why the prior 5 u64 fields needed no change to `size()`; the `- 8` only shrank the 60-byte reserve to 52 for no benefit. Reverted `size()` to the base `60 + 8 + size_of` (60 padding + 8 discriminator + struct) in 71036eb03. Full suite still 89/89 with the +8 allocation.

packages/spl-utils/src/constants.ts
```diff
@@ -12,7 +12,8 @@ export const MOBILE_MINT = new PublicKey(
 
 export const IOT_MINT = new PublicKey("iotEVVZLEywoTn1QdwNPddxPWszn3zFhEot3MfL9fns");
 
-// TODO: Replace with actual HNT feed
+// Mainnet HNT/USD Pyth pull-oracle push account (shard 0 for HNT_PRICE_FEED_ID),
+// kept continuously fresh on-chain.
 export const HNT_PYTH_PRICE_FEED = new PublicKey(
   "4DdmDswskDxXGpwHrXUfn2CNUm9rt21ac79GHNTN3J33"
```
bryzettler, line 18: Pro feed being updated every ~5mins `He5mhwVQQNvjFxqjEjFDb7enJWFwFJ7Rq7zknqBz89A5`

programs/helium-sub-daos/src/instructions/calculate_utility_score_v0.rs
```diff
@@ -14,6 +21,13 @@ pub struct CalculateUtilityScoreArgsV0 {
 
 pub const TESTING: bool = std::option_env!("TESTING").is_some();
 
+// Mirror of `data_credits::mint_data_credits_v0::HNT_PRICE_FEED_ID`. Duplicated here
+// (rather than imported) because data-credits depends on this crate, not the reverse.
+pub const HNT_PRICE_FEED_ID: [u8; 32] = [
+  0x64, 0x9f, 0xdd, 0x7e, 0xc0, 0x8e, 0x8e, 0x2a, 0x20, 0xf4, 0x25, 0x72, 0x98, 0x54, 0xe9, 0x02,
+  0x93, 0xdc, 0xbe, 0x23, 0x76, 0xab, 0xc4, 0x71, 0x97, 0xa1, 0x4d, 0xa6, 0xff, 0x33, 0x97, 0x56,
+];
+
```
bryzettler, line 30: will have to be updated

programs/helium-sub-daos/src/backstop.rs
```diff
@@ -0,0 +1,410 @@
...
+/// `calculate_utility_score_v0` pins the supplied price account to this key so a caller
+/// cannot substitute a different (stale-but-valid) HNT price update to steer the backstop.
+///
+/// Operational constraint: this constant and the account the end-epoch cron forwards must
+/// move together. A closed or stale canonical account fails safe (the content checks in
+/// `read_hnt_price` make the epoch dormant, issuance continues). But repointing the cron
+/// at a *different* account without also upgrading this constant hard-fails the pin
+/// (`InvalidPriceOracle`) and halts issuance. To rotate the Pyth feed account, upgrade this
+/// constant and update the cron in the same rollout.
+pub const HNT_PYTH_PRICE_FEED: Pubkey = pubkey!("4DdmDswskDxXGpwHrXUfn2CNUm9rt21ac79GHNTN3J33");
```
bryzettler, line 44: update

programs/hpl-crons/src/instructions/queue_end_epoch.rs
```diff
@@ -5,7 +5,7 @@ use helium_sub_daos::{
   accounts::CalculateUtilityScoreV0, instruction::IssueRewardsV0, CalculateUtilityScoreArgsV0,
```
bryzettler, line 5: Well have to close the existing cron an recreate it since this adds a hnt_price_oracle

bryzettler (approved):

PR #1200, opened 2026-07-07
feat(blockchain-api): build wallet transactions server-side (tokens, DC, claims, automation, squads, data-only)

madninja:
Adds the blockchain-api endpoints that let the `helium-wallet` CLI build all of its Solana transactions server-side and sign + submit locally. This is the API counterpart to helium-wallet-rs#560; together they move transaction construction out of the wallet SDK and behind the API.

## Endpoint groups
- **Tokens**: transfer, multi-transfer, burn, memo.
- **Data credits**: burn, mint, delegate.
- **Hotspots**: single-hotspot claim, whole-wallet claim, burn, transfer, update/assert.
- **Claim automation (tuktuk)**: single-cron lifecycle with granular claim entries, funding/top-up, requeue, and close.
- **Squads v4**: proposal lifecycle (approve / reject / cancel / execute / propose-config-change) plus an optional propose mode on the action endpoints (token and DC transfer/burn, DC delegate, hotspot burn/transfer) that builds the action as a vault proposal instead of a direct transaction.
- **Data-only hotspots**: issue and onboard (IoT and mobile).

## Notable points
- Data-only issue/onboard are built on-chain through the data-only escrow, with the issue transaction co-signed by the ECC verifier. They do not go through the maker onboarding server, which requires a pre-registered maker record and cannot self-serve a data-only hotspot. The issue endpoint verifies the verifier returned the same message it submitted (only a signature added) before relaying it, and requires an https verifier URL.
- Swap quotes exclude RFQ (JupiterZ) routes, which return maker-co-signed transactions a self-signing wallet cannot complete, while keeping the getQuote/getTokens rate-limit cache from develop.
- Action endpoints return a consistent envelope (`transactionData` + `estimatedSolFee`) so the wallet can decode, display, and sign every command uniformly.

## Verification
E2E on a surfpool mainnet fork (automation lifecycle, squads propose/execute, token and DC burns) and against devnet: a token transfer and a data-only mobile issue + onboard both land and finalize on-chain, with only the ECC verifier as an external dependency.

### DIFF
--- a/packages/blockchain-api-client/src/schemas/hotspots.ts
+++ b/packages/blockchain-api-client/src/schemas/hotspots.ts
@@ -127,7 +127,6 @@ export const SetupAutomationInputSchema = z.object({
   // Raw crontab string (clockwork format) the cron fires on.
   cronSchedule: z.string().min(1),
   duration: z.number().int().min(1), // Number of claims to pre-fund
-  totalHotspots: z.number().int().min(1),
 });
 
 export const FundAutomationInputSchema = z.object({
--- a/packages/blockchain-api-client/src/schemas/hotspots.ts
+++ b/packages/blockchain-api-client/src/schemas/hotspots.ts
@@ -165,6 +165,26 @@ export const RemoveEntityFromAutomationInputSchema = z.object({
   index: z.number().int().min(0),
 });
 
+// ---- Data-only hotspot onboarding ----
+// The onboarding server builds the issue (via the ECC verifier) and onboard
+// transactions; these endpoints relay them for local signing.
+
+export const IssueDataOnlyHotspotInputSchema = z.object({
+  walletAddress: WalletAddressSchema,
+  // Base64 add-gateway token (BlockchainTxnAddGatewayV1 envelope) from the hotspot.
+  addGatewayTxn: z.string().min(1),
+});
+
+export const OnboardDataOnlyHotspotInputSchema = z.object({
+  walletAddress: WalletAddressSchema,
+  network: z.enum(["iot", "mobile"]),
+  hotspotAddress: HeliumPublicKeySchema,
+  lat: z.number().min(-90).max(90).optional(),
+  lng: z.number().min(-180).max(180).optional(),
+  elevation: z.number().optional(), // IoT only
+  gain: z.number().optional(), // IoT only
+});
+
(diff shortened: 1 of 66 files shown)


packages/blockchain-api-client/src/schemas/hotspots.ts
```diff
@@ -92,13 +118,15 @@ export const GetAutomationStatusInputSchema = z.object({
   walletAddress: WalletAddressSchema,
 });
 
+// Best-effort classification of a raw cron string, surfaced on the status
+// output for display only. Funding math is per-firing and never depends on it.
 export const AutomationScheduleSchema = z.enum(["daily", "weekly", "monthly"]);
 
 export const SetupAutomationInputSchema = z.object({
   walletAddress: WalletAddressSchema,
-  schedule: AutomationScheduleSchema,
```
bryzettler, line 99: why was this dropped?

packages/blockchain-api-client/src/schemas/hotspots.ts
```diff
@@ -208,7 +306,8 @@ export const AutomationStatusOutputSchema = z.object({
   isOutOfSol: z.boolean(),
   currentSchedule: z
     .object({
-      schedule: AutomationScheduleSchema,
```
bryzettler, line 211: same question

packages/blockchain-api-client/src/schemas/hotspots.ts
madninja: Good catch. Restored in 15f76a009: `schedule` now accepts either a preset cadence (daily/weekly/monthly) or a raw clockwork crontab string. Presets resolve to a crontab server-side via the existing `getScheduleCronString` helper. The raw-cron rework had narrowed this to crontab-only, which is what dropped the short forms.

packages/blockchain-api-client/src/schemas/hotspots.ts
madninja: This one is not actually dropped. The diff reads as a deletion because a new `cron` (raw crontab) field was inserted above it and `schedule` was re-added just below; the status output still surfaces `schedule` as a best-effort, display-only classification next to the raw `cron`. The input side (line 99) accepts the presets again as of 15f76a009.

PR #1202, opened 2026-07-10
Zyrb/review/single hotspot claim

bryzettler:


packages/blockchain-api/src/server/api/routers/hotspots/procedures/hotspot-cnft.ts
```diff
@@ -0,0 +1,126 @@
...
+  const assetEndpoint = env.ASSET_ENDPOINT || connection.rpcEndpoint;
+
+  const [{ asset, args, accounts, remainingAccounts }, contractExists] =
+    await Promise.all([
+      proofArgsAndAccounts({
+        connection,
+        assetId: new PublicKey(assetId),
+        assetEndpoint,
+      }),
+      hasRewardContract(hotspotPubkey),
```
madninja: Low: `hasRewardContract` now runs in this `Promise.all`, before the ownership check at line 88. The pre-refactor burn/transfer code only queried the reward contract after a passing ownership check. Two deltas: (1) an unauthorized caller now always triggers this DB lookup, and (2) if the lookup rejects, the raw error surfaces as a 500 before the `UNAUTHORIZED` check is reached, so a would-be 401 can become a 500. Authorization itself is unchanged (the `CONFLICT` guard still sits after `UNAUTHORIZED`, so no reward-contract state leaks to an unauthorized caller). If you want the original ordering back, keep `proofArgsAndAccounts` in the `Promise.all` but move `hasRewardContract` to just before the `contractExists` guard.

packages/blockchain-api/src/server/api/routers/hotspots/procedures/transferHotspot.ts
```diff
@@ -65,84 +34,34 @@ export const transferHotspot = publicProcedure.hotspots.transferHotspot.handler(
...
+      errors,
+    });
 
-    // Validate public keys
-    let payerPubkey: PublicKey;
     let recipientPubkey: PublicKey;
-
     try {
-      payerPubkey = new PublicKey(walletAddress);
       recipientPubkey = new PublicKey(recipient);
```
madninja: Low: the recipient is parsed here, after the full `resolveOwnedHotspotCnft` preamble. The pre-refactor code parsed payer + recipient together at the top, before any RPC. Net effect: for a malformed recipient on a hotspot the caller does not own, the original returned `BAD_REQUEST` first, whereas now an `UNAUTHORIZED`/`NOT_FOUND`/`CONFLICT` from the preamble is thrown first and the bad recipient is masked. No bad transaction is ever built (this parse still precedes `createTransferInstruction`); only the returned error code and ordering differ. Parsing the recipient before the preamble would restore the original ordering.

madninja (commented):
Behavior-preservation review of the refactor: it's faithful on every correctness/authorization dimension I checked. Two low-severity error-ordering notes below, neither weakens authorization or changes the resulting transaction.

PR #1205, opened 2026-07-13
ci: pin Node to 24.13.1 via .node-version

madninja:
Pin the repo's Node.js version to 24.13.1 and standardize on it across local dev, CI, and the service containers.

## What

- **`.node-version` = `24.13.1`** at the repo root — read by fnm, nvm, asdf, Volta, so local tooling auto-selects the version CI already uses. Previously nothing bridged local shells to CI's version, so a machine's global Node (e.g. 26, which breaks mocha/yargs) could silently be used for `anchor test`.
- **`engines.node` = `">=24.13.0 <25"`** in the root `package.json` (advisory; no `engine-strict`).
- **CI reads `.node-version`.** Every real `actions/setup-node` (`setup-ts`, `npm-publish`) uses `node-version-file: .node-version` instead of a duplicated `NODE_VERSION` env, and the unused `NODE_VERSION` is removed from all six workflows. Dropped the inert `node-version` inputs to the `setup-anchor` composite action (it declares no inputs / runs no `setup-node`).
- **Service containers moved to Node 24.** Every Node service Dockerfile base image bumped `node:22-alpine → node:24-alpine`, so deployed containers run the same Node major the repo builds and tests with. The per-service `Test Docker Builds` CI jobs build and install each image on the new base, so they verify the bump. Rust-based service images are unaffected.
- **blockchain-api e2e moved onto the pin** (was hard-coded `"20.x"`, a carryover from #1145; Next 15 / React 19 run on Node 24), and **`@types/node` in `packages/blockchain-api` bumped `^20 → ^24`** to match its runtime. Note on lockfile churn: this is *not* a workspace-wide dedup — `@types/node` isn't root-pinned, so the shared dev-tooling (ts-node, changesets, tailwindcss, …) resolves its loose `@types/node` peer to the highest available, and bumping blockchain-api to `^24` also pulls that shared copy `20.19.37 → 24.13.3`. Service packages pinned `^18.x`/`^20.x` are unchanged, so the lockfile still contains several `@types/node` versions (12/18/20/24). Aligning those is out of scope.

## Effective Node version changes

| Context | Before | After |
| --- | --- | --- |
| Local + CI build/test | 24.13.1 (via duplicated env) | 24.13.1 (via `.node-version`) |
| Service containers | 22 | **24** |
| blockchain-api e2e | 20 | **24** |

## Verification

CI (base `develop`): `tests.yaml` (Rust + TS suites) green; the `blockchain-api-e2e` matrix (17 jobs) runs on Node 24 and is green (the one `reward-contract` flake — a `surfpool` tx-confirmation timeout — passed on re-run); the `Test Docker Builds` matrix builds every service image on `node:24-alpine`.


madninja:
Good catch — corrected the PR description. You're right that this is not a workspace-wide `@types/node` dedup: the lockfile still resolves 12/18/20/24, and `20.19.37` remains for the `^18`/`^20`-pinned service packages. What actually happened is that bumping `packages/blockchain-api` to `^24` pulled the shared, un-root-pinned dev-tooling copy of `@types/node` from `20.19.37` to `24.13.3`; the service packages are untouched, and aligning their ranges is intentionally left out of scope.

bryzettler (approved):

PR #1206, opened 2026-07-13
feat: migrate governance positions + getPositions, harden fee-payer rent accounting

bryzettler:
## Summary

Extends `/migrate` to migrate **VSR governance positions** (vehnt/veiot/vemobile) alongside tokens and hotspots, adds `governance.getPositions` for the wallet UI, and hardens fee-payer rent accounting. Positions are discovered server-side (registrar-agnostic, no new input) and transferred with `transferPositionV0`. Migration stays rent-neutral: the fee payer fronts destination-ATA rent and recovers it by closing the source ATA.

## Changes

- **feat: migrate governance positions** — enumerate every VSR position the source owns (freeze-authority classification) and transfer each to the destination; re-enumerated per paginated call, so migrated positions drop out naturally.
- **feat: `governance.getPositions`** — `GET /governance/positions/wallet/{wallet}` returns mint, position, registrar, `amountDeposited` (`TokenAmountOutput` shape), `numActiveVotes`, lockup.
- **fix: skip stale-enumerated positions** — verify the source ATA still holds the NFT before building each transfer; avoids `TransferPositionV0` → `AccountNotInitialized (3012)` when a lagged RPC index reports a closed ATA.
- **fix: group create/close instructions** — token (`[createAta, transfer, close]`) and position (`[transferPositionV0, close]`) are built as grouped `TransactionInstruction[][]` so each pair stays in one tx. Closes a rent-drain where a flat list could split a fee-payer-funded create from its offsetting close, letting a caller land the create alone and strand the rent. Enforced at build layer, independent of client submission. Mirrors the split-hotspot pattern.

## Testing

- `migration.test.ts`: 10/10 (incl. VSR position migration).
- `governance.test.ts`: 33/33 (incl. getPositions `amountDeposited` assertions).

## Security review — known & accepted (both pre-existing, `/migrate` is public so client trust isn't a safeguard)

1. **Mini-fanout subsidy (~0.01 SOL)** — intended cost of recreating the reward-split config at the destination; bounded by existing rate limiting. No fix that doesn't break legitimate fanout migrations.
2. **Isolated Jito tip (~10k lamports)** — after the grouping fix, the only remaining strandable cost, in the same bucket as the irreducible base tx fee. Rate limiting is the right control; a full server-side atomic-submission architecture would only shave the tip, not base fees. Revisit if a new non-self-neutralizing fee-payer-funded instruction is added.


packages/blockchain-api/src/server/api/routers/migration/procedures/migrate.ts
```diff
@@ -625,6 +674,23 @@ export const migrate = publicProcedure.migration.migrate.handler(
       }
     }
 
+    // Step 1b: Batch governance position transfers (always included, before
+    // hotspots). Each transfer needs only source signing besides the fee payer.
+    if (positionTransferGroups.length > 0) {
```
madninja: **[1 / Medium] First batch (tokens + positions) has no `MAX_JITO_BUNDLE_TXS` guard and isn't paginated.**

Steps 1 and 1b push all token/position drafts unconditionally; only the hotspot steps (2/3) respect the bundle limit, and Step 4 hard-throws `"Unable to fit Jito tip instruction into the transaction bundle"` once `allDrafts.length >= MAX_JITO_BUNDLE_TXS`. So if the auto-enumerated positions + full-balance tokens alone produce >5 txs, migration fails with **no recovery path** — re-enumeration is identical on every retry, and the position set isn't caller-controllable (`nextParams` only paginates hotspots).

This widens a pre-existing token-only latent case by adding auto-discovered positions. Suggest counting first-batch drafts against the limit and spilling overflow into `nextParams`, or documenting/enforcing a hard cap on first-batch size.

packages/blockchain-api/src/server/api/routers/migration/procedures/migrate.ts
```diff
@@ -88,53 +92,52 @@ const FANOUT_FUNDING_AMOUNT = solToLamportsBN(0.01).toNumber();
...
+    const { sourceWallet, destinationWallet, hotspots, tokens } = input;
 
-    // 3a. Validation — either wallet can be in the allowlist, or password bypasses it
-    const passwordValid =
-      !!password && !!env.MIGRATION_PASSWORD && password === env.MIGRATION_PASSWORD;
+    // Rate limit per wallet pair and per client IP. Pair is checked first; on a
+    // trip we skip the IP check so a rejected request doesn't consume IP budget.
+    const headerStore = await headers();
+    const clientIp =
+      headerStore.get("x-forwarded-for")?.split(",")[0]?.trim() || "unknown";
```
madninja: **[2 / Medium] Rate limiter isn't an abuse boundary — keys are caller-controlled.**

`x-forwarded-for.split(",")[0]` is the *leftmost*, client-supplied hop, which is trivially spoofable unless the ingress overwrites XFF (in which case you usually want the right-most trusted hop, not `[0]`). Wallet pairs are also freely generated. An attacker rotating both bypasses both limiters. Meanwhile `getPositions` runs under `publicProcedure` with no throttle at all.

On-chain drain is largely self-limiting (an attacker's rent flows to the fee payer; only the ~0.01 SOL fanout subsidy leaks, and that needs a real hotspot), so the accepted-risk note is fair there. The genuinely unbounded surface is **server-side RPC work** (see the `getPositionsForOwner` note). Suggestions: key the IP limiter off the trusted hop / a platform client-IP header (confirm what the ingress guarantees); throttle `getPositions`; and note in the PR that this is a courtesy throttle (and per-pod, so effective limit × pod count), not a security control.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/list-positions.ts
```diff
@@ -0,0 +1,52 @@
...
+  const tokenAccounts = await connection.getParsedTokenAccountsByOwner(owner, {
+    programId: TOKEN_PROGRAM_ID,
+  });
+
+  const positions: OwnedPosition[] = [];
+  for (const { account } of tokenAccounts.value) {
+    if (account.data.parsed.info.tokenAmount.amount !== "1") continue;
+
+    const mint = new PublicKey(account.data.parsed.info.mint);
+    const freezeAuthority = (await getMint(connection, mint)).freezeAuthority;
```
madninja: **[3 / Medium — perf] O(n) sequential RPC round-trips.**

This loop awaits `getMint` + `getAccountInfo(freezeAuthority)` one account at a time, so a wallet with many amount-1 accounts (all NFTs) triggers hundreds of serial RPC calls per request — on the public `getPositions` endpoint and again inside `migrate`. Combined with the throttle gap above, it's a cheap RPC-amplification lever, and it's slow for legitimate NFT-heavy wallets.

Batch it: `getMultipleAccounts` for the mints and for the freeze-authority accounts, or `Promise.all` the per-account work. The `getParsedTokenAccountsByOwner` response already carries mint + amount.

packages/blockchain-api/src/server/api/routers/migration/procedures/migrate.ts
```diff
@@ -88,53 +92,52 @@ const FANOUT_FUNDING_AMOUNT = solToLamportsBN(0.01).toNumber();
...
+    BUBBLEGUM_PROGRAM_ID
   );
   return bubblegumAuthorityPDAKey;
 }
 
 export const migrate = publicProcedure.migration.migrate.handler(
   async ({ input, errors }) => {
-    const { sourceWallet, destinationWallet, hotspots, tokens, password } =
-      input;
+    const { sourceWallet, destinationWallet, hotspots, tokens } = input;
```
madninja, line 133: **[4 / Low — cleanup] Dead `password` field.**

The gate removal drops `MIGRATION_PASSWORD` (env) and the handler's `password` read here, but `MigrateInputSchema` in `packages/blockchain-api-client/src/schemas/migration.ts` still declares `password: z.string().optional()`. The API now accepts a field it silently ignores — sweep-after-removal miss. Remove it from the schema.

packages/blockchain-api/src/server/api/routers/migration/procedures/migrate.ts
```diff
@@ -180,70 +189,105 @@ export const migrate = publicProcedure.migration.migrate.handler(
...
+        // its rent refund makes the migration rent-neutral for the fee payer.
+        // Fetch state BEFORE pushing any instruction — a frozen or empty account
+        // must not leave the fee payer paying unrefunded dest-ATA rent.
         const [mintInfo, sourceAtaInfo] = await Promise.all([
           getMint(connection, mintKey),
           getAccount(connection, sourceAta).catch(() => null),
         ]);
 
+        if (!sourceAtaInfo || sourceAtaInfo.amount <= BigInt(0)) {
+          warnings.push(`Skipping ${token.mint}: no balance to migrate`);
```
madninja, line 270: **[5 / Low] `rawAmount <= 0 → continue` still gates SPL tokens despite advisory-amount semantics.**

The loop still short-circuits on `rawAmount = BigInt(token.amount) <= 0` before reaching this full-balance logic. Since the client amount is now advisory for SPL (you always move the full on-chain balance), a caller sending `amount: "0"` for a genuinely funded token silently skips it. For the SPL path, drop the amount-based short-circuit and rely on the on-chain check (`!sourceAtaInfo || sourceAtaInfo.amount <= 0`); keep `rawAmount` only for the WSOL/native transfer where the client amount is authoritative.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/list-positions.ts
```diff
@@ -0,0 +1,52 @@
...
+type VsrProgram = Awaited<ReturnType<typeof initVsr>>;
+
+export interface OwnedPosition {
+  mint: PublicKey;
+  position: PublicKey;
+}
+
+/**
+ * Enumerate every VSR governance position owned by a wallet, registrar-agnostic.
+ * Positions are supply-1, decimals-0 frozen NFTs whose mint freeze authority is
```
madninja: **[6 / Nit] Comment vs. code.** Says positions are "decimals-0 frozen NFTs", but decimals aren't checked — the freeze-authority-owned-by-VSR check is what actually classifies them. Either add `decimals === 0` to the filter or drop that clause from the comment.

madninja (commented):
Reviewed the substantive changes (the grouping/atomicity fix, VSR position migration, `getPositions`, and the rate-limiter swap). Core mechanics are correct: I verified the batcher pushes each group all-or-nothing (`slice(0, prevLen)` rollback), and `transferPositionV0` creates the dest ATA on the fee payer + leaves `PositionV0`/`numActiveVotes` untouched, so the paired close keeps it rent-neutral and mid-vote positions migrate fine.

Leaving inline notes, ranked. #1 is the only one that can hard-fail a real user with no workaround; the rest are cleanup/perf/hardening. The accepted-risk section is fair on the on-chain economics — the main caveat is that the rate limiter isn't an abuse boundary (keys are caller-controlled) and the unbounded surface is server RPC cost, not fee-payer drain.

madninja (approved):

PR #1215, opened 2026-07-21
Fix/queue end epoch heap oom

bryzettler:
Why queue_end_epoch blew the heap

  HIP 149 threads hnt_price_oracle and the Mobile epoch-info accounts into every calculate_utility_score_v0 in the compiled end-epoch transaction, growing compiled_tx. The handler then .clone()d both
  compiled transactions into the TaskReturnV0s, doubling the now-larger allocations and pushing past the 32KB BPF heap when run via tuktuk's runTask. Fix: move instead of clone — the values aren't
  used afterward.

### DIFF
--- a/.changeset/hip149-idls-regen.md
+++ b/.changeset/hip149-idls-regen.md
@@ -1,0 +1,5 @@
+---
+"@helium/idls": patch
+---
+
+Republish IDL types with the HIP 149 changes from #1197: `queue_end_epoch` now requires the `hnt_price_oracle` account and `calculate_utility_score_v0` accepts an optional `hnt_price_oracle`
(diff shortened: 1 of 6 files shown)


madninja (approved):
Approved. The core fix is correct: `compiled_tx` and `compiled_reschedule_tx` are each used exactly once with no later use, so the two `.clone()` calls were pure waste that doubled peak SBF heap during `write_return_tasks` serialization and pushed it past the 32KB limit. Moving instead of cloning halves peak heap. Verified against the full handler.

The regression test is the right call and reproduces the exact prod path (queue task -> crank CPIs `queue_end_epoch` -> writes both return txs) at the default heap. The `check-idls-changeset` gate is a good structural fix for the stale published-idls problem.

Two non-blocking notes:
- Heap headroom is ~2x now but the 32KB ceiling stands; a third subDao or larger issue_rewards accounts could regress it. Worth a short comment near the fix so nobody re-adds a clone.
- The `update-data-credits.ts` change is pure prettier quote reformatting unrelated to the OOM; consider dropping it to keep the hotfix diff clean.

PR #1232, opened 2026-08-03
fix(blockchain-api): allow an off-curve owner when deriving associated token addresses

madninja:
## Problem

A Squads vault owns its assets through a program address, which is off the ed25519 curve. Eleven call sites derived associated token addresses without `allowOwnerOffCurve`, so `getAssociatedTokenAddressSync` threw `TokenOwnerOffCurveError` and took down the whole request rather than one derivation. Reproduced against a mainnet vault-held position by invoking the real helper:

```
RESULT: THREW -> TokenOwnerOffCurveError:
```

`buildClaimInstructions` is one of those call sites, and claim, undelegate and delegate all call it first, so all three were unavailable to any vault-held veHNT position with an unclaimed epoch. `delegate.ts` derived the reward account the same way when enabling autoclaim, so a vault could not turn automation back on either.

The derived address is identical either way and the programs re-validate the account they receive (`claim_rewards_v1` constrains `delegator_ata` by `associated_token::authority` and `position_token_account` by `token::mint` plus a positive balance), so this decides whether a multisig can use an endpoint at all, not who is allowed to do what.

## Scope

This started larger and was cut back deliberately. It also changed transaction packing so a vault's batches would survive being re-wrapped as a Squads proposal, and skipped the Jito tip for a payer that cannot sign. Both are dropped, because the benefit turned out to be blocked downstream:

- `getJitoTipTransaction` builds a `SystemProgram.transfer` **from** the fee payer, so a vault can never sign a tip.
- The submit path picks bundle-vs-single from transaction count alone and throws `JitoMissingTipError` when no tip is present.

So a multi-transaction batch for a vault fails at submission with or without the packing change, and a smaller size budget buys nothing until that is fixed. Everything else in this PR is a plain correctness fix, so it ships on its own.

## Known follow-ups, not in this PR

- **The invariant is unenforced.** Nothing stops a new call site omitting the flag. The right fix is an `ata(mint, owner)` wrapper that always passes it, plus an ESLint restriction on importing `getAssociatedTokenAddressSync` directly, which fixes the class rather than scanning for instances. That means migrating all 30 existing call sites, which does not belong in a fix this size.
- **`positions/create.ts` and `positions/split.ts` attach an ephemeral mint keypair.** A wrapping wallet keeps the message and drops that signature, so a vault gets a proposal whose execution can only fail. These endpoints should reject an off-curve fee payer with a typed error.
- **Fee gates overstate the requirement for a vault**, since `jitoTipCost` is added whenever there is more than one transaction on mainnet.
- **Squads propose mode does not cover governance.** `squadsProposeFields` is spread into the token, data-credit and hotspot schemas only; no governance procedure reads `input.multisig` or calls `buildActionProposal`. `buildActionProposal` is also single-proposal, so a delegation recovery spanning many transactions would need a multi-proposal variant.
- **`pnpm test:unit` is referenced by no workflow**, so the four existing unit tests in this package never run in CI.

### DIFF
--- a/.changeset/multisig-held-wallet-batching.md
+++ b/.changeset/multisig-held-wallet-batching.md
@@ -1,0 +1,9 @@
+---
+"@helium/blockchain-api": patch
+---
+
+Let a multisig-held wallet use the governance endpoints. A Squads vault owns its assets through a program address, and eleven call sites derived associated token addresses without `allowOwnerOffCurve`, so `getAssociatedTokenAddressSync` threw `TokenOwnerOffCurveError` and failed the whole request rather than one derivation.
+
+`buildClaimInstructions` is one of them, and the claim, undelegate and delegate procedures all call it first, so claiming delegation rewards, undelegating, and re-delegating an expired position were unavailable to every vault-held veHNT position that had an unclaimed epoch. `delegate.ts` also derived the reward account this way when enabling autoclaim, so a vault could not turn automation back on.
+
+The derived address is identical either way and the on-chain programs re-validate the account they receive against their own `associated_token::authority` and `token::authority` constraints, so this decides whether a multisig can use an endpoint at all, not who is allowed to do what.
--- a/.github/workflows/blockchain-api-e2e.yml
+++ b/.github/workflows/blockchain-api-e2e.yml
@@ -38,6 +38,30 @@ jobs:
           retention-days: 1
           if-no-files-found: error
 
+  # Unit tests need no chain, no secrets and no surfpool, so they run as their
+  # own job and report in about a minute rather than waiting on the matrix.
+  unit-tests:
+    needs: build-idls
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    defaults:
+      run:
+        working-directory: packages/blockchain-api
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Download anchor types and IDLs
+        uses: actions/download-artifact@v4
+        with:
+          name: anchor-types
+          path: target
+
+      - uses: ./.github/actions/setup-ts/
+
+      - name: Run unit tests
+        run: pnpm test:unit
+
   e2e-tests:
     needs: build-idls
     runs-on: ubuntu-latest
--- a/packages/blockchain-api/src/lib/utils/can-sign.ts
+++ b/packages/blockchain-api/src/lib/utils/can-sign.ts
@@ -1,0 +1,15 @@
+import { PublicKey } from "@solana/web3.js";
+
+/**
+ * Whether `wallet` can sign a transaction itself. An address off the ed25519
+ * curve is a program address, typically a Squads vault: it holds the assets but
+ * has no key, so a transaction built for it is delivered by re-wrapping it into
+ * a multisig proposal that the members approve and execute.
+ *
+ * That re-wrap keeps the transaction's message and discards its signatures, so
+ * an action whose transaction is signed by an ephemeral keypair as it is built
+ * cannot be served to such a wallet.
+ */
+export function canSign(wallet: PublicKey): boolean {
+  return PublicKey.isOnCurve(wallet.toBytes());
+}
--- a/packages/blockchain-api/src/server/api/routers/data-credits/procedures/mint.ts
+++ b/packages/blockchain-api/src/server/api/routers/data-credits/procedures/mint.ts
@@ -63,7 +63,8 @@ export const mint = publicProcedure.dataCredits.mint.handler(
       : new PublicKey(owner);
     const recipientDcAta = getAssociatedTokenAddressSync(
       DC_MINT,
-      recipientPubkey
+      recipientPubkey,
+      true
     );
     const recipientDcAtaInfo = await connection.getAccountInfo(recipientDcAta);
     const ataRent = recipientDcAtaInfo ? 0 : RENT_COSTS.ATA;
--- a/packages/blockchain-api/src/server/api/routers/governance/procedures/delegation/claim-rewards.ts
+++ b/packages/blockchain-api/src/server/api/routers/governance/procedures/delegation/claim-rewards.ts
@@ -131,7 +131,7 @@ export const claimRewards =
 
       // Check which reward ATAs need creation
       const rewardAtaKeys = claimResult.rewardMints.map((mint) =>
-        getAssociatedTokenAddressSync(mint, walletPubkey),
+        getAssociatedTokenAddressSync(mint, walletPubkey, true),
       );
       const rewardAtaAccounts = await connection.getMultipleAccountsInfo(rewardAtaKeys);
       const missingAtaCount = rewardAtaAccounts.filter((a) => !a).length;
--- a/packages/blockchain-api/src/server/api/routers/governance/procedures/delegation/delegate.ts
+++ b/packages/blockchain-api/src/server/api/routers/governance/procedures/delegation/delegate.ts
@@ -551,6 +551,7 @@ export const delegate = publicProcedure.governance.delegatePositions.handler(
                 delegatorAta: getAssociatedTokenAddressSync(
                   HNT_MINT,
                   walletPubkey,
+                  true,
                 ),
                 task,
                 nextTask:
--- a/packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/build-claim-instructions.ts
+++ b/packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/build-claim-instructions.ts
@@ -225,7 +225,8 @@ export async function buildClaimInstructions(
           mint: position.mint,
           positionTokenAccount: getAssociatedTokenAddressSync(
             position.mint,
-            walletPubkey
+            walletPubkey,
+            true
           ),
           positionAuthority: walletPubkey,
           registrar: position.account.registrar,
@@ -254,7 +255,8 @@ export async function buildClaimInstructions(
               delegatorPool: daoAcc.delegatorPool,
               delegatorAta: getAssociatedTokenAddressSync(
                 daoAcc.hntMint,
-                walletPubkey
+                walletPubkey,
+                true
               ),
               delegatorPoolCircuitBreaker: accountWindowedBreakerKey(
                 daoAcc.delegatorPool
@@ -275,7 +277,8 @@ export async function buildClaimInstructions(
               delegatorPool: subDaoAcc.delegatorPool,
               delegatorAta: getAssociatedTokenAddressSync(
                 subDaoAcc.dntMint,
-                walletPubkey
+                walletPubkey,
+                true
               ),
               delegatorPoolCircuitBreaker: accountWindowedBreakerKey(
                 subDaoAcc.delegatorPool
--- a/packages/blockchain-api/src/server/api/routers/governance/procedures/positions/create.ts
+++ b/packages/blockchain-api/src/server/api/routers/governance/procedures/positions/create.ts
@@ -57,6 +57,7 @@ import {
   buildBatchedTransactions,
 } from "../helpers";
 import type { InstructionGroup } from "../helpers";
+import { canSign } from "@/lib/utils/can-sign";
 
 export const create = publicProcedure.governance.createPosition.handler(
   async ({ input, errors }) => {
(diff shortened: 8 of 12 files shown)


packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/pack-groups.ts
```diff
@@ -0,0 +1,145 @@
...
+}): { batches: PackedBatch[]; stoppedEarly: boolean } {
+  const budget = sizeBudget(feePayer);
+
+  // A group carrying its own keypair is signed as the transaction is built. A
+  // payer that cannot sign is served by re-wrapping the message into a multisig
+  // proposal, which keeps the message and drops that signature, and Squads
+  // requires such signers to be declared when the proposal is created. Refuse
+  // rather than hand back a proposal whose execution can only fail.
+  if (!canSign(feePayer) && groups.some((group) => group.signers?.length)) {
+    throw new Error(
```
bryzettler: This deliberate refusal surfaces as an opaque `INTERNAL_SERVER_ERROR` 500 plus a Sentry capture: `positions/create.ts` always passes `signers: [mintKeypair]`, so every off-curve wallet calling `createPosition` deterministically hits this throw, and the v1 route's `onError` treats any non-ORPCError as a 500 and captures it. The client never sees this message, and each attempt pages Sentry for an expected, user-triggerable rejection.

Since procedures already use typed errors for expected rejections, consider throwing a typed error here (or having callers map it to `errors.BAD_REQUEST`).

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/pack-groups.ts
```diff
@@ -0,0 +1,145 @@
...
+        metadata: currentMetadata!,
+        signers: currentSigners,
+      });
+
+      if (batches.length >= maxTxs) {
+        stoppedEarly = true;
+        break;
+      }
+
+      currentInstructions = [...group.instructions];
```
bryzettler: A single instruction that exceeds the budget is only rejected when it comes *first* — carried over mid-stream it's emitted as an over-budget batch with no fit check.

Example with the off-curve budget (818): groups `[G1 (700 bytes), lone-ix G2 (900 bytes)]` → the loop pushes `[G1]` and sets `currentInstructions = [G2]` without measuring it, and the final push emits the 900-byte batch. It serializes fine at build time (`buildOrSplit` only splits on hard >1232 failures), so the vault receives a transaction that can't be wrapped into a Squads proposal. Same input with G2 first throws "Single instruction exceeds max transaction size" — order-dependent.

The structure is inherited from the old inline loop, but the tighter off-curve budget makes it newly reachable. Measuring the carried-over group here (and throwing the same oversize error) closes it; easy to unit-test with the synthetic-instruction helpers this PR adds.

Suggested fix (message stays accurate: after the oversize expansion above, only a single-instruction group can exceed the budget on its own):

```suggestion
      currentInstructions = [...group.instructions];
      let carriedSize: number;
      try {
        carriedSize = measure(currentInstructions);
      } catch {
        carriedSize = MAX_TX_SIZE;
      }
      if (carriedSize > budget) {
        throw new Error(
          `Single instruction exceeds max transaction size (${carriedSize} > ${budget})`,
        );
      }
```

bryzettler (commented):
Review of the off-curve/Squads changes. Line comments below cover what's anchorable in the diff; two issues live in files this PR doesn't touch, so they go here:

**1. `splitPosition` bypasses the new ephemeral-signer guard** (`governance/procedures/positions/split.ts` ~L160)

`split.ts` builds directly via `buildVersionedTransaction({ ..., signers: [newMintKeypair] })` with no `canSign` check and no `packGroups`, so an off-curve (Squads-vault) wallet gets a mint-keypair-signed transaction back. Re-wrapping it into `vault_transaction_create` (ephemeralSigners: 0) keeps the message but drops the mint signature — exactly the failure mode the guard in `pack-groups.ts` refuses: members approve, execution fails on-chain after quorum and rent are spent. The invariant ("refuse rather than hand back a proposal whose execution can only fail") currently only covers the batched path; split should hit the same refusal.

**2. Fee gates still charge `jitoTipCost` for payers that no longer get a tip**

The estimated-fee/INSUFFICIENT_FUNDS gates add `jitoTipCost` whenever `versionedTransactions.length > 1` on mainnet, but with this PR an off-curve payer never receives a tip transaction. A vault whose balance sits between the true requirement and requirement + tip gets a spurious `INSUFFICIENT_FUNDS`, and every off-curve caller sees an overstated estimate. The condition should mirror the tip decision (`canSign(feePayer) && shouldUseJitoBundle(...)`). Affected: `delegation/claim-rewards.ts` (~L126), `delegation/delegate.ts` (~L272/L622), `delegation/undelegate.ts`, `positions/create.ts` (~L373), `proxy/assign.ts` (~L360), `relinquish-position-votes.ts`.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/build-batched-transactions.ts
```diff
@@ -272,7 +172,15 @@ export async function buildBatchedTransactions({
...
   const versionedTransactions = built.map(({ tx }) => tx);
 
-  if (shouldUseJitoBundle(transactions.length, cluster)) {
+  // A Jito bundle lands its transactions together or not at all, which a wallet
+  // that cannot sign can never honour: each transaction becomes a separate
+  // multisig proposal, approved and executed whenever its members get to it. The
+  // tip would also have to be signed by the payer that cannot sign. Such a payer
+  // therefore submits its transactions individually and pays no tip.
+  if (
+    canSign(feePayer) &&
```
bryzettler: The comment's promise that such a payer "submits its transactions individually" is currently an unstated client obligation: `transaction-submission.ts` (~L255) decides bundle-vs-single from transaction count alone, and `requireBundleHasTipAccount` throws `JitoMissingTipError` ("...upgrade your Helium Wallet app...") when no tip account is write-locked. A client that submits this tip-less multi-tx batch in one call gets a misleading hard failure, and nothing in the build response signals "submit these individually".

Either the submit path should learn off-curve payer ⇒ no bundle / sequential submission, or the build response should carry a flag telling clients to submit individually — right now the contract only lives in this comment.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/build-batched-transactions.ts
```diff
@@ -159,104 +150,13 @@ export async function buildBatchedTransactions({
...
-
-    if (packedBatches.length > effectiveMaxTxs) {
-      packedBatches.length = effectiveMaxTxs;
-      stoppedEarly = true;
-    }
-  }
+  const { batches: packedBatches, stoppedEarly } = packGroups({
+    groups,
+    feePayer,
+    maxTxs: effectiveMaxTxs,
```
bryzettler: `effectiveMaxTxs` still reserves one slot for the Jito tip on mainnet even when `canSign(feePayer)` is false and the gate below guarantees no tip is ever appended. A vault with exactly 5 batches gets 4 transactions + `hasMore: true` and pays a second full build round trip for a slot that can never be used. Combined with the smaller off-curve budget, multisig flows pay ~25% extra calls. One-liner: only reserve the slot when the payer can sign.

Suggested fix (`canSign` is already imported; on devnet both branches equal `maxTxs`):

```suggestion
    maxTxs: canSign(feePayer) ? effectiveMaxTxs : maxTxs,
```

packages/blockchain-api/tests/unit/squads-wrap-budget.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  );
+};
+
+/**
+ * Size of the Squads proposal a wrapping wallet has to submit to offer
+ * `instructions` to the multisig's members. A proposal too large to serialize is
+ * reported as Infinity, since that is equally unsubmittable; any other failure
+ * propagates rather than passing for the wrong reason.
+ */
+function wrappedSize(
```
bryzettler: `wrappedSize()` hand-builds the vault_transaction_create + proposal_create pair instead of calling the production `wrapAsVaultProposal()` (`squads/procedures/helpers.ts:247`), so `SQUADS_WRAP_RESERVE` is validated against a private copy of the wrap rather than the wrap the codebase performs. If `wrapAsVaultProposal` changes (memo handling, ephemeralSigners, an extra instruction), this test keeps passing while real proposals overflow 1232 bytes again — the exact regression it exists to catch. Same drift risk for `computeBudget()` (copies the unexported `COMPUTE_BUDGET_PLACEHOLDERS`) and `isOversize()` (copies `buildOrSplit`'s error-string matching). Worth exporting the real helpers and testing against them.

packages/blockchain-api/tests/unit/ata-off-curve-owner.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    expect(tracked.length).to.be.greaterThan(0);
+
+    const offenders: string[] = [];
+    let scanned = 0;
+    for (const file of tracked) {
+      const source = readFileSync(path.join(packageRoot, file), "utf8");
+      for (const { line, args } of calls(source)) {
+        scanned++;
+        // The value is the point, not the arity: `false` reinstates the throw.
+        if (args[2] !== "true") {
```
bryzettler: This scanner enforces the invariant at the wrong altitude: it requires the literal token `true` as the third argument of a literal `getAssociatedTokenAddressSync(` call. An aliased import (`import { getAssociatedTokenAddressSync as getAta }`), the async `getAssociatedTokenAddress` (throws the same `TokenOwnerOffCurveError`), or a correct boolean variable either bypasses the scan (off-curve 500s return in prod) or fails CI on correct code. The parser is also string/comment-unaware, so a paren or comma inside a string argument desyncs the depth counter.

A small `ata()` wrapper that always passes `allowOwnerOffCurve: true`, plus banning direct imports via the package's existing ESLint setup, would fix the class instead of scanning for instances.

madninja:
Cut this back to the associated-token fix only; the packing and Jito changes are gone (`git diff origin/develop` is now 8 files, +29/-7).

The packing half was dropped because its benefit is blocked downstream: `getJitoTipTransaction` builds the transfer *from* the fee payer, so a vault can never sign a tip, and the submit path picks bundle-vs-single from transaction count alone and throws `JitoMissingTipError` with no tip present. A multi-transaction batch for a vault therefore fails at submission with or without a corrected size budget.

That resolves the review comments on `pack-groups.ts` (order-dependent oversize check, untyped refusal, untested signer accumulation), `effectiveMaxTxs`, the Jito submit-path contract, and the wrap-fixture drift, by removing the code they applied to.

Two comments survive as follow-ups recorded in the description: the `ata()` wrapper plus an ESLint ban is the right shape for enforcing this (agreed, and it needs all 30 call sites, which is too much for a fix this size), and `split.ts`/`create.ts` should reject an off-curve fee payer with a typed error since they attach an ephemeral mint keypair. The fee-gate `jitoTipCost` point no longer applies here since the tip behaviour is unchanged.

madninja:
Status of each review comment, after cutting the packing half and adding the typed refusal.

**Addressed**

- `positions/split.ts` / `positions/create.ts` ephemeral mint keypair — both now refuse a fee payer that cannot sign with a typed `BAD_REQUEST`. Worth flagging that you were more right than the comment stated: `create.ts`'s unconditional deposit-balance check derived an ATA that threw for an off-curve owner, so createPosition was stopped *by accident* before this PR. Fixing that derivation removes the accidental stop and makes the unexecutable path reachable, so this was a defect the reduced diff created rather than a pre-existing one. `split.ts` had no such accidental stop and was reachable already.
- `pack-groups.ts:54` untyped refusal causing a 500 plus a Sentry capture — the refusal is now a typed error at the procedure boundary instead.

**Resolved by removing the code they applied to** (verified: `pack-groups.ts` and both test files are gone, and `build-batched-transactions.ts` / `constants.ts` are byte-identical to develop)

- `pack-groups.ts:130` order-dependent oversize check. Note this structure still exists in the inline loop on develop; your comment said the tighter off-curve budget made it *newly* reachable, and that budget is gone, so it reverts to latent rather than fixed.
- `build-batched-transactions.ts:156` `effectiveMaxTxs` reserving a tip slot.
- `build-batched-transactions.ts:181` the submit-path contract living only in a comment.
- `squads-wrap-budget.test.ts:80` validating against a private copy of the wrap instead of `wrapAsVaultProposal`.
- Fee gates charging `jitoTipCost` — the tip behaviour is unchanged now, so the gates match it again.

**Not addressed, recorded in the description**

- `ata-off-curve-owner.test.ts:83` — you are right that scanning for instances is the wrong altitude, and I deleted that test rather than keep a guard that passes on `allowOwnerOffCurve: false`. I attempted the `ata()` wrapper plus an ESLint ban and backed out: it needs all 30 call sites migrated with import surgery, which is too much churn for a fix this size and too easy to get subtly wrong mechanically. So the invariant is currently unenforced, and that is stated in the description rather than papered over.

**Coverage honesty:** the only new test is on `canSign` itself. Mutating the predicate turns the suite red; mutating the guard away at the `create.ts` call site does not, because nothing reaches those handlers without env and a connection. The call sites rest on review, not tests.

bryzettler (approved):

PR #1233, opened 2026-08-03
HIP 149: activate the supplement mint and forward its destinations

madninja:
Activates HIP 149 Decision 2 and wires Decision 4's destination through the cron.

**Not deployable as it stands.** `SUPPLEMENT_VAULT_TOKEN_ACCOUNT` and
`COUNCIL_FANOUT_TOKEN_ACCOUNT` are still the placeholder, and a window that opens against a
placeholder mints nothing and fails every epoch. Both get their real values, in this branch,
before it leaves draft.

## Mint start

`INFLATION_START = 1785931200` (2026-08-05T12:00:00Z), so the first mint lands at the
2026-08-06T00:00:00Z epoch crank. The value sits between two crank boundaries rather than on
one, which fixes which epoch mints first regardless of clock drift or a late crank, and makes
the flat window exactly 360 cranks (through 2027-08-01) and the taper exactly 720.

Under `TESTING` the start stays at the dormant sentinel. Once a window is open both destination
accounts are mandatory in `issue_rewards_v0`, and every localnet test closes epochs without
them. The cost of that gate is that no integration test exercises the active window, including
the supply-tracking correction in `calculate_utility_score_v0`; removing it means reworking the
exact-`currentHntSupply` assertions in `tests/helium-sub-daos.ts`, which belongs in its own PR.

## Deploy order

`queue_end_epoch` forwards both destinations, read from the `helium-sub-daos` constants so the
two programs cannot disagree about where the supplement goes. A task is queued an epoch before
it runs, so window state at queue time says nothing about window state at execution: withholding
the accounts until a window is observed open would make the first epoch of a window fail closed
and stop the self-rescheduling chain. A placeholder constant is withheld instead, since Anchor
deserializes a forwarded account as a `TokenAccount` whether or not the window is open.

So the two mint destinations and `INFLATION_START` have to be real in one deploy, and both
programs ship together. A `helium-sub-daos` carrying a real start alongside an `hpl-crons` that
forwards nothing halts every epoch from the first crank of the window, which halts all rewards.
Already-queued tasks were compiled without the two accounts, so the cron needs the same
re-bootstrap the backstop rollout used.

`end-epoch` takes both accounts as flags, so the permissionless manual fallback still closes an
epoch inside a window. It passed nulls, which fails once the supplement is live. Each is pinned
on chain to its constant, so a wrong value fails the instruction rather than misrouting a mint.

## Coverage

Mutation-tested, 8 enumerated and 8 measured. Caught: inverting the placeholder guard, forwarding
unconditionally, swapping the `TESTING` branches, moving the start a day, a 359-epoch flat window,
and a 250bp Council cut.

Not caught, both at the call site: `supplement_vault: None` and `council_vault: None` in
`queue_end_epoch`. No test can close these, because the destinations are compile-time constants
that are deliberately placeholders in a `TESTING` build, so absence is the correct expectation
there and an assertion would pass either way. The check moved to the deploy runbook instead:
decode the queued `end epoch` task before the boundary and confirm the `issue_rewards_v0`
instruction carries both accounts.

### DIFF
--- a/packages/helium-admin-cli/src/create-council-fanout.ts
+++ b/packages/helium-admin-cli/src/create-council-fanout.ts
@@ -153,6 +153,26 @@ export async function run(args: any = process.argv) {
   );
 
   // Schedule the first distribution task; the fanout self-reschedules thereafter.
+  //
+  // `schedule_task_v0` declares `has_one = task_queue / next_task / next_pre_task` on the
+  // fanout, and Anchor's client resolver satisfies those by fetching it. The instruction
+  // above creates the fanout in this same transaction, so there is nothing on chain to fetch
+  // and resolution fails before anything is sent. Each of those accounts is known ahead of
+  // time instead: `initialize_mini_fanout_v0` stores the task queue it was handed, and sets
+  // `next_task` and `next_pre_task` to the fanout's own key, the "no next task" sentinel that
+  // the scheduling constraint accepts.
+  const [queueAuthority] = PublicKey.findProgramAddressSync(
+    [Buffer.from("queue_authority", "utf-8")],
+    program.programId
+  );
+  const [taskQueueAuthority] = PublicKey.findProgramAddressSync(
+    [
+      Buffer.from("task_queue_authority", "utf-8"),
+      TASK_QUEUE_ID.toBuffer(),
+      queueAuthority.toBuffer(),
+    ],
+    tuktukProgram.programId
+  );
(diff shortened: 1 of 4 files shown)


madninja:
Superseded by the stack: #1235 (measurement harness, pre-sizing, CLI fixes) and #1236 (activation constants, stacked on it). Same content, split so the heap work can land on its own, and rebuilt on signed commits.

PR #1235, opened 2026-08-03
HIP 149: activate the supplement mint, and make the end-epoch heap cost measurable

madninja:
Activates the HIP 149 operations-and-growth supplement, and fixes the 32KB heap problem that
activating it exposed.

## Activation

`INFLATION_START = 1785931200` (2026-08-05T12:00:00Z), so the first supplement mint lands at the
2026-08-06T00:00:00Z epoch crank. The value sits between two crank boundaries rather than on one, so
neither clock drift nor a late crank changes which epoch mints first, and the windows come to exactly
360 and 720 cranks. Under `TESTING` it stays at the dormant sentinel, because an open window makes
both destination accounts mandatory and every localnet test closes epochs without them.

| | Address |
|---|---|
| Receiving Entity vault (Squads v4 `2g7qF5d3…`, index 0) | `Cuy6WiVdHE6Wznqos86AUwLpA4zPfPABYq3qszMa6E6A` |
| `SUPPLEMENT_VAULT_TOKEN_ACCOUNT` | `AGPDcgpXan5RB2Y9usHvdJmmJHpyyYqcQm8KouRkK6f4` |
| Council fanout (5 seated members, equal shares) | `2pCpQYrUmQor9igsF4cY7tkmd3jBQzHFEB75gtpjdWA2` |
| `COUNCIL_FANOUT_TOKEN_ACCOUNT` | `EYbAXgLq1aRr9a9y55DbjfMdXNrZuMa69WXN9c1UR6eK` |

Both exist on chain. Each is derived in a test rather than restated: the vault one from the vault and
the HNT mint, the Council one through the whole chain that produced it — mini_fanout program id,
creating wallet as namespace, seed, fanout PDA, then its associated token account. Repointing either
means changing one of those inputs deliberately, and mutating the address or the namespace fails the
suite.

The deployed rate and window are pinned the same way, against the amounts and dates HIP 149
publishes rather than against the constants themselves: 98,000 HNT per sub-DAO per epoch, the
196,000 HNT/day total, the taper opening at the flat rate, 360 flat cranks then 720 tapering, and
which crank is the first to mint. The two window ends are `INFLATION_START + N * EPOCH_LENGTH`, so
what is pinned is `N` and the rate each crank sees; restating the derived timestamps would only
duplicate arithmetic the compiler already does.

`hpl-crons` forwards both destinations into the queued `issue_rewards_v0`. A task is queued an epoch
before it runs, so window state at queue time says nothing about window state at execution: were the
accounts withheld until a window was observed open, the first epoch of that window would settle
against a task built without them and fail closed, leaving that epoch to be settled by hand through
the `end-epoch` CLI. (The reschedule is a separate task on the same trigger, so the self-rescheduling
chain survives a failed settlement; what does not survive is `queue_end_epoch` itself failing, which
is what the heap work below is about.) A placeholder constant is withheld instead, since Anchor
deserializes a forwarded account as a `TokenAccount` regardless of window state.

## The heap, and why this nearly shipped broken

`queue_end_epoch` runs at CPI depth 2 under `run_task_v0`, sharing one 32KB heap with it, and it has
exhausted that heap in production once already (#1215). Forwarding the two destinations costs **136
bytes**. That was enough to overflow it, because the task was already within roughly that much of the
limit. The localnet regression test caught it before deploy; deployed, it would have failed the
end-epoch cron every epoch, which halts all rewards.

The only signal for how close the task sat was that localnet test, which needs an anchor build and a
running validator to return a boolean. So a unit test now measures it. The instruction set builds
from plain values instead of the Anchor context, and a thread-local tally wraps `System`, counting
allocations and ignoring frees — `BumpAllocator::dealloc` is literally `// I'm a bump allocator, I
don't free`, so a program is bounded by bytes ever allocated, not peak live. It warms up first,
because the first measured region on a thread also captures ~1.5KB of one-time initialisation.

Having a number made the fixes findable, and there are two.

The larger one is the compile itself. `tuktuk_program::compile_transaction` builds two HashMaps that
grow from empty by doubling, plus a `remaining_accounts` Vec this caller discards; on a bump
allocator every intermediate table stays resident for the life of the instruction. Compiling the
same instructions with buffers sized from the known slot count, and a linear scan in place of the
hashing, saves **~8,600 bytes**. At this scale, ~73 slots over ~31 distinct accounts, the scan is
cheaper than the allocations it avoids, and `tests/hpl-crons.ts` now asserts the compute side stays
clear of the crank turner's limit so that trade cannot silently invert.

The smaller one: each calculate instruction appends two Mobile epoch infos to the Vec
`to_account_metas` returns at exact capacity, so every push reallocs and copies the whole thing.
Pre-sizing both returns **1,020 bytes**, 7.5x what the destinations cost.

| | Bytes |
|---|---|
| develop today | 24,464 |
| + pre-sizing the account-metas Vec | 23,444 |
| + carrying both destinations | 24,104 |
| **this branch, also replacing `compile_transaction`** | **14,968** |

The handler allocates ~9.5KB less while carrying the destinations than develop does without them.
The 14,968 figure is what the unit test asserts, against a 16,000 budget. The localnet heap test
passes.

Two caveats on that figure: it is a fraction of the real budget, since `run_task_v0` spends part of
the shared heap before this handler runs, and it is measured with the host allocator rather than the
SBF one, so it detects drift rather than asserting 32KB. The localnet test stays the ground truth.

Things that did **not** help, measured rather than assumed: splitting the task in one invocation
costs +1,219 bytes because each compiled transaction carries its own copy of the shared account set;
moving `no_emit` out costs +4,031; dropping a value early reclaims nothing on a bump allocator; and
rebuilding the signer seeds instead of cloning them is a wash at 158 bytes either way. Address lookup
tables do not apply — `CompiledTransactionV0.accounts` stores full pubkeys, and tuktuk's
`lookup_tables` serve the crank turner's outer transaction against the 1232-byte size cap, a
different limit.

Indices and the privilege counts in a compiled transaction are `u8`. Past 255 accounts they would
truncate silently and point instructions at the wrong accounts rather than failing, so the compile
refuses that case, ahead of the counting loop so the counters cannot overflow first. The end-epoch
set uses 32.

Getting the privilege boundaries wrong makes a task unexecutable rather than over-privileged:
`run_task_v0` derives signer privilege from `signer_seeds` under a forced `[b"custom", task_queue]`
prefix and never from these counts, and writability is capped by the crank turner's outer
transaction, which it builds from these same counts. The failure mode is a stuck epoch, which is
what `presized_compile_matches_upstream` guards, across three shapes: the two the handler compiles,
plus one that passes an account read-only before passing it as a writable signer, since neither real
shape exercises the signer half of the cross-instruction privilege accumulation.

## CLI

`create-council-fanout` could not create a fanout at all: `schedule_task_v0` declares `has_one` on
the fanout for `task_queue`, `next_task` and `next_pre_task`, and Anchor's resolver satisfies those by
fetching an account the previous instruction creates in the same transaction, so resolution exhausted
its depth before anything was sent. Those accounts are known ahead of time and are now passed. It also
passes the payer as the rent refund, since `initialize_mini_fanout_v0` stores the payer and ignores
the account given under that name.

`end-epoch` takes both destinations as flags, so the permissionless fallback can still close an epoch
inside a supplement window. Each is pinned on chain to its constant, so a wrong value fails the
instruction rather than misrouting a mint.

## Scope of the pinned addresses

`COUNCIL_FANOUT_TOKEN_ACCOUNT` fixes where the carve-out is minted, not where it ends up. The
fanout's `owner` is the governance multisig vault and never the Receiving Entity, so the seated-member
set changes only through a multisig action; but a seated member can point their own share at a
delegate under their own signature, and the owner can close the fanout, sweeping its balance and
freeing the PDA for re-initialisation by the namespace keypair. Monitoring belongs on the fanout's
shares, not only on the address.

## Deploy

Both programs ship together: a `helium-sub-daos` with a real start alongside an `hpl-crons` that
forwards nothing halts every epoch of the window. Already-queued tasks were compiled without the two
accounts, so the cron needs the same re-bootstrap the backstop rollout used. Before the boundary,
decode the queued `end epoch` task and confirm the `issue_rewards_v0` instruction carries both
accounts.

Preconditions to confirm at deploy, both outside this diff: the HNT mint circuit breaker
(`3QqEkX1juz4bW8mCQBMJLWjEpHqYawvbEUvpc6iwYom3`) is Absolute at 350,000 HNT/day, which clears the
196,000 supplement plus current draw; and HIP 149 sets mint start at Council seating + 2 weeks, with
any delay extending it equally, so `INFLATION_START` encodes a seating date on or before
2026-07-22.

### DIFF
--- a/packages/helium-admin-cli/src/create-council-fanout.ts
+++ b/packages/helium-admin-cli/src/create-council-fanout.ts
@@ -16,6 +16,7 @@ import {
 import * as multisig from "@sqds/multisig";
 import os from "os";
 import yargs from "yargs/yargs";
+import { COUNCIL_FANOUT_TOKEN_ACCOUNT } from "./hip149";
 import { loadKeypair } from "./utils";
 
 // HIP 149 Decision 4: create the Advisory Council compensation fanout. The supplement mint in
@@ -151,6 +152,21 @@ export async function run(args: any = process.argv) {
     .prepare();
   instructions.push(initIx);
 
+  // The fanout PDA is seeded by the `namespace` signer and `--seed`. Anchor resolves that
+  // signer to the provider wallet, so a different wallet, or a different seed, silently
+  // produces a different PDA, a different HNT associated token account, and a fanout that
+  // `issue_rewards_v0` rejects on its account pin. That would fail every epoch close inside a
+  // supplement window. Members, owner and schedule do not enter the derivation, so this only
+  // constrains the two inputs that have to stay fixed. Checked before anything is sent.
+  const tokenAccount = getAssociatedTokenAddressSync(HNT_MINT, miniFanout!, true);
+  if (!tokenAccount.equals(COUNCIL_FANOUT_TOKEN_ACCOUNT)) {
+    throw new Error(
+      `Derived fanout token account ${tokenAccount.toBase58()} does not match the ` +
+        `COUNCIL_FANOUT_TOKEN_ACCOUNT pinned in issue_rewards_v0 ` +
+        `(${COUNCIL_FANOUT_TOKEN_ACCOUNT.toBase58()}). Wrong signing wallet or wrong --seed.`
+    );
+  }
+
   // Fund the fanout so it can pay its own crank reward and stay scheduled.
   instructions.push(
     SystemProgram.transfer({
@@ -206,12 +222,6 @@ export async function run(args: any = process.argv) {
     computeUnitLimit: 500000,
   });
 
-  const tokenAccount = getAssociatedTokenAddressSync(
-    HNT_MINT,
-    miniFanout!,
-    true
-  );
-
   console.log("Council fanout created.");
   if (argv.multisig) {
     console.log(`  multisig:                    ${argv.multisig}`);
@@ -223,7 +233,7 @@ export async function run(args: any = process.argv) {
   members.forEach((m) => console.log(`    - ${m.toBase58()}`));
   console.log("");
   console.log(
-    "  >>> Council fanout token account (set COUNCIL_FANOUT_TOKEN_ACCOUNT):"
+    "  >>> Council fanout token account (matches COUNCIL_FANOUT_TOKEN_ACCOUNT):"
   );
   console.log(`      ${tokenAccount.toBase58()}`);
 }
--- a/packages/helium-admin-cli/src/create-council-fanout.ts
+++ b/packages/helium-admin-cli/src/create-council-fanout.ts
@@ -37,6 +37,9 @@ import { loadKeypair } from "./utils";
 //     members from that point forward, per the HIP).
 //   - Keep the fanout funded with lamports; it pays its own crank reward and silently
 //     unschedules itself if it runs dry.
+//   - Closing the fanout refunds its rent to whoever paid to create it, not to the owner:
+//     initialize_mini_fanout_v0 stores `ctx.accounts.payer.key()` as `rent_refund` and never
+//     reads the account passed under that name. Pick the payer accordingly.
 export async function run(args: any = process.argv) {
   const yarg = yargs(args).options({
     wallet: {
@@ -137,7 +140,12 @@ export async function run(args: any = process.argv) {
       payer: wallet.publicKey,
       owner,
       taskQueue: TASK_QUEUE_ID,
-      rentRefund: owner,
+      // The account required here does not decide where rent goes:
+      // initialize_mini_fanout_v0 stores `ctx.accounts.payer.key()` and never reads this one.
+      // Passing the payer keeps the argument honest about the refund destination, and keeps a
+      // vault that has no other role in this instruction out of the writable set. Change the
+      // payer to change where closing the fanout refunds its rent.
+      rentRefund: wallet.publicKey,
       mint: HNT_MINT,
     })
     .prepare();
@@ -153,6 +161,26 @@ export async function run(args: any = process.argv) {
   );
 
   // Schedule the first distribution task; the fanout self-reschedules thereafter.
+  //
+  // `schedule_task_v0` declares `has_one = task_queue / next_task / next_pre_task` on the
+  // fanout, and Anchor's client resolver satisfies those by fetching it. The instruction
+  // above creates the fanout in this same transaction, so there is nothing on chain to fetch
+  // and resolution fails before anything is sent. Each of those accounts is known ahead of
+  // time instead: `initialize_mini_fanout_v0` stores the task queue it was handed, and sets
+  // `next_task` and `next_pre_task` to the fanout's own key, the "no next task" sentinel that
+  // the scheduling constraint accepts.
+  const [queueAuthority] = PublicKey.findProgramAddressSync(
+    [Buffer.from("queue_authority", "utf-8")],
+    program.programId
+  );
+  const [taskQueueAuthority] = PublicKey.findProgramAddressSync(
+    [
+      Buffer.from("task_queue_authority", "utf-8"),
+      TASK_QUEUE_ID.toBuffer(),
+      queueAuthority.toBuffer(),
+    ],
+    tuktukProgram.programId
+  );
   const taskQueueAcc = await tuktukProgram.account.taskQueueV0.fetch(
     TASK_QUEUE_ID
   );
@@ -160,9 +188,14 @@ export async function run(args: any = process.argv) {
   instructions.push(
     await program.methods
       .scheduleTaskV0({ taskId, preTaskId })
-      .accounts({
+      .accountsPartial({
         payer: wallet.publicKey,
         miniFanout: miniFanout!,
+        taskQueue: TASK_QUEUE_ID,
+        queueAuthority,
+        taskQueueAuthority,
+        nextTask: miniFanout!,
+        nextPreTask: miniFanout!,
         task: taskKey(TASK_QUEUE_ID, taskId)[0],
         preTask: taskKey(TASK_QUEUE_ID, preTaskId)[0],
       })
--- a/packages/helium-admin-cli/src/end-epoch.ts
+++ b/packages/helium-admin-cli/src/end-epoch.ts
@@ -76,6 +76,16 @@ export async function run(args: any = process.argv) {
         type: "number",
         describe: "The timestamp to start ending epochs from",
       },
+      supplementVault: {
+        type: "string",
+        describe:
+          "HIP 149 Decision 2 Receiving Entity vault HNT token account. Required to close an epoch inside a supplement window; issue_rewards_v0 pins it to SUPPLEMENT_VAULT_TOKEN_ACCOUNT, so a wrong value fails the instruction rather than misrouting the mint. Omit while the supplement is dormant.",
+      },
+      councilVault: {
+        type: "string",
+        describe:
+          "HIP 149 Decision 4 Council fanout HNT token account. Same requirement and same on-chain pinning as --supplementVault.",
+      },
     });
     const argv = await yarg.argv;
     process.env.ANCHOR_WALLET = argv.wallet;
@@ -89,6 +99,12 @@ export async function run(args: any = process.argv) {
     const unixNow = new Date().valueOf() / 1000;
     const [dao] = daoKey(hntMint);
     const basePriorityFee = Number(argv.basePriorityFee || 1);
+    const supplementVault = argv.supplementVault
+      ? new PublicKey(argv.supplementVault)
+      : null;
+    const councilVault = argv.councilVault
+      ? new PublicKey(argv.councilVault)
+      : null;
 
     const subDaos = await heliumSubDaosProgram.account.subDaoV0.all([
       {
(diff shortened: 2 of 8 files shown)


programs/helium-sub-daos/src/supplement.rs
```diff
@@ -10,22 +10,35 @@
...
+/// Mint start, 2026-08-05T12:00:00Z. Epochs settle on the 00:00 UTC boundary, so the first
+/// supplement mint lands at the 2026-08-06T00:00:00Z crank, closing the Council's pre-mint
+/// review window. The value sits between two crank boundaries rather than on one, so neither
+/// Solana clock drift nor a late crank can change which epoch mints first, and the windows
+/// come to whole crank counts: 360 flat, then 720 tapering.
+///
+/// Under TESTING the start stays a far-future sentinel (2100-01-01), keeping the supplement
+/// dormant for the localnet suite: an open window makes both destination accounts mandatory
+/// in `issue_rewards_v0`, and every test there closes epochs without them.
+pub const INFLATION_START: i64 = if TESTING {
```
bryzettler, line 51: **Devnet builds get the live mainnet schedule.** Devnet deploys build with `testing: false, devnet: true` (see `develop-release-program.yaml` / `manual-devnet-deploy.yaml`), so `TESTING` is unset and devnet gets `INFLATION_START = 2026-08-05T12:00Z` plus the two mainnet ATA pins. Once the window opens, `issue_rewards_v0` on devnet will require both destination accounts, which are mainnet-mint ATAs that can't exist on devnet — devnet epochs will stop closing after 2026-08-06 unless devnet-equivalent accounts are set up. If devnet should stay dormant like localnet, gating on the `devnet` feature (same pattern as the other programs' `#[cfg(feature = "devnet")]` constants) would do it:

```rust
#[cfg(feature = "devnet")]
pub const INFLATION_START: i64 = 4_102_444_800;
#[cfg(not(feature = "devnet"))]
pub const INFLATION_START: i64 = if TESTING {
  4_102_444_800
} else {
  1_785_931_200
};
```

packages/helium-admin-cli/src/end-epoch.ts
```diff
@@ -89,6 +99,12 @@ export async function run(args: any = process.argv) {
     const unixNow = new Date().valueOf() / 1000;
     const [dao] = daoKey(hntMint);
     const basePriorityFee = Number(argv.basePriorityFee || 1);
+    const supplementVault = argv.supplementVault
+      ? new PublicKey(argv.supplementVault)
+      : null;
```
bryzettler: Two operational hazards with these flags:

1. **Backfill overshoot.** `issue_rewards_v0` computes the supplement from `Clock::get()` at execution time, not from the epoch being settled. If the cron falls behind and this script backfills several old epochs while the window is open, each backfilled epoch mints the full per-epoch supplement — N epochs behind means N extra supplements minted in one day. The taper/end boundaries also shift in wall-clock terms: cranks delayed past a boundary mint at the later rate. Probably acceptable given HIP 149's ‘360 cranks’ framing, but worth a comment here so an operator backfilling during the window knows the supplement follows crank time, not epoch time.

2. **Silent omission.** These default to `null`, and a `null` vault during an open window makes `issue_rewards_v0` fail — which this script swallows into `errors[]` while the loop advances. Since the correct values are compile-time constants (`SUPPLEMENT_VAULT_TOKEN_ACCOUNT` / `COUNCIL_FANOUT_TOKEN_ACCOUNT`) and the program pins them anyway (a wrong value can only fail, never misroute), consider defaulting the flags to those constants so the cron path can't forget them.

tests/hpl-crons.ts
```diff
@@ -189,6 +189,25 @@ describe("hpl-crons", () => {
...
+    // queue_end_epoch trades allocation for computation: it dedupes and indexes accounts by
+    // linear scan rather than by hashing, because the 32KB heap is the scarcer budget and a
+    // bump allocator never reclaims what a HashMap's growth chain leaves behind. That trade is
+    // only safe while the compute side stays clear of its own ceiling, so measure it here
+    // rather than reasoning about it: the crank turner sets the limit, and a regression that
+    // pushes past it fails the epoch exactly as an overflow would.
+    const executed = await provider.connection.getTransaction(sig, {
+      commitment: "confirmed",
+      maxSupportedTransactionVersion: 0,
+    });
```
bryzettler: `getTransaction` immediately after a confirmed send can return `null` on a busy RPC (the transaction is confirmed but not yet queryable), which would fail this assertion flakily rather than for a real CU regression. A short retry makes the measurement robust:

```suggestion
    let executed: Awaited<
      ReturnType<typeof provider.connection.getTransaction>
    > = null;
    for (let i = 0; i < 10 && !executed; i++) {
      executed = await provider.connection.getTransaction(sig, {
        commitment: "confirmed",
        maxSupportedTransactionVersion: 0,
      });
      if (!executed) await new Promise((r) => setTimeout(r, 500));
    }
```

packages/helium-admin-cli/src/create-council-fanout.ts
```diff
@@ -153,16 +161,41 @@ export async function run(args: any = process.argv) {
...
     TASK_QUEUE_ID
   );
   const [taskId, preTaskId] = nextAvailableTaskIds(taskQueueAcc.taskBitmap, 2);
   instructions.push(
     await program.methods
       .scheduleTaskV0({ taskId, preTaskId })
-      .accounts({
+      .accountsPartial({
         payer: wallet.publicKey,
         miniFanout: miniFanout!,
```
bryzettler, line 209: **The script never asserts the derived fanout token account equals the pinned `COUNCIL_FANOUT_TOKEN_ACCOUNT`.** The fanout PDA is seeded by the `namespace` signer, which silently defaults to the provider wallet when `--namespace` isn't passed — run it with the wrong wallet and you get a fully-functional fanout whose ATA doesn't match the constant baked into `issue_rewards_v0`, and every supplement-window epoch close then fails on the account pin. Since the expected address is a known constant, an early assert turns that into an immediate, obvious failure:

```ts
const EXPECTED_COUNCIL_FANOUT_TOKEN_ACCOUNT = new PublicKey(
  "EYbAXgLq1aRr9a9y55DbjfMdXNrZuMa69WXN9c1UR6eK"
);
if (!tokenAccount.equals(EXPECTED_COUNCIL_FANOUT_TOKEN_ACCOUNT)) {
  throw new Error(
    `Derived fanout token account ${tokenAccount.toBase58()} does not match the address pinned in issue_rewards_v0. Wrong --namespace?`
  );
}
```

bryzettler (commented):
Ran a deep review of this branch vs `develop` (multi-agent pass over the heap fix, the HIP 149 activation, and the CLI changes). The pre-sized compile, its budget test, and the u8 bound all check out — the comments below are the findings I think are worth addressing before the 2026-08-05 activation, roughly in priority order. Nice touch moving the `TooManyAccounts` guard ahead of the tally in the latest push.

bryzettler (approved):

programs/helium-sub-daos/src/supplement.rs
madninja: Confirmed and fixed in 19872791b. Verified the chain rather than just the claim: both devnet workflows pass `testing: false, devnet: true`, and `build-anchor` only sets the env var when `testing` is true, so `TESTING` was unset and devnet compiled the live start.

`INFLATION_START` now has a `devnet` arm, using the `#[cfg(feature = "devnet")]` pattern from voter-stake-registry, hpl-crons and dc-auto-top. Gating the start alone is sufficient: with both windows closed the destinations are never required, so they need no gating of their own.

Mutation-tested in both directions and with the gate deleted, all three caught by `deployed_window_boundaries_match_the_published_schedule`, which now asserts the dormant arm for `TESTING || cfg!(feature = "devnet")`. Suite runs green on both feature arms.

packages/helium-admin-cli/src/end-epoch.ts
madninja: Both parts confirmed and fixed in 19872791b.

Silent omission: the flags now default to the pinned constants, the only values that can succeed while a window is open. Passing the system program id withholds one deliberately. Confirmed the failure mode you described: `mainLoop` collects each `issueRewardsV0` failure into `errors` and advances.

Backfill overshoot: comment added at the point the vaults are resolved. One bound worth recording alongside it, since it caps the exposure: at a 350,000 HNT/day absolute threshold the HNT mint circuit breaker admits roughly one supplement per day on top of normal emissions, so a multi-epoch backfill fails on the breaker rather than overminting without limit.

tests/hpl-crons.ts
madninja: Applied in 19872791b, with a note on why the retry is there rather than a bare retry. The anchor job passes on the new head.

packages/helium-admin-cli/src/create-council-fanout.ts
madninja: Confirmed and fixed in 19872791b, and it is slightly worse than described: there is no `--namespace` flag at all, so Anchor resolves that signer to the provider wallet unconditionally, and the token account was derived only after `sendInstructionsWithPriorityFee` had already run.

The check now happens before anything is sent. Members, owner and schedule do not enter the PDA derivation, so this constrains only the signing wallet and `--seed`, the two inputs that have to stay fixed. The address lives in one module alongside the `end-epoch` default rather than being written twice.

PR #1248, opened 2026-08-09
Constrain a mini-fanout pre task to one of two shapes

madninja:
A `pre_task` stored on a `MiniFanoutV0` is one of two shapes: a remote transaction the Helium
oracle signs, or a compiled transaction carrying no signer seeds. Anything else fails with a new
`InvalidPreTask` error (6011, appended).

The rule is applied where a pre task is queued — `schedule_task_v0::schedule_impl` and
`distribute_v0` — both through `MiniFanoutV0::pre_task_to_queue`, so it has a single home.

All 47 live fanouts already satisfy it: 38 remote on the oracle signer, 8 compiled with no seeds,
1 none. `welcome-pack` hardcodes `ORACLE_SIGNER` when it CPIs `initialize_mini_fanout_v0`, and no
TypeScript path builds a compiled pre task.

### `devnet` feature forwarding

`mini-fanout`'s `devnet` feature did not forward to `shared-utils/devnet`, which `ORACLE_SIGNER`
is gated on. This is the first code in the program to read that constant, so a devnet build would
have resolved the mainnet key. Measured: `cargo tree -p mini-fanout --features devnet` gave
`shared-utils FEATURES=[default]`, where `hpl-crons` and `welcome-pack` both give
`[default,devnet]` — they are the only other programs reading it, and both already forward.

Devnet has 15 fanouts, 12 with a remote pre task: 7 on the devnet signer and 5 on the mainnet
signer, so some devnet fanouts will stop scheduling either way. `close_mini_fanout_v0` does not
route through the check, so owners retain an exit. Mainnet is unaffected.

### Test fixture

The suite was passing a pre task that never reached the chain intact, so the pre-task path was
untested. `CompiledV0` is a tuple variant, so `{compiledV0: tx}` encodes an **empty** transaction
rather than the memo; and `compileTransaction` returns the account list separately for
`queue_task_v0` to merge, so a pre task stored on the fanout has to carry its own accounts. With
both corrected the memo pre task actually runs, which is what makes `distribute_v0`'s
`PreTaskNotRun` gate mean anything.

`tests/hpl-crons.ts` and `start-cron.ts` already use the array form, so mini-fanout was the only
offender.

mini-fanout 0.1.3 → 0.1.4.

### DIFF
--- a/packages/blockchain-api-client/src/contracts/reward-contract.ts
+++ b/packages/blockchain-api-client/src/contracts/reward-contract.ts
@@ -104,7 +104,7 @@ export const rewardContract = oc
                 z.object({
                     entityPubKey: HeliumPublicKeySchema.describe("The public key of the hotspot entity"),
                     signerWalletAddress: WalletAddressSchema.describe("The wallet address of the caller"),
-                    expirationDays: z.number().int().positive().max(365).default(7).describe("Number of days until the invite expires"),
+                    expirationDays: z.number().int().positive().max(14).default(3).describe("Number of days until the invite expires"),
                 })
             )
             .errors({
--- a/.changeset/mini-fanout-pre-task-shape.md
+++ b/.changeset/mini-fanout-pre-task-shape.md
@@ -2,6 +2,6 @@
 "@helium/idls": patch
 ---
 
-`mini-fanout` holds a pre task to one of two shapes — a remote transaction the Helium oracle
-signs, or a compiled transaction carrying no signer seeds — and gains an `InvalidPreTask` error
-for anything else.
+`mini-fanout` queues a pre task only when it is one of two shapes: a remote transaction the
+Helium oracle signs, or a compiled transaction carrying no signer seeds. Anything else fails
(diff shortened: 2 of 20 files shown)


bryzettler (approved):

packages/blockchain-api-client/src/contracts/reward-contract.ts
```diff
@@ -104,7 +104,7 @@ export const rewardContract = oc
                 z.object({
                     entityPubKey: HeliumPublicKeySchema.describe("The public key of the hotspot entity"),
                     signerWalletAddress: WalletAddressSchema.describe("The wallet address of the caller"),
-                    expirationDays: z.number().int().positive().max(365).default(7).describe("Number of days until the invite expires"),
+                    expirationDays: z.number().int().positive().max(14).default(3).describe("Number of days until the invite expires"),
```
bryzettler, line 107: Should be fine, I cant remember why we had 365 originally

bryzettler (approved):

PR #1254, opened 2026-08-12
Build program errors only where they are raised

madninja:
An `AnchorError` owns its name and message as `String`s, so `error!()` allocates. Passed to
`ok_or`, it is built on every call, including the calls that succeed. A program's heap is a bump
allocator whose `dealloc` is a no-op, so an error that is never returned still consumes it for the
rest of the instruction.

Sixteen sites, of which one concentration matters: `mini_fanout::distribute_v0` builds eight per
share, inside the loop that assigns share payouts. The largest live fanout account is 4,810 bytes,
around fifty shares, so a distribution there builds roughly four hundred errors it discards. The
remaining sites are one per call.

`mini-fanout` distributions are not failing today — the deepest transaction sampled used 335,374
compute units and none of the recent failures are allocation failures — so this is headroom rather
than a live fault. The same construct in a five-instruction task elsewhere was enough to exhaust a
heap, which is why it is worth removing while the sites are known.

`clippy::or_fun_call` names this pattern and CI now fails on it, so it cannot come back quietly.
Reintroducing one of these fails the lint. `update_mini_fanout_v0` used
`unwrap_or(Pubkey::default())`, which allocates nothing, and becomes `unwrap_or_default()` so the
lint needs no exception.

### DIFF
--- a/programs/mini-fanout/src/instructions/distribute_v0.rs
+++ b/programs/mini-fanout/src/instructions/distribute_v0.rs
@@ -124,8 +124,7 @@ pub fn handler<'info>(
     .iter()
     .map(|s| s.total_dust as u128)
     .sum::<u128>()
-    .checked_div(DUST_PRECISION)
-    .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
+    / DUST_PRECISION;
   let mut remaining = (token_account.amount as u128).saturating_sub(total_dust);
 
   // 3. Assign fixed payouts in order, saturating if not enough left
@@ -165,28 +164,18 @@ pub fn handler<'info>(
       .ok_or_else(|| error!(ErrorCode::ArithmeticError))?
       .checked_div(total_shares as u128)
       .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
-    let payout = u64::try_from(
-      amount
-        .checked_div(DUST_PRECISION)
-        .ok_or_else(|| error!(ErrorCode::ArithmeticError))?,
-    )
-    .map_err(|_| error!(ErrorCode::ArithmeticError))?;
-    let dust = u64::try_from(
-      amount
-        .checked_rem(DUST_PRECISION)
-        .ok_or_else(|| error!(ErrorCode::ArithmeticError))?,
-    )
-    .map_err(|_| error!(ErrorCode::ArithmeticError))?
-    .checked_add(share.total_dust)
-    .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
+    let payout =
+      u64::try_from(amount / DUST_PRECISION).map_err(|_| error!(ErrorCode::ArithmeticError))?;
+    let dust = u64::try_from(amount % DUST_PRECISION)
+      .map_err(|_| error!(ErrorCode::ArithmeticError))?
+      .checked_add(share.total_dust)
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
     if dust >= DUST_PRECISION as u64 {
       payouts[i] = payout
         .checked_add(1)
         .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
-      new_dusts[i] = dust
-        .checked_sub(DUST_PRECISION as u64)
-        .ok_or_else(|| error!(ErrorCode::ArithmeticError))?
-        .into();
+      // The branch condition is what makes this subtraction safe.
+      new_dusts[i] = (dust - DUST_PRECISION as u64).into();
     } else {
       payouts[i] = payout;
       new_dusts[i] = dust as u128;
--- a/programs/mini-fanout/src/instructions/distribute_v0.rs
+++ b/programs/mini-fanout/src/instructions/distribute_v0.rs
@@ -125,7 +125,7 @@ pub fn handler<'info>(
     .map(|s| s.total_dust as u128)
     .sum::<u128>()
     .checked_div(DUST_PRECISION)
-    .ok_or(error!(ErrorCode::ArithmeticError))?;
+    .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
   let mut remaining = (token_account.amount as u128).saturating_sub(total_dust);
 
   // 3. Assign fixed payouts in order, saturating if not enough left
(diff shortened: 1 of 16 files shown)


programs/mini-fanout/src/instructions/distribute_v0.rs
```diff
@@ -160,32 +160,32 @@ pub fn handler<'info>(
...
     let amount = remaining
       .checked_mul(share_val)
-      .ok_or(error!(ErrorCode::ArithmeticError))?
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?
       .checked_mul(DUST_PRECISION)
-      .ok_or(error!(ErrorCode::ArithmeticError))?
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?
       .checked_div(total_shares as u128)
-      .ok_or(error!(ErrorCode::ArithmeticError))?;
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
```
bryzettler, line 166: Not introduced by this PR, but it's in the function this PR touches: `total_shares` (line 152) is the one arithmetic site here that is *unchecked* — `share_indices.iter().map(..).sum::<u32>()`. `Share::Share { amount }` is a full `u32` and initialize/update only bound the share *count* (`MAX_SHARES`), not magnitude, so two shares of 3 000 000 000 overflow the sum. With `overflow-checks = true` in the root Cargo.toml that's a panic on every `distribute_v0` — no `ErrorCode` — until the owner rewrites the shares.

Widening the sum to `u128` (as `total_dust` already is) fixes it, and `checked_div` still catches the `total_shares == 0` case:

```rust
  let total_shares: u128 = share_indices
    .iter()
    .map(|&i| mini_fanout.shares[i].share.as_u32() as u128)
    .sum();
```

which makes this line:

```suggestion
      .checked_div(total_shares)
      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
```

programs/mini-fanout/src/instructions/distribute_v0.rs
```diff
@@ -125,7 +125,7 @@ pub fn handler<'info>(
     .map(|s| s.total_dust as u128)
     .sum::<u128>()
     .checked_div(DUST_PRECISION)
-    .ok_or(error!(ErrorCode::ArithmeticError))?;
+    .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
```
bryzettler: Since this PR is about not building errors on paths that don't need them: `DUST_PRECISION` is a nonzero const, so this `checked_div` can never return `None` — the error arm is unreachable. Plain division deletes it outright:

```suggestion
    / DUST_PRECISION;
```

programs/mini-fanout/src/instructions/distribute_v0.rs
```diff
@@ -160,32 +160,32 @@ pub fn handler<'info>(
...
-      .ok_or(error!(ErrorCode::ArithmeticError))?
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?
       .checked_div(total_shares as u128)
-      .ok_or(error!(ErrorCode::ArithmeticError))?;
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
     let payout = u64::try_from(
       amount
         .checked_div(DUST_PRECISION)
-        .ok_or(error!(ErrorCode::ArithmeticError))?,
+        .ok_or_else(|| error!(ErrorCode::ArithmeticError))?,
```
bryzettler: Same as above — division by the nonzero const `DUST_PRECISION` can't fail, so the error arm is dead:

```suggestion
        / DUST_PRECISION,
```

programs/mini-fanout/src/instructions/distribute_v0.rs
```diff
@@ -160,32 +160,32 @@ pub fn handler<'info>(
...
     )
     .map_err(|_| error!(ErrorCode::ArithmeticError))?;
     let dust = u64::try_from(
       amount
         .checked_rem(DUST_PRECISION)
-        .ok_or(error!(ErrorCode::ArithmeticError))?,
+        .ok_or_else(|| error!(ErrorCode::ArithmeticError))?,
```
bryzettler: `checked_rem` by the nonzero const `DUST_PRECISION` can't fail either:

```suggestion
        % DUST_PRECISION,
```

bryzettler (commented):
A few observations from reviewing this branch (inline suggestions below). One can't be attached inline because the file isn't in the diff:

**`package.json` clippy script drift** — CI now enforces `-D clippy::or_fun_call`, but the root `clippy` npm script (the only local lint entry point; there's no `clippy.toml` or `[workspace.lints]`) wasn't updated. It's also already missing CI's `-A clippy::uninlined-format-args` and `-A clippy::manual_is_multiple_of`. A dev who writes a fresh `.ok_or(error!(..))` will pass `yarn clippy` locally and then fail the Test Rust Lint job. Suggested one-liner to keep them in agreement:

```json
"clippy": "cargo clippy --all-targets -- -D warnings -D clippy::or_fun_call -A clippy::result_large_err -A clippy::too_many_arguments -A clippy::uninlined-format-args -A clippy::manual_is_multiple_of -A ambiguous_glob_reexports -A unexpected-cfgs",
```

programs/mini-fanout/src/instructions/distribute_v0.rs
```diff
@@ -160,32 +160,32 @@ pub fn handler<'info>(
...
     if dust >= DUST_PRECISION as u64 {
       payouts[i] = payout
         .checked_add(1)
-        .ok_or(error!(ErrorCode::ArithmeticError))?;
+        .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
       new_dusts[i] = dust
         .checked_sub(DUST_PRECISION as u64)
-        .ok_or(error!(ErrorCode::ArithmeticError))?
+        .ok_or_else(|| error!(ErrorCode::ArithmeticError))?
         .into();
```
bryzettler: This `checked_sub` is guarded by `dust >= DUST_PRECISION as u64` on line 182, so it can't underflow — another dead error arm:

```suggestion
      new_dusts[i] = (dust - DUST_PRECISION as u64).into();
```

After these four, the remaining `checked_*` sites (`checked_mul` ×2, `checked_div(total_shares)`, `checked_add`) are exactly the ones that can actually fail, which makes the live failure points easier to see.

bryzettler (approved):

madninja:
Took four of five inline suggestions and the `package.json` one; holding one back. Verified each against the code rather than applying on faith.

**Taken** — `DUST_PRECISION` is `u128 = 1_000_000_000_000`, a nonzero constant, so lines 127/170/176 could never return `None`, and `if dust >= DUST_PRECISION` proves the subtraction at 187. These are better than what the PR originally did: the errors aren't deferred, they're gone, so that's four allocations removed unconditionally rather than made lazy. `90f089a`.

**Already done** — the `ok_or` → `ok_or_else` part of the line-167 comment is in the branch already (163/165/167 are `ok_or_else`); that suggestion reads against the base.

**One caution worth recording**, since it's adjacent to the three that were taken: line 165 is `checked_div(total_shares as u128)` and **`total_shares` can be zero** — `share_indices` holds only `Share::Share` entries and nothing stops every `amount` being 0. Applying "replace `checked_div` with `/`" across the block would turn a handled `ArithmeticError` into a division-by-zero panic. It keeps its check.

**Held back: the `total_shares: u32` overflow.** The finding is real — `sum()` into `u32` with `overflow-checks = true` panics above `u32::MAX`, and the owner sets the weights. But blast radius is that one fanout and it's recoverable by updating to smaller weights, and it's unrelated arithmetic in money-movement code with no test. I'd rather it went in on its own with a test that a large-weight fanout still distributes, than ride an approved lint sweep. Happy to file it.

**`package.json` drift** — taken, and the diagnosis needs one correction. `or_fun_call` is allow-by-default (proven: tuktuk CI was green for months with nine such sites until `-D` was added), so that half is exactly right: local passes, CI fails. But the two missing `-A` flags cut the other way — those lints are warn-by-default, so omitting the allow made the local script **stricter** than CI. I ran it: the script exits **101** on `develop` today, on `clippy::manual_is_multiple_of`. It was already broken, which is how the `or_fun_call` gap went unnoticed. Flags now match the workflow exactly, verified by comparing the parsed flag sets, and it exits 0. `b61edf3`.

Note this leaves two hand-synced copies of the lint set, which is how it drifted twice. The drift-proof fix is `[workspace.lints.clippy]` plus `[lints] workspace = true` in each member manifest — worth doing, but ~20 manifests, so not here.

programs/mini-fanout/src/instructions/distribute_v0.rs
madninja: Applied in 90f089a. `total_dust` divides by the constant directly, so the error is gone rather than deferred.

programs/mini-fanout/src/instructions/distribute_v0.rs
madninja: Applied in 90f089a: `u64::try_from(amount / DUST_PRECISION)`.

programs/mini-fanout/src/instructions/distribute_v0.rs
madninja: Applied in 90f089a: `u64::try_from(amount % DUST_PRECISION)`.

programs/mini-fanout/src/instructions/distribute_v0.rs
madninja: Applied in 90f089a, with a comment recording that the branch condition is what makes the subtraction safe.

programs/mini-fanout/src/instructions/distribute_v0.rs
madninja: Agreed it is real, and agreed with your own framing that it is not introduced here — so I have filed it as #1255 rather than fold it in. This PR is a lint sweep; that is unrelated arithmetic in a program that moves tokens, and it wants a test that a fanout whose weights exceed `u32::MAX` still distributes. Nothing exercises that range today.

Resolving this thread against the issue so the finding is tracked rather than buried in a merged PR.

One adjacent note worth recording, since three suggestions in this review replaced `checked_div` with `/`: line 165 divides by `total_shares`, which **can be zero** — `share_indices` holds only `Share::Share` entries and nothing stops every `amount` being 0. That one keeps its check; applying the same transformation there would turn a handled `ArithmeticError` into a division-by-zero panic.

PR #1263, opened 2026-08-17
Export dc-auto-top leg liveness and DCA input gauges

madninja:
## Why

A `dc-auto-top` leg is self-rescheduling: each run queues its own next task as its last step. A healthy leg therefore always points at a task whose trigger is in the future.

The ways a leg ends without a reschedule are a stale or wrong-owner price oracle, lamports below the rent-exempt plus crank-reward floor, and a reverting swap or run. tuktuk then retries the task until it goes stale, sweeps it, and nothing restarts the leg without `schedule_task_v0`.

Two things the trigger gauge does not cover, which is why the USDC balance gauge and a third `leg="dca"` series exist. A USDC shortfall logs `Skipping DCA` and the HNT leg still reschedules, so the gauge stays healthy while no HNT is being bought. And `auto_top_off.dca` is written but never read back, so a DCA whose task chain stops is invisible: `next_hnt_task` stays in the future and the next run funds another DCA from the same USDC account.

Nothing observed that. The carrier offload leg's HNT half sat broken for several days and the only alert that would eventually have noticed is a balance threshold that trips ~2 days later, once the HNT buffer has drained.

## What

`solana_auto_top_off_task_trigger{name,leg}` reports the trigger time of the task each leg currently points at, and `0` when none is scheduled. `leg` is `dc`, `hnt`, or `dca` for the DCA the HNT leg funds; the `dca` series is dropped entirely once that DCA is gone or has no orders left, so a finished DCA does not read as a stalled one. One expression covers both failure shapes:

```
time() - solana_auto_top_off_task_trigger{leg=~"hnt|dca"} > 900
```

A healthy leg has a future trigger so the difference is negative; a leg that is failing to execute has a past one; and `0` makes the same expression catch a swept task rather than needing a second clause. The threshold is crank latency rather than a schedule period: the difference is negative for a healthy leg whatever the cron interval, so waiting a whole interval would delay detection by that interval.

Also exported, since they are the two inputs the HNT leg's DCA reads and neither was visible:

- auto-top-off USDC balances (`solana_account_balance{type="usdc"}`). The DCA moves a whole run's worth up front, and a shortfall shrinks or skips it while the leg still reschedules, so this balance needs its own alert rather than showing up on the trigger gauge.
- USDC/USD pyth publish time (`solana_pyth_publish_time{name="usdc"}`). The leg refuses to start on a feed older than five minutes. Only the HNT feed was exported.

This finally uses the `USDC_MINT` constant that has been declared and unreferenced in `index.ts`.

## Notes for review

- **`helium_mobile_auto_topoff/hnt` reports 0 today.** `Bvge1…`'s HNT leg has no task: its stored input price oracle is owned by the legacy Pyth receiver, which the program no longer accepts, so the leg cannot run until the account is updated. Helium Mobile is also decommissioned as a separate payer (its traffic moved onto the carrier offload router `112c85vb…`), so whether that leg is revived is an operator decision. Until it is, **an alert on `leg="hnt"` should scope to `name="carrier_auto_topoff"`**; `leg="dc"` is meaningful for both accounts. Revisit the scope if the leg is brought back rather than treating the 0 as permanent.
- **Why this matters more now.** Every reverting failure in either leg leaves it unscheduled (a stale Pyth read, a circuit-breaker trip, an overflow, any account mismatch), and tuktuk sweeps the failing task after `stale_task_age`. With production USDC about to flow through the HNT refill leg, an unnoticed dead leg is idle capital on one side or a DC reserve that stops replenishing on the other. This gauge is the signal that fires within an hour of either.
- **Compatibility with the dc-auto-top changes merged this week.** The account is decoded through the IDL, so the two appended schedule fields do not move anything read here. `next_task == auto_top_off` is the "nothing scheduled" sentinel the program writes after a run with no reward and, as of #1313's follow-ups, after `update_auto_top_off_v0`, so a freshly updated top-off reads 0 until `schedule_task_v0` runs, which is the correct state. The description check already treats a task at a reused id that another program queued as not this leg's task.
- **An RPC failure deliberately does not zero the gauge.** The handler logs and leaves the previous reading, because reporting a transient read failure as a stalled leg would page on every blip; `watch` retries every 5 minutes. The tradeoff is that a persistent partial failure holds a stale healthy-looking value, which is the same property `solana_pyth_publish_time` already has.
- **Dependencies.** `@helium/tuktuk-sdk` is pinned to the workspace's `^0.1.1` and `@helium/tuktuk-dca-sdk` is a workspace link. The lockfile was regenerated from `develop`'s with the pinned `pnpm@9.15.4`, so `git diff origin/develop -- pnpm-lock.yaml` is three added importer entries under `packages/monitor-service` and nothing else.

## Verification

`pnpm turbo run build --filter=@helium/monitor-service` is green, run against a tree using the same pre-built idls plus `target/types` placeholder the Dockerfile itself sets up, which is what CI's `test-docker-builds` gate exercises.

`pnpm exec tsc --noEmit -p packages/monitor-service/tsconfig.build.json` is clean, and confirmed to be measuring the changed file by planting a type error in it and watching the check go red.

The monitors' decode path was also run against mainnet and the values it would emit read back, rather than relying on the build alone. At `now = 1789496162` (2026-09-15T18:16:02Z):

```
solana_auto_top_off_task_trigger{name="carrier_auto_topoff",leg="dc"}          1789574400   (time()-this = -78238s, healthy)
solana_auto_top_off_task_trigger{name="carrier_auto_topoff",leg="hnt"}         1789574700   (time()-this = -78538s, healthy)
solana_auto_top_off_task_trigger{name="carrier_auto_topoff",leg="dca"}         (no series)  (no DcaV0 at auto_top_off.dca)
solana_auto_top_off_task_trigger{name="helium_mobile_auto_topoff",leg="dc"}    1789574400   (healthy)
solana_auto_top_off_task_trigger{name="helium_mobile_auto_topoff",leg="hnt"}            0   (no task)
solana_auto_top_off_task_trigger{name="helium_mobile_auto_topoff",leg="dca"}   (no series)  (no DcaV0 at auto_top_off.dca)
solana_pyth_publish_time{name="usdc"}                                         1789496227
solana_account_balance{name="carrier_auto_topoff",type="usdc"}                         0
solana_account_balance{name="helium_mobile_auto_topoff",type="usdc"}                   0
```

Each `dc`/`hnt` task was matched by description, so the ownership check is exercised on live data. No `DcaV0` account exists on mainnet yet, so the `dca` leg's `remove` path is what the readings above exercise; its publish path is not covered by them.

The `timestamp` decode path is exercised by the non-zero readings, so the `now`-trigger fallback is the only unexercised branch.


bryzettler (commented):
Two comments describe a shortfall as ending the leg when the program skips and reschedules, the DCA the HNT leg funds has no series of its own, the gauge help states the wrong stall rule, and the lockfile carries two unrelated hunks.

packages/monitor-service/src/index.ts
```diff
@@ -235,6 +236,27 @@ async function run() {
     );
   }
 
+  // The DCA leg buys its HNT with the USDC held here, transferring a whole run's worth
+  // up front. A shortfall reverts the run rather than shrinking it, which stops the leg
+  // rescheduling itself, so this balance is watched alongside the HNT it buys.
```
bryzettler: This comment does not match the program in this branch. `top_off_hnt_v0.rs` logs `Skipping DCA` on a rent or `dca_mint` shortfall and still falls through to the reschedule at the end of the instruction, so a USDC shortfall does not stop the leg rescheduling itself. That is precisely why this balance needs its own alert rather than relying on the trigger gauge, so it is worth saying so here.

```suggestion
  // The DCA leg buys its HNT with the USDC held here, moving a whole run's worth up
  // front. A shortfall skips the DCA but the leg still reschedules itself, so the trigger
  // gauge stays healthy and this balance needs its own alert.
```

packages/monitor-service/src/monitors/autoTopOff.ts
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * tuktuk reuses task ids, so once a leg's task is swept another program's task takes the
+ * same address and the stored pointer still resolves. The trigger it carries is then
+ * somebody else's, and a stalled leg reads as a healthy one. dc-auto-top stamps each task
+ * it queues with `topoff <leg> <truncated auto top off key>`, so the description is what
+ * distinguishes its own task from a squatter at the same id.
+ */
+function ownsTask(description: string, leg: Leg, address: string): boolean {
+  // The program truncates the key to fit tuktuk's description limit: 32 - "topoff dc ".len
+  // for the DC leg and one char less for the longer "topoff hnt " prefix.
```
bryzettler: The derivation in this comment is wrong even though the literals are right: `"topoff dc "` is 10 characters, not 14, and `"topoff hnt "` is 11, not 15, so neither number is the prefix length and `hnt` is not "one char less" for being longer. The constants are copied verbatim from the program (`top_off_dc_v0.rs` uses `[..(32 - 14)]`, `top_off_hnt_v0.rs` uses `[..(32 - 15)]`), so pointing at those call sites is both accurate and what a reader needs to re-verify the match. No code change.

```suggestion
  // Mirrors the slice the program takes in top_off_dc_v0 ([..32 - 14]) and
  // top_off_hnt_v0 ([..32 - 15]).
```

packages/monitor-service/src/monitors/autoTopOff.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Tracks whether each leg of a `dc-auto-top` account is still being cranked.
+ *
+ * Both legs are self-rescheduling: a run reschedules itself as its last step, so a
+ * healthy leg always points at a task whose trigger is in the future. Every way a leg
+ * can break — a stale or wrong-owner price oracle, an empty DCA input account, a
+ * `dca` PDA left over from an undrained run, lamports below the crank reward, a
+ * reverting swap — ends the same way, with no reschedule. tuktuk then retries the
+ * task until it goes stale and is swept, after which nothing restarts the leg without
+ * `schedule_task_v0`. So the trigger time of the task each leg points at is the single
+ * signal that covers all of them.
```
bryzettler: Two of the listed failure modes do not end without a reschedule, so the docblock oversells what this one gauge covers. An empty DCA input account is logged as a skip and the HNT leg still reschedules, and the leftover-`dca`-PDA case is gone now that `top_off_hnt_v0.rs` derives a fresh PDA from `dca_index`. Listing only the cases that really stop the reschedule, and naming the two that do not, is what justifies the USDC gauge and the `dca` leg alongside this one.

```suggestion
 * Both legs are self-rescheduling: a run reschedules itself as its last step, so a
 * healthy leg always points at a task whose trigger is in the future. The ways a leg
 * ends without a reschedule are a stale or wrong-owner price oracle, lamports below the
 * crank reward, and a reverting swap or run. tuktuk then retries the task until it goes
 * stale and is swept, after which nothing restarts the leg without `schedule_task_v0`.
 * So the trigger time of the task each leg points at is the single signal that covers
 * all of them.
 *
 * An empty DCA input account does not stop the leg rescheduling — the run logs a skip
 * and reschedules — and neither does a DCA chain that dies after it is funded, since
 * nothing reads `auto_top_off.dca` back. That is why the USDC balance gauge and the
 * `dca` leg below exist: a DCA whose task chain stops has no series of its own, while
 * `next_hnt_task` stays in the future.
```

packages/monitor-service/src/monitors/autoTopOff.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  watch(autoTopOff, async (raw) => {
+    if (!raw) return;
+    try {
+      const acc = dcAutoTopProgram.coder.accounts.decode(
+        "autoTopOffV0",
+        raw.data
+      );
+      await publish("dc", acc.nextTask);
+      await publish("hnt", acc.nextHntTask);
```
bryzettler, line 93: These two legs leave a gap: the DCA the HNT leg funds has no series of its own. `top_off_hnt_v0.rs` decides whether to start a DCA from the HNT balance alone, and `auto_top_off.dca` is written there but never read back, so a DCA whose task chain stops is not noticed by anything here — `next_hnt_task` stays in the future and the next run funds another DCA from the same USDC account. A third `leg="dca"` series closes that, and `remove` rather than `set(0)` keeps a finished DCA from reading as a stalled one.

The change spans package.json and several regions of this file, so it cannot be a single suggestion:

```ts
// packages/monitor-service/package.json — dependencies, alphabetical among the @helium entries:
"@helium/tuktuk-dca-sdk": "workspace:^",

// autoTopOff.ts — imports:
import { init as initTuktukDca } from "@helium/tuktuk-dca-sdk";

// autoTopOff.ts — the Leg union:
type Leg = "dc" | "hnt" | "dca";

// autoTopOff.ts — ownsTask, so the dca leg checks tuktuk-dca's own stamp
// (`dca {key[..(32 - 4)]}`, from initialize_dca_v0.rs and check_repay_v0.rs):
if (leg === "dca") {
  return description === `dca ${address.slice(0, 32 - 4)}`;
}

// autoTopOff.ts — monitorAutoTopOff, next to the other two init calls:
const dcaProgram = await initTuktukDca(provider);

// autoTopOff.ts — publish takes the key that is stamped into the description,
// which is the auto top off for dc/hnt and the DCA itself for dca (tuktuk-dca never
// parks its own key as a sentinel, so the equals check is a no-op for that leg):
async function publish(leg: Leg, task: PublicKey, owner: PublicKey = autoTopOff) {
  const ownerAddress = owner.toBase58();
  const acc = task.equals(owner)
    ? null
    : await tuktukProgram.account.taskV0.fetchNullable(task);
  if (!acc || !ownsTask(acc.description, leg, ownerAddress)) {
  ...

// autoTopOff.ts — in the watch callback, right after the dc and hnt publishes:
const dca = await dcaProgram.account.dcaV0.fetchNullable(acc.dca);
if (!dca || dca.numOrders === 0) {
  autoTopOffTaskTrigger.remove({ name: label, leg: "dca", address });
} else {
  await publish("dca", dca.nextTask, acc.dca);
}
```

`numOrders` and `nextTask` are `DcaV0.num_orders` / `DcaV0.next_task` in `programs/tuktuk-dca/src/state.rs`; the generated IDL types were not available to check the camel-cased names against, so please confirm them against a built `@helium/idls` before merging.

packages/monitor-service/src/metrics.ts
```diff
@@ -98,3 +98,14 @@ export const pythPublishTime = new client.Gauge({
   labelNames: ["name", "address"],
 });
 register.registerMetric(pythPublishTime);
+
+export const autoTopOffTaskTrigger = new client.Gauge({
+  name: "solana_auto_top_off_task_trigger",
+  help:
+    "Unix trigger time of the tuktuk task this auto top off leg currently points at, " +
+    "or 0 when no task is scheduled. A leg is stalled when time() - this exceeds one " +
+    "cron interval: a healthy leg always points at a future trigger, because each run " +
```
bryzettler: "exceeds one cron interval" makes this alert fire a whole schedule period late. Both legs always schedule the next cron slot (the HNT leg adds 300s to `get_next_time`, and `schedule_task_v0` advances to the next slot), so a healthy leg's trigger is always in the future and `time() - this` is negative. Any positive value beyond crank latency is already a stall; with a daily schedule the current threshold delays detection by a day. Phrasing it like the `pythPublishTime` help just above also lets it name the new `dca` leg.

```suggestion
    "Unix trigger time of the tuktuk task this auto top off leg (dc, hnt, or the dca it " +
    "funded) points at, or 0 when none is scheduled; a healthy leg always points at a " +
    "future trigger, so the leg is stalled once time() - this is positive for longer " +
    "than crank latency",
```

pnpm-lock.yaml
```diff
@@ -27419,7 +27424,7 @@ snapshots:
 
   metro-runtime@0.76.9:
     dependencies:
-      '@babel/runtime': 7.29.7
+      '@babel/runtime': 7.26.0
```
bryzettler: This hunk is unrelated to the manifest change in this PR: it downgrades the `metro-runtime@0.76.9` snapshot's `@babel/runtime` from 7.29.7 to 7.26.0. The same hunk set also drops the `deprecated:` line from the `eslint@9.39.4` entry around line 10371, which is a deleted line so it cannot be suggested here. Both look like drift from resolving the lockfile with a different pnpm than the `pnpm@9.15.4` the root package.json pins — regenerating with that version leaves only the `packages/monitor-service` importer entries changed.

```suggestion
      '@babel/runtime': 7.29.7
```

packages/monitor-service/src/index.ts
madninja: Taken, with the wording made true both before and after #1318: a shortfall now reads as shrinking or skipping the DCA while the leg still reschedules, so the trigger gauge stays healthy and this balance needs its own alert.

packages/monitor-service/src/monitors/autoTopOff.ts
madninja: Taken verbatim. Confirmed the literals on this branch: top_off_dc_v0.rs:274 and schedule_task_v0.rs:379 use [..(32 - 14)], top_off_hnt_v0.rs:461 and schedule_task_v0.rs:407 use [..(32 - 15)].

packages/monitor-service/src/monitors/autoTopOff.ts
madninja: Taken verbatim. Verified in top_off_hnt_v0.rs on this branch: both shortfalls msg! "Skipping DCA" (lines 341 and 360) and fall through to the reschedule at line 455, and the DCA PDA is derived from auto_top_off.dca_index (schedule_task_v0.rs:141, advanced at top_off_hnt_v0.rs:431).

packages/monitor-service/src/monitors/autoTopOff.ts
madninja: Added as sketched. Field names confirmed against a fresh anchor build rather than the checked-in types: target/types/tuktuk_dca.ts has dcaV0 (:1378) with nextTask (:1435) and numOrders (:1459), and target/types/dc_auto_top.ts has autoTopOffV0.dca (:2402). The dca stamp literal is [..(32 - 4)] in initialize_dca_v0.rs:190 and check_repay_v0.rs:184. No other site enumerates leg values. Live read of both mainnet top-offs: neither auto_top_off.dca points at an existing DcaV0, so both take the remove path.

packages/monitor-service/src/metrics.ts
madninja: Taken verbatim, and the PR body alert expression is now time() - solana_auto_top_off_task_trigger{leg=~"hnt|dca"} > 900 with a note that the threshold is crank latency rather than a schedule period.

pnpm-lock.yaml
madninja: Fixed by starting from develop pnpm-lock.yaml and re-resolving with the pinned pnpm 9.15.4. git diff origin/develop -- pnpm-lock.yaml is now three added importer entries under packages/monitor-service (dc-auto-top-sdk, tuktuk-dca-sdk, tuktuk-sdk) and nothing else; neither the @babel/runtime nor the eslint hunk reappeared, so nothing had to be restored by hand.

bryzettler (approved):

PR #1270, opened 2026-08-24
helium-sub-daos: HIP 150 Decisions 2 and 3 (on-chain)

madninja:
On-chain implementation of [HIP 150](https://github.com/helium/HIP/blob/main/0150-mobile-deployer-prioritization.md), isolated in one program for security review. Only the deployer earnings backstop changes: the target minimum rises to 80% of the carrier-paid USD, and the top-up that delivers it is minted straight into the Mobile data rewards escrow.

## What each decision needs on chain

**Decision 1, per-Hotspot multipliers: nothing.** A multiplier applies to the data credits derived from a Hotspot's rewardable bytes, and the payer burns those credits. `track_dc_burn_v0` accumulates whatever DC is burned with no per-Hotspot notion, so `dc_burned` already aggregates every multiplier in force, and both bounds of the band are computed from it. Tickets, the 1 to 5 range and the round-down are oracle-side.

**Decision 2, direct minting and an 80% target: the substance of this PR.** `issue_rewards_v0` divides only the part of the epoch's mint that is not the top-up, and adds the top-up to the escrow mint on the Mobile pass. IoT, the HST cut and the 6% delegator slice therefore take no share of HNT minted to support Mobile deployers, and the top-up is sized at the shortfall itself rather than grossed up by `1 / (mobile_share x bucket)` so that the surviving fraction equalled it. That is what makes 80% affordable instead of an increase in supply.

Unchanged: the ceiling at three times the carrier-paid USD, and the rule that the top-up may never exceed recent HNT destruction (`max(0, smoothed_hnt_burned - net_emissions_cap)`, shared with the HIP 20 re-emit).

**Decision 3, the Mobile data bucket: nothing, and one constant deleted.** `MOBILE_DATA_BUCKET_PERCENT` existed because the rewards escrow held the data bucket and the Service Provider bucket together, so the escrow mint was not the deployer pool and the chain needed a fraction to model an allocation only the oracles could see. With the Service Provider allocation at zero the escrow *is* the deployer pool, so both readers use the amount actually minted:

- `staker_overflow` caps the deployer pool it is handed, so the pool measured against the ceiling and the pool minted are the same number. A `delegator_rewards_percent` other than today's 6% cannot make them disagree.
- The baseline takes the delegation slice from `dao.delegator_rewards_percent`, at the same scale `issue_rewards_v0` divides by.

How a sub-DAO's rewards are allocated within that pool stays the oracles' concern. The invariant this depends on is recorded where it is read: the escrow is the deployer pool only while the Service Provider allocation is zero.

## No new state

`total_rewards` keeps its meaning, the epoch's whole DAO-level mint, so `current_hnt_supply` and every off-chain reader of it are untouched. What changed is that only part of it is divided.

`split_total_rewards()` recovers `(split_base, top_up)` from it, because every term that is not the top-up is already available to `issue_rewards_v0`: the emission schedule and the net-emissions cap on `DaoV0`, and the epoch's `smoothed_hnt_burned` on the `DaoEpochInfoV0` that `calculate_utility_score_v0` wrote. So there is no new account, no new field and no change to either instruction's account list. Nothing migrates, and end-of-epoch tasks queued before the upgrade still execute rather than failing closed on a missing account.

## The band cannot invert, and an unusable price is dormant

The floor converts its USD target at `ema - 2 * conf` while the cap converts its own at `ema`. At an 80% target the floor sits at 4/15 of the cap, so a Pyth confidence width above ~37% of the EMA price would put the target above the ceiling, and the top-up would be minted only for the cap to redirect it to veHNT delegators: burned HNT re-emitted to stakers rather than to deployers. The target is clamped to the ceiling where it is computed, so that top-up is never minted. What the cap still trims is the residual between this baseline and the pool `issue_rewards_v0` mints from the freshly-smoothed percent share.

`read_hnt_price` only requires `ema - 2 * conf` to be positive, so a price can be positive and still too small to convert the carrier-paid USD into a representable amount of HNT. `scale_dc_to_hnt` reports that rather than saturating, and both halves of the band go dormant for the epoch.

## Rollout

**Ship with the oracle release.** `hnt_rewards_issued` now includes the top-up, and the Mobile verifier reads `hnt_rewards_issued + delegation_rewards_issued` as `epoch_emissions` for its Service Provider share. That share has to reach zero in the same rollout or it would take 24% of an inflated base.

**Settle every pass of an epoch under one program version.** Each `split_total_rewards` call conserves, so any single pass mints what the epoch recorded. But every pass derives the split independently, and one epoch's mint is `total_rewards + other_passes_share x (their_base - the_Mobile_base)`, so a divergence between passes mints HNT the epoch's `current_hnt_supply` does not account for. The crank settles every pass in one transaction, where the inputs cannot move. A manual settlement through `end-epoch.ts` sends each pass separately, and no on-chain guard can detect this upgrade landing mid-settlement because the earlier passes ran under code that did not split at all. Verify no epoch is part-settled before deploying.

The generated IDL is byte-identical to the base branch, so no `@helium/idls` changeset is needed.

## Deliberately not changed

- **No end dates for either concession.** The 80% target is a plain constant; returning it is a governance-voted program upgrade, not a hardcoded expiry.
- **The one-epoch smoothing lag** between the share that sizes the top-up and the share used at issuance, which is inherent and within the HIP's best-effort-per-epoch aim.

## Review follow-ups

**The baseline takes the HST cut.** `issue_rewards_v0` applies `(100 - hst_percent) / 100` to the DAO-level mint before the sub-DAO split, so a baseline computed without it overstates what deployers receive and under-sizes the shortfall by that same percentage. `hst_percent` is now a `BackstopInput` field read from `DaoV0::hst_emission_schedule`, for the reason `delegator_rewards_percent` is read rather than hardcoded: `update_dao_v0` takes the whole schedule, so today's 0% is a value, not an invariant. A schedule with no entry at the epoch timestamp reads as 0 rather than halting, which is the under-deliver direction.

**The top-up is minted once per epoch, including under `TESTING`.** `is_mobile` is `TESTING || sub_dao.key() == MOBILE_SUB_DAO`, so a `TESTING` build with N sub-DAOs minted N top-ups against the one `total_rewards` recorded. The mint is now gated on the epoch's first pass under `TESTING` (`num_rewards_issued` is incremented after the mints). Production behaviour is unchanged: the term const-folds away and `is_mobile` already selects one pass.

**`PERCENT_SCALE` has one definition.** Moved to `state.rs` beside the field it scales, and `initialize_dao_v0` and `update_dao_v0` now validate against it instead of re-spelling `100 * 10_0000000`. Its doc said "2 decimals"; the field is 8, as `DaoV0::delegator_rewards_percent` already says.

Also: two assertions that could not fail (a split whose halves sum to the total by construction, and a redirect asserted against its own function body) replaced or dropped, and a two-condition `let ... else` split into the guards it fused.

## Known coverage gap

The `TESTING` first-pass gate is not itself covered: no suite issues rewards for two sub-DAOs in one epoch, so reverting it would go unnoticed. The HST read is covered both ways -- a unit test on the baseline, and a localnet epoch settled at 30% HST that asserts delivery still lands on the target, which the pre-fix baseline misses.


riobah:
> _Decision 1 needs nothing on chain_

Would not we be having this "Per-Hotspot reward multiplier" in the Helium Entity Manager, somewhere like `MobileHotspotInfoV0`?

I'm trying to understand how to access this information, to be able to make it available in the apps

madninja:
> > _Decision 1 needs nothing on chain_
> 
> Would not we be having this "Per-Hotspot reward multiplier" in the Helium Entity Manager, somewhere like `MobileHotspotInfoV0`?
> 
> I'm trying to understand how to access this information, to be able to make it available in the apps

it'll be in multiplier tickets going into the oracles, and the multiplier used will written as part of the verified packet burn reports

programs/helium-sub-daos/src/backstop.rs
```diff
@@ -181,26 +232,58 @@ pub fn compute_backstop(input: &BackstopInput) -> BackstopOutput {
...
+/// for, in either direction.
+///
+/// The end-of-epoch crank compiles every pass into a single transaction, where the inputs
+/// cannot move and the split is identical. The two ways they can diverge are a governance
+/// `update_dao_v0` to the emission schedule or the net-emissions cap landing between passes
+/// of a manual settlement, which sends each pass as its own transaction; and an upgrade of
+/// this program landing mid-settlement, which no on-chain guard can detect because the
+/// earlier passes ran under code that did not split at all. Settle every pass of an epoch
+/// under one program version.
+pub fn split_total_rewards(
```
bryzettler, line 261: `split_total_rewards` reconstructs the top-up inside `issue_rewards_v0` from `emission_schedule.get_emissions_at(ts)` and `min(smoothed_hnt_burned, net_emissions_cap)` — governance-mutable accounts — rather than reading the single number `calculate_utility_score_v0` already computed.

The doc comment above states the consequence itself: every pass derives the split independently, and a divergence between passes mints HNT the epoch's `current_hnt_supply` does not account for, with no on-chain guard able to detect the program-upgrade case. Storing the value removes that class instead of documenting it. There is precedent in the same account: `deployer_cap_hnt` is already carried on `DaoEpochInfoV0` so `issue_rewards_v0` need not recompute the ceiling. `DaoEpochInfoV0::size()` is `60 + 8 + size_of::<DaoEpochInfoV0>()`, and the added `u64` fits in that slack — the new layout needs 308 bytes against the 364 an existing account was allocated — so no resize or migration.

```rust
// state.rs, right after deployer_cap_hnt
pub backstop_top_up: u64,

// calculate_utility_score_v0.rs, beside the deployer_cap_hnt write
ctx.accounts.dao_epoch_info.backstop_top_up = backstop.top_up;

// issue_rewards_v0.rs, in place of the split_total_rewards call
let backstop_top_up = ctx.accounts.dao_epoch_info.backstop_top_up;
let total_emissions = ctx
  .accounts
  .dao_epoch_info
  .total_rewards
  .saturating_sub(backstop_top_up);
```

`split_total_rewards`, its doc comment, and its two tests then go away.

programs/helium-sub-daos/src/instructions/issue_rewards_v0.rs
```diff
@@ -208,24 +224,29 @@ pub fn handler(ctx: Context<IssueRewardsV0>, args: IssueRewardsArgsV0) -> Result
...
+  //   escrow, so all of it reaches deployers instead of a fraction surviving the splits.
+  // - The earnings cap holds deployers at no more than three times the carrier-paid USD:
+  //   any HNT in the data bucket and top-up above the ceiling is redirected from the
+  //   escrow to the shared delegator pool. That redirect is a bucket-to-bucket move, so
+  //   the HNT is paid to veHNT delegators instead of deployers and staker claims already
+  //   key off the DAO-level delegation_rewards_issued. A zero ceiling (no carrier burn
+  //   this epoch, or no price oracle) disables it.
   let is_mobile = TESTING || ctx.accounts.sub_dao.key() == crate::backstop::MOBILE_SUB_DAO;
-  let escrow_before_overflow = rewards_amount
+  let top_up = if is_mobile { backstop_top_up } else { 0 };
```
bryzettler: `is_mobile` is `TESTING || sub_dao.key() == MOBILE_SUB_DAO`, so in a TESTING build every sub-DAO pass takes this branch and mints the full top-up: an epoch with N sub-DAOs mints N of them, while `calculate_utility_score_v0` recorded exactly one in `total_rewards` and in `current_hnt_supply`. The floor-delivery test issues for a single sub-DAO, so it cannot observe the extra mints.

`num_rewards_issued` is incremented after the mints below, so it identifies the epoch's first pass and keeps the top-up the once-per-epoch DAO-level amount it is sized as. Production behaviour is unchanged: `TESTING` is false there, and `is_mobile` already selects one pass.

```suggestion
  // is_mobile is every sub-DAO under TESTING, so gate the mint on the epoch's first pass
  // there: the top-up is a DAO-level amount and must be minted once, not once per sub-DAO.
  let is_top_up_pass =
    is_mobile && (!TESTING || ctx.accounts.dao_epoch_info.num_rewards_issued == 0);
  let top_up = if is_top_up_pass { backstop_top_up } else { 0 };
```

programs/helium-sub-daos/src/backstop.rs
```diff
@@ -263,17 +353,64 @@ mod tests {
...
+    ] {
+      let (split_base, top_up) =
+        split_total_rewards(out.total_rewards, emission, 91_000_00000000, net_cap);
+      assert_eq!(
+        split_base + top_up,
+        out.total_rewards,
+        "emission {emission}, net_cap {net_cap}: the split lost or invented HNT"
+      );
+    }
+  }
```
bryzettler: `split_base + top_up == total_rewards` holds for every input by construction: `split_total_rewards` returns `(split_base, total_rewards - split_base)`, so no `emission` or `net_cap` in the table can make this assertion fail. The comment says a moved base is what is being exercised, but the assertion is blind to it.

Pinning the derived `top_up` per row makes the moved base observable — the `u64::MAX` and `net_cap = u64::MAX` rows clamp to `total_rewards` and yield 0, the others shift by the amount the base moved.

```suggestion
  #[test]
  fn split_total_rewards_conserves_under_a_moved_base() {
    // A manual settlement that changed the emission schedule or the net-emissions cap
    // between the two instructions moves the derived base. The halves sum to total_rewards
    // whatever that base is, so pin the derived top-up per row instead: that is the
    // quantity a moved base actually changes.
    let out = compute_backstop(&base_input(0.10, 91_000_00000000));
    let total = out.total_rewards;
    for (emission, net_cap, expected_top_up) in [
      (EMISSION, NET_CAP, out.top_up),
      (EMISSION * 2, NET_CAP, total - (EMISSION * 2 + NET_CAP)),
      (EMISSION / 2, NET_CAP, total - (EMISSION / 2 + NET_CAP)),
      (u64::MAX, NET_CAP, 0),
      (EMISSION, 0, total - EMISSION),
      (EMISSION, u64::MAX, 0),
    ] {
      let (split_base, top_up) = split_total_rewards(total, emission, 91_000_00000000, net_cap);
      assert_eq!(
        top_up, expected_top_up,
        "emission {emission}, net_cap {net_cap}: a moved base moves the derived top-up"
      );
      assert_eq!(split_base + top_up, total);
    }
  }
```

If the stored-`backstop_top_up` change is taken, this test and `split_total_rewards_inverts_the_computation` disappear along with the function.

programs/helium-sub-daos/src/backstop.rs
```diff
@@ -342,69 +479,155 @@ mod tests {
...
+    for pool in [13_870_00000000_u64, 20_000_00000000, 5_000_00000000] {
+      assert_eq!(
+        staker_overflow(pool, cap),
+        pool.saturating_sub(cap),
+        "pool {pool}: the redirect is a function of the pool and the ceiling alone"
+      );
+    }
+  }
+
+  #[test]
```
bryzettler: `staker_overflow(pool, cap)` is `pool.saturating_sub(cap)` for any nonzero `cap`, so the assertion restates the function body and holds however that body is rewritten. The three tests above already pin the same behaviour against independently written expected values (2,950 HNT redirected, 0 inside the band, never more than the pool).

The range below drops this test and the `#[test]` line under it, leaving the `#[test]` above to attach to `mobile_share_zero_is_dormant`.

```suggestion
```

programs/helium-sub-daos/src/backstop.rs
```diff
@@ -59,6 +82,13 @@ pub struct BackstopInput {
   pub mobile_dc_burned: u64,
   /// Mobile sub-DAO 30-epoch EMA percent share (`previous_percentage`), scaled by `u32::MAX`.
   pub mobile_share: u32,
+  /// `DaoV0::delegator_rewards_percent`, scaled by `PERCENT_SCALE`. The rest of the
+  /// sub-DAO's emission is minted to the rewards escrow, and with the Service Provider
+  /// allocation at zero the escrow is the Mobile data deployer pool, so `1 - this` is the
+  /// fraction of the sub-DAO slice that reaches deployers. Read from chain rather than
+  /// hardcoded as a bucket percentage, so the baseline cannot disagree with the amount
+  /// `issue_rewards_v0` actually mints.
+  pub delegator_rewards_percent: u64,
```
bryzettler, line 89: `deployer_baseline` below is computed from `base_total` un-reduced, while `issue_rewards_v0` applies `(100 - hst_percent) / 100` to `total_emissions` before the sub-DAO split. The two agree only while `hst_emission_schedule` yields 0%, which the comment on `deployer_baseline` asserts as a fact rather than reading.

`update_dao_v0` accepts `hst_emission_schedule` as a whole `Option<Vec<PercentItem>>`, so a future nonzero entry silently inflates the baseline and under-sizes the floor by that same percentage. `delegator_rewards_percent` is already carried in this struct rather than hardcoded for exactly this reason; the HST cut is the other slice taken before deployers see the money.

```rust
// BackstopInput, beside delegator_rewards_percent
pub hst_percent: u64,

// compute_backstop, applied the way issue_rewards_v0 applies it: before the sub-DAO split
let after_hst =
  (base_total as u128).saturating_mul(100_u128.saturating_sub(input.hst_percent as u128)) / 100;
let deployer_baseline = after_hst.saturating_mul(share) / (u32::MAX as u128) * deployer_percent
  / (PERCENT_SCALE as u128);

// calculate_utility_score_v0, mirroring how issue_rewards_v0 reads it
hst_percent: dao
  .hst_emission_schedule
  .get_percent_at(end_of_epoch_ts)
  .unwrap()
  .into(),
```

`base_input` in the unit tests sets it to 0, so the existing worked examples are unchanged.

programs/helium-sub-daos/src/backstop.rs
```diff
@@ -43,8 +58,16 @@ pub const MOBILE_SUB_DAO: Pubkey = pubkey!("Gm9xDCJawDEKDrrQW6haw94gABaYzQwCq4ZQ
...
+/// The target minimum, as a percentage of the carrier-paid USD the Mobile sub-DAO's
+/// `dc_burned` measures (HIP 150 Decision 2). Hardcoded.
+pub const DEPLOYER_TARGET_PERCENT: u128 = 80;
+
+/// The earnings ceiling, as a multiple of the same carrier-paid USD (HIP 149 Decision 1).
+pub const DEPLOYER_CAP_MULTIPLE: u128 = 3;
+
+/// Scale of `DaoV0::delegator_rewards_percent`: 100% with 2 decimals of accuracy.
+/// `issue_rewards_v0` divides the delegator slice by this same figure.
+pub const PERCENT_SCALE: u64 = 100 * 10_0000000;
```
bryzettler: The constant replaces the literal in `issue_rewards_v0`, but two other sites still spell it out: `initialize_dao_v0.rs:92` and `update_dao_v0.rs:68` both validate `delegator_rewards_percent` with `require_gte!(100_u64.checked_mul(10_0000000).unwrap(), ...)`. Three definitions of one scale, two of them linked.

It also describes a `DaoV0` field rather than anything about the backstop, so `state.rs` next to `delegator_rewards_percent` is the natural home — both instruction files and `issue_rewards_v0` already glob-import `state::*`, so the reference needs no path.

```rust
// state.rs, above DaoV0
/// Scale of `DaoV0::delegator_rewards_percent`: 100% with 8 decimals of accuracy. Every
/// site that validates or divides by that field reads it from here.
pub const PERCENT_SCALE: u64 = 100 * 10_0000000;

// initialize_dao_v0.rs
require_gte!(PERCENT_SCALE, args.delegator_rewards_percent);

// update_dao_v0.rs
require_gte!(PERCENT_SCALE, delegator_rewards_percent);
```

Note the doc line here says "2 decimals of accuracy"; the field is 8 (`100 * 10_0000000` = 1e10).

programs/helium-sub-daos/src/backstop.rs
```diff
@@ -91,78 +124,96 @@ pub struct BackstopOutput {
...
+    top_up: 0,
+    total_rewards: base_total,
+    deployer_cap_hnt: deployer_cap_hnt.unwrap_or(0),
+  };
+
+  // mobile_share == 0 is a genesis or fully un-delegated Mobile sub-DAO, which takes no
+  // share of the emission and so has no deployer pool for the band to act on.
+  let (Some(deployer_cap_hnt), true) = (deployer_cap_hnt, input.mobile_share != 0) else {
+    return dormant;
+  };
```
bryzettler, line 184: `let (Some(x), true) = (a, b) else` fuses two unrelated bail-outs — a ceiling that will not fit in `u64`, and a Mobile share of zero — into one tuple destructure whose second arm matches a literal `true`. The comment above it only explains the share half, so the price half arrives unannounced inside a pattern.

Two sequential guards say the same thing and let each rationale sit against the condition it explains.

```suggestion
  // mobile_share == 0 is a genesis or fully un-delegated Mobile sub-DAO, which takes no
  // share of the emission and so has no deployer pool for the band to act on.
  if input.mobile_share == 0 {
    return dormant;
  }

  // An unrepresentable ceiling leaves the epoch with no band at all, per the note above.
  let Some(deployer_cap_hnt) = deployer_cap_hnt else {
    return dormant;
  };
```

bryzettler (commented):
Quality review: 7 comments — the main one proposes storing the backstop top-up on DaoEpochInfoV0 instead of re-deriving it at issue time; the rest cover a TESTING-build over-mint, two tautological tests, the HST cut missing from the baseline, PERCENT_SCALE dedup, and a guard-clause cleanup.

bryzettler (approved):

programs/helium-sub-daos/src/backstop.rs
madninja: Not taking this one, but the rejection is narrower than I first thought, so it is worth writing down what it rests on.

**Where you are right.** I initially reasoned that the derivation fails safe when an upgrade lands mid-settlement, because a pre-upgrade `total_rewards` would be `emission + min(smoothed, net_cap)` exactly and `split_total_rewards` would return `top_up = 0`. That is wrong: the base branch already adds a (grossed-up) top-up into `total_rewards`, so a pre-upgrade epoch carries a non-zero one. An IoT pass settled under the old code splits that top-up across the sub-DAOs, and a Mobile pass settled after the upgrade recovers it and mints it whole -- the epoch then mints `iot_share x top_up` more than `current_hnt_supply` accounts for. Storing the value does close that, because a pre-upgrade account reads 0 out of its slack and both passes behave identically.

**Why the derivation stays anyway.** No new field on `DaoEpochInfoV0`. A new field changes the generated IDL, so the "byte-identical to the base branch" property goes away and *Check @helium/idls Changeset* starts demanding a bump for a change that alters no interface.

What the hole actually needs is a deploy precondition, which the PR body already carries: verify no epoch is part-settled before deploying, and settle every pass of an epoch under one program version. The crank compiles every pass into a single transaction, so this is only reachable through a manual `end-epoch.ts` settlement that straddles the upgrade -- and it is an operational check either way, since the surrounding rollout (the oracle release, the Service Provider share reaching zero) has to be sequenced by hand regardless.

The other six are applied, plus the HST cut you flagged. Detail in the PR body.

PR #1271, opened 2026-08-25
Seed the DCA PDA from a monotonic index

madninja:
## Why

`top_off_hnt_v0` seeded its DCA PDA with a hardcoded `index: 0`, so the leg had exactly one slot for the life of the account. A DCA that fails to drain and close occupies that slot permanently, every later run fails `init` on it, and a failed run never reschedules — so the HNT leg stops.

Recovery is not currently available. `close_dca_v0` is the designed cleanup, but its `authority` is the `AutoTopOffV0` PDA and `dc-auto-top` exposes no CPI wrapper to sign for it, and it also takes `next_task` as a strict `Account<TaskV0>`, which fails to deserialize once tuktuk has swept the task — the state it exists to clear. Both need program changes and are not in this PR.

The likeliest trigger is mundane. Every DCA order task is a `RemoteV0` task pointing at a hosted URL, and `stale_task_age` on the queue is 172,800s, so a two-day outage of that service leaves the DCA orphaned. The sweep is permissionless and pays the cranker, so it happens promptly once the task ages out.

## What changed

`dca_index` advances once per DCA created, before the next task is compiled, so the next run always targets a fresh slot. An orphan then costs one day's USDC to sweep later rather than the use of the feature. It is carved out of `AutoTopOffV0.reserved`, leaving every other field at its existing byte offset, so the two live accounts need no migration and read it as 0.

Two lamport paths in the same handler could also stop the leg, and are hardened here:

- The crank-reward shortfall branch marks the leg unscheduled so `schedule_task_v0` can restart it, but it called `load_mut` while an earlier immutable borrow was still live, returned `AccountBorrowFailed`, and wrote nothing. That marker is what distinguishes "no task scheduled" from "scheduled but not running".
- The DCA rent transfer had no rent-exemption guard. A DCA refunds its rent when it closes, so in steady state nothing is drawn; an abandoned one never refunds and each replacement draws fresh rent. Debiting past rent exemption fails the run, so it now skips the DCA and leaves the leg rescheduling.

Both comparisons now saturate — subtracting the rent-exempt minimum from a balance already below it would panic under overflow checks.

## Verification

Each guard was mutated and the suite confirmed red for the intended reason, not merely red:

| mutation | result |
|---|---|
| move `dca_index` after `reserved` | layout test fails (`left: 494, right: 490`) |
| advance the index *after* the next task is compiled | every pre-existing assertion passes; only the compiled-account-list assertion fails |
| drop the borrow fix | `TopOffHntV0` fails with "tries to borrow reference for an account which is already borrowed" |
| force the affordability guard true | `sub_lamports` overflows, the run dies and never reschedules |

The index test asserts the next task's compiled account list rather than `dca_index` alone: an index advanced too late still reports 1 while leaving that task pointed at the drained slot.

The bankrun suite writes the `AutoTopOffV0` directly, since the shortfall needs an account holding barely more than its own rent exemption and `initialize_auto_top_off_v0` funds one properly. It adds a file rather than moving one, so the localnet suite still feeds the CU sampler.

## Swap payer headroom

Checked while building the bankrun control, and healthy: the DCA swap payer `7F9Ncw7pzHZW44j5hwJ38TFU6kQ778Uhtb6sMEAM7PHv` holds 0.598 SOL.

Worth recording why that balance matters, since nothing enforces it. `dc-auto-top` sizes the DCA rent using a `dca_url` length that `tuktuk-dca` does not allocate, so the payer keeps the difference — 250,560 lamports for the mainnet URL — and a system account below its own rent exemption of 890,880 fails the runtime's rent check. The top-up therefore cannot fund a DCA creation from a zero balance on its own; it works because the payer holds a balance already. The practical floor is around 8,331,120 lamports, enough to stay above rent exemption after paying roughly 7,440,240 per DCA.

An abandoned DCA never refunds that rent, so each one costs the payer about 0.0074 SOL. The current balance is roughly 79 DCAs of headroom, against about 13 on `auto_top_off`, so `auto_top_off` is the binding constraint and the guard added here covers it. A monitor threshold well above the floor, say 0.05 SOL, would give ample warning.

### DIFF
--- a/programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
+++ b/programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
@@ -383,9 +383,23 @@ pub fn handler<'info>(
         // Advance the slot before the next task is compiled below, so the next run targets a
         // fresh PDA whether or not this DCA drains and closes. `init` on an occupied PDA fails,
         // and that failure would take the no-reschedule path and stop the HNT leg.
-        ctx.accounts.auto_top_off.load_mut()?.dca_index = dca_index.wrapping_add(1);
+        //
+        // `dca` names the DCA this account is currently feeding. Nothing else advances it, so
+        // leaving it behind would point it at a slot that has since closed or been abandoned.
+        let mut auto_top_off = ctx.accounts.auto_top_off.load_mut()?;
+        auto_top_off.dca_index = dca_index.wrapping_add(1);
+        auto_top_off.dca = ctx.accounts.dca.key();
       }
     }
+  }
+
+  // The crank reward above was sized for two tasks because a DCA was wanted. Every path that
+  // wants one and does not start it returns only the topoff task, so the second reward is
+  // returned rather than left with the queue: those are the paths taken when the account is
+  // short of lamports or dca_mint, which is when it can least afford to leak either.
+  if needs_dca && dca_tasks.is_empty() {
+    ctx.accounts.task_queue.sub_lamports(min_crank_reward)?;
+    auto_top_off_info.add_lamports(min_crank_reward)?;
   }
 
   // Schedule next HNT topoff task
--- a/programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
+++ b/programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
@@ -270,6 +270,7 @@ pub fn handler<'info>(
       let swap_amount_per_order = auto_top_off.dca_swap_amount;
       let interval_seconds = auto_top_off.dca_interval_seconds;
       let dca_signer = auto_top_off.dca_signer;
+      let dca_index = auto_top_off.dca_index;
       drop(auto_top_off);
 
       // Calculate rent needed for DCA account and associated token account
@@ -322,7 +323,7 @@ pub fn handler<'info>(
           seeds,
         ),
         InitializeDcaArgsV0 {
-          index: 0,
+          index: dca_index,
           num_orders: num_orders as u32,
           swap_amount_per_order,
           interval_seconds,
@@ -337,6 +338,11 @@ pub fn handler<'info>(
       // Add DCA tasks to our task list
       let result_tasks = dca_result.get();
       dca_tasks.extend(result_tasks.tasks.clone());
+
+      // Advance the slot before the next task is compiled below, so the next run targets a fresh
+      // PDA whether or not this DCA drains and closes. `init` on an occupied PDA fails, and that
+      // failure would take the no-reschedule path and stop the HNT leg.
+      ctx.accounts.auto_top_off.load_mut()?.dca_index = dca_index.wrapping_add(1);
     }
   }
 
--- a/programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
+++ b/programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
@@ -286,21 +286,44 @@ pub fn handler<'info>(
       let total_rent = dca_rent + ata_rent;
       let rent_needed = total_rent.saturating_sub(ctx.accounts.dca_custom_signer.lamports());
 
+      // Both inputs are checked for the same reason: the run either affords the whole DCA or
+      // does not start it. Anything short reverts the run, and a reverted run never
+      // reschedules, so a shortfall would stop the leg rather than skip a day.
+      //
       // A DCA refunds its rent to the custom signer when it drains and closes, so in steady
       // state rent_needed is 0. An abandoned DCA never refunds, so each replacement draws
-      // fresh rent from here. Debiting past rent exemption fails the whole run, which stops
-      // the leg; skipping the DCA leaves it rescheduling and buys time to refill. Mirrors the
-      // crank-reward check above.
-      let affordable = auto_top_off_info.lamports().saturating_sub(min_rent_exempt) >= rent_needed;
-      if !affordable {
+      // fresh rent from here.
+      let spendable = auto_top_off_info.lamports().saturating_sub(min_rent_exempt);
+      let rent_ok = spendable >= rent_needed;
+      if !rent_ok {
         msg!(
           "Skipping DCA: needs {} lamports rent, {} available above rent exemption",
           rent_needed,
-          auto_top_off_info.lamports().saturating_sub(min_rent_exempt)
+          spendable
         );
       }
 
-      if affordable {
+      // initialize_dca_nested_v0 moves the whole run's USDC up front and takes the order count
+      // as a u32, so a count that does not fit is as unfundable as a balance that falls short.
+      let plan = u32::try_from(num_orders)
+        .ok()
+        .and_then(|orders| {
+          Some((
+            orders,
+            swap_amount_per_order.checked_mul(u64::from(orders))?,
+          ))
+        })
+        .filter(|(_, usdc_needed)| ctx.accounts.dca_mint_account.amount >= *usdc_needed);
+      if plan.is_none() {
+        msg!(
+          "Skipping DCA: {} orders at {} each exceeds the {} dca_mint available",
+          num_orders,
+          swap_amount_per_order,
+          ctx.accounts.dca_mint_account.amount
+        );
+      }
+
+      if let (true, Some((orders, _))) = (rent_ok, plan) {
         // Transfer SOL from auto_top_off to custom signer for DCA rent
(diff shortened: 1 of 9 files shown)


programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
```diff
@@ -161,10 +161,13 @@ pub fn handler<'info>(
 
   let auto_top_off_info = ctx.accounts.auto_top_off.to_account_info();
   let min_rent_exempt = Rent::get()?.minimum_balance(auto_top_off_info.data_len());
-  if auto_top_off_info.lamports() - min_rent_exempt >= total_crank_reward {
+  if auto_top_off_info.lamports().saturating_sub(min_rent_exempt) >= total_crank_reward {
```
bryzettler, line 164: `total_crank_reward` is `min_crank_reward * num_tasks_used` with `num_tasks_used = 1 + needs_dca`, and it is debited to the queue here, before the new affordability guard below. When that guard skips the DCA only one task is returned, so one `min_crank_reward` stays in the queue on every tick the account is short — the branch that is meant to preserve lamports leaks them. `rent_needed` depends only on `dca_url.len()` and the custom signer balance, both readable before this line, so folding affordability into `needs_dca` (or debiting the second reward inside the `else` branch) makes the decision single-point. The pre-existing `num_orders == 0` path has the same leak.

bryzettler, line 164: Same guard in `top_off_dc_v0.rs` (`auto_top_off_acc.lamports() - min_rent_exempt >= total_crank_reward`) still uses plain `-`; under `overflow-checks` it panics instead of taking the unscheduled branch. Worth the same `saturating_sub` while here.

tests/dc-auto-topoff-bankrun.ts
```diff
@@ -0,0 +1,339 @@
...
+  "6HAuqASbHEh4w4REJEUUUCginTLfj1kwCh215ZLtMkrT"
+);
+
+/**
+ * `AutoTopOffV0` field offsets, discriminator included. Reaching the rent-shortfall branch
+ * needs an account holding barely more than its own rent exemption, and
+ * `initialize_auto_top_off_v0` funds one properly, so the state is written directly rather
+ * than produced. `state.rs`'s layout test pins these same offsets from the Rust side.
+ */
+const OFF = {
```
bryzettler: This offset table, `discriminator()`, and the `put`/`write*` calls re-implement Anchor's account encoder. `program.coder.accounts.encode("autoTopOffV0", { ... })` produces the same 944 bytes (the struct is self-padded, and `state.rs`'s layout test already pins that), and `tests/mini-fanout-bankrun.ts` uses it for the same job. That removes the second copy of the layout, the `crypto` import, and the entries here that nothing reads (`dataCredits`, `subDao`, `dcMint`, `dao`, `dcAccount`, `escrowAccount`, `circuitBreaker`, `dca`).

tests/dc-auto-topoff-bankrun.ts
```diff
@@ -0,0 +1,339 @@
...
+  let taskQueue: PublicKey;
+  let hntMint: PublicKey;
+  let dcaMint: PublicKey;
+
+  const queueAuthority = queueAuthorityKey()[0];
+
+  const send = (instructions: anchor.web3.TransactionInstruction[]) =>
+    provider.sendAndConfirm(new Transaction().add(...instructions));
+
+  async function createMint(decimals: number): Promise<PublicKey> {
```
bryzettler, line 92: `createMint` and `ataWith` duplicate `createMint` / `createAtaAndMint` from `@helium/spl-utils`, which `tests/dc-auto-topoff.ts` already uses with the same provider shape; the bankrun provider supports the connection calls they need.

bryzettler (commented):
One lamport-accounting gap in the new skip path, one stale field, and a few duplication cleanups in the handler and the bankrun fixture. Inline comments below.

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
```diff
@@ -282,61 +286,105 @@ pub fn handler<'info>(
...
+      //
+      // A DCA refunds its rent to the custom signer when it drains and closes, so in steady
+      // state rent_needed is 0. An abandoned DCA never refunds, so each replacement draws
+      // fresh rent from here.
+      let spendable = auto_top_off_info.lamports().saturating_sub(min_rent_exempt);
+      let rent_ok = spendable >= rent_needed;
+      if !rent_ok {
+        msg!(
+          "Skipping DCA: needs {} lamports rent, {} available above rent exemption",
+          rent_needed,
```
bryzettler, line 301: The spendable-lamports expression is computed twice here (and once more in the crank-reward check above). One binding keeps the check and the log from disagreeing. (Merging the `if !affordable` / `if affordable` pair into one `if`/`else` would also help, but that range crosses deleted lines so it isn't a one-click suggestion.)

```suggestion
      let spendable = auto_top_off_info.lamports().saturating_sub(min_rent_exempt);
      let affordable = spendable >= rent_needed;
      if !affordable {
        msg!(
          "Skipping DCA: needs {} lamports rent, {} available above rent exemption",
          rent_needed,
          spendable
        );
      }
```

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
```diff
@@ -282,61 +286,105 @@ pub fn handler<'info>(
...
+                output_price_oracle: ctx.accounts.hnt_price_oracle.to_account_info(),
+                input_account: ctx.accounts.dca_input_account.to_account_info(),
+                dca_payer_account: ctx.accounts.dca_mint_account.to_account_info(),
+                task_queue: ctx.accounts.task_queue.to_account_info(),
+                system_program: ctx.accounts.system_program.to_account_info(),
+                associated_token_program: ctx.accounts.associated_token_program.to_account_info(),
+                token_program: ctx.accounts.token_program.to_account_info(),
+              },
+              task: ctx.remaining_accounts[1].to_account_info(),
             },
```
bryzettler, line 363: `AutoTopOffV0.dca` is written once in `initialize_auto_top_off_v0` as the slot-0 PDA and nothing advances it when `dca_index` does, so after the first DCA it points at a slot that may be closed or orphaned. Setting it here keeps the field on the DCA that was just created.

```suggestion
        let mut auto_top_off = ctx.accounts.auto_top_off.load_mut()?;
        auto_top_off.dca_index = dca_index.wrapping_add(1);
        auto_top_off.dca = ctx.accounts.dca.key();
```

bryzettler (commented):
Replacing two of the earlier comments with committable suggestions.

madninja:
Thanks — five of six were right and are applied in `3b9ed0643`. Dispositions, with the check behind each:

**Crank-reward leak (applied).** Confirmed: `num_tasks_used = 1 + needs_dca` is debited at line 164 before any affordability check, and every skip path returns one task. Rather than fold affordability into `needs_dca` — the USDC half of the decision needs `num_orders`, so that would mean hoisting the oracle maths above the debit — the second reward is returned at the single point that covers every such path, including the pre-existing zero-order one. Pinned by asserting the account's lamports move by exactly one `min_crank_reward` on the skip; disabling the refund fails that assertion.

**`top_off_dc_v0` plain subtraction (applied).** Now saturates, same as its counterpart.

**`dca` stale (applied).** Confirmed the field is written only in `initialize_auto_top_off_v0` and never advanced. It now follows the DCA being created, asserted against the slot-0 PDA in the funded case; setting it to `Pubkey::default()` fails that assertion.

**Coder instead of the offset table (applied).** Verified before swapping: `coder.accounts.encode` produces exactly 944 bytes, matching the hand-rolled layout, because the struct is self-padded. Note the account key is `autoTopOffV0` — the camelCase in the suggestion is right and `AutoTopOffV0` throws `Unknown account`, which is what `program.coder` normalises to.

**Spendable computed twice (already fixed).** That suggestion is against an earlier commit; `2704238e4` had already introduced the single `spendable` binding while adding the USDC check.

**spl-utils mint/ATA helpers — not applied.** This one does not hold. `createAtaAndMintInstructions` checks for the account with `provider.connection.getAccountInfo`, and `BankrunConnectionProxy.getAccountInfo` *throws* `Could not find <pubkey>` for a missing account rather than returning null, so all three tests fail with it. That is why `mini-fanout-bankrun.ts` also builds these locally — it uses the *idempotent* create instruction, which needs no existence check. The local helpers stay, with that reason recorded in the file so the next reader does not retry it.

bryzettler (approved):

PR #1276, opened 2026-08-27
fix: skip sim-derived loaded-accounts-data-size limit on wallet-signed txs

bryzettler:
## Problem

blockchain-api preflight sims failed with `MaxLoadedAccountsDataSizeExceeded` on `governance.assignProxies` (prod `my-helium` logs, 2026-08-27 12:13Z). Decoding the failing message shows two Lighthouse (`L2TExMFKdjpN9kozasaurPirfHy9P8sbXoAN1qA3S95`) instructions that this repo never adds — the user's wallet appended them at signing time. Lighthouse loads ~424 KiB of program data, which exceeds the ~10% headroom of the sim-derived `setLoadedAccountsDataSizeLimit` (2,424,832 bytes) we attached at build time.

Any tx we return unsigned to a wallet that injects guard ixs (Phantom, Solflare) is exposed.

## Fix

- `@helium/spl-utils`: add `deriveLoadedAccountsDataSizeLimit` (default `true`) to `withPriorityFees` and `batchInstructionsToTxsWithPriorityFee`. When `false`, the sim runs without a data-size ceiling and no data-size ix is emitted. An explicit `loadedAccountsDataSizeLimit` is still honored.
- blockchain-api: set it `false` at both build chokepoints — `buildVersionedTransaction` and every direct `batchInstructionsToTxsWithPriorityFee` call. Every tx built there is wallet-signed or wallet-co-signed (including `reward-contract/claim` and `welcomePacks/claim`, which pass a server fee payer), so all of them now run under the runtime 64 MiB default. Only callers outside blockchain-api that never hand a tx to a wallet keep the derived limit.
- distributor-oracle `client.ts` and `useSplitPosition` also build wallet-signed txs; both opt out.
- `batchInstructionsToTxsWithPriorityFee` size probe omits the data-size ix when the opt-out guarantees the real tx won't carry one.
- `distributor-oracle` `formTransaction` also opts out, matching `formBulkTransactions`.

## Why not floor wallet-signed txs up to the 16 MiB default instead?

Considered. A fixed ceiling is a guess about what wallets append, and the set of guards is outside our control. The fee saving from a 16 MiB request vs the 64 MiB runtime default is small on these txs; a failed tx costs the user more. So wallet-signed txs get no data-size ix at all.

## Verification

- `tests/priority-fees.ts`: 27 passing, including two new cases for the opt-out (no ix in sim or output; explicit limit still emitted).

- `spl-utils` `tsc -b` clean.
- blockchain-api `tsc --noEmit`: same 6 pre-existing `tests/e2e` errors before and after; none from this change.

### DIFF
--- a/.changeset/wallet-signed-data-size-limit.md
+++ b/.changeset/wallet-signed-data-size-limit.md
@@ -1,4 +1,5 @@
 ---
+"@helium/blockchain-api": patch
 "@helium/spl-utils": patch
 "@helium/distributor-oracle": patch
 "@helium/voter-stake-registry-hooks": patch
--- a/.changeset/wallet-signed-data-size-limit.md
+++ b/.changeset/wallet-signed-data-size-limit.md
@@ -1,6 +1,5 @@
 ---
 "@helium/spl-utils": patch
-"@helium/blockchain-api": patch
 ---
 
 Add `deriveLoadedAccountsDataSizeLimit` opt-out to `withPriorityFees` and `batchInstructionsToTxsWithPriorityFee`, and disable sim-derived loaded-accounts-data-size limits on all blockchain-api wallet-signed transactions. Wallets append guard instructions (Lighthouse) after sizing, which exceeded the derived limit and failed with `MaxLoadedAccountsDataSizeExceeded`.
--- a/.changeset/wallet-signed-data-size-limit.md
+++ b/.changeset/wallet-signed-data-size-limit.md
@@ -1,0 +1,6 @@
+---
+"@helium/spl-utils": patch
+"@helium/blockchain-api": patch
+---
+
+Add `deriveLoadedAccountsDataSizeLimit` opt-out to `withPriorityFees` and `batchInstructionsToTxsWithPriorityFee`, and disable sim-derived loaded-accounts-data-size limits on all blockchain-api wallet-signed transactions. Wallets append guard instructions (Lighthouse) after sizing, which exceeded the derived limit and failed with `MaxLoadedAccountsDataSizeExceeded`.
--- a/.changeset/wallet-signed-data-size-limit.md
+++ b/.changeset/wallet-signed-data-size-limit.md
@@ -1,5 +1,7 @@
 ---
 "@helium/spl-utils": patch
+"@helium/distributor-oracle": patch
+"@helium/voter-stake-registry-hooks": patch
 ---
 
-Add `deriveLoadedAccountsDataSizeLimit` opt-out to `withPriorityFees` and `batchInstructionsToTxsWithPriorityFee`, and disable sim-derived loaded-accounts-data-size limits on all blockchain-api wallet-signed transactions. Wallets append guard instructions (Lighthouse) after sizing, which exceeded the derived limit and failed with `MaxLoadedAccountsDataSizeExceeded`.
+Add `deriveLoadedAccountsDataSizeLimit` opt-out to `withPriorityFees` and `batchInstructionsToTxsWithPriorityFee`, and disable sim-derived loaded-accounts-data-size limits on wallet-signed transactions in blockchain-api, distributor-oracle, and voter-stake-registry-hooks. Wallets append guard instructions (Lighthouse) after sizing, which exceeded the derived limit and failed with `MaxLoadedAccountsDataSizeExceeded`.
--- a/packages/blockchain-api/src/lib/utils/build-transaction.ts
+++ b/packages/blockchain-api/src/lib/utils/build-transaction.ts
@@ -74,6 +74,9 @@ export async function buildVersionedTransaction({
       ...draftWithLuts,
       addressLookupTables,
       connection,
+      // The wallet signs this tx and may append guard ixs (Lighthouse) that
+      // load more account data than a sim-derived limit allows.
+      deriveLoadedAccountsDataSizeLimit: false,
     });
   } catch (error) {
     console.warn(
--- a/packages/blockchain-api/src/server/api/routers/hotspots/procedures/automation-transaction.ts
+++ b/packages/blockchain-api/src/server/api/routers/hotspots/procedures/automation-transaction.ts
@@ -97,6 +97,8 @@ export const buildAutomationTransactionResponse = async ({
     await batchInstructionsToTxsWithPriorityFee(provider, instructions, {
       addressLookupTableAddresses: [getHeliumLookupTable()],
       commitment: "finalized",
+      // Wallet-signed: guard ixs may be appended (see withPriorityFees).
+      deriveLoadedAccountsDataSizeLimit: false,
     })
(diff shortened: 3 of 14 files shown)


madninja:
## Review

The fix is correct and the diagnosis holds. Traced `governance.assignProxies` -> `buildBatchedTransactions` -> `buildVersionedTransaction`, which is the single chokepoint for all 30 `buildVersionedTransaction` call sites, so the reported failure is covered. All 3 `migrate.ts` batch calls share `batchOpts`, so none was missed. The spl-utils logic is coherent end to end: `simCeiling` undefined -> `simValidatesDataSizeCeiling` false -> no ix in the sim tx -> the derive block is gated off -> `resolvedDataSizeLimit` stays undefined -> no ix emitted. An explicit `loadedAccountsDataSizeLimit` still wins, as claimed.

Nothing below blocks merge on correctness. Finding 1 is the one worth closing first: the harness already exists, so the cost is minutes, and without it the flag's default is unpinned.

### Findings

| # | Finding | Criticality | Code impact | Recommended action |
|---|---|---|---|---|
| 1 | **No test for the new branch.** `tests/priority-fees.ts` has 20+ cases covering every other branch of this exact logic, with the mock-sim harness already built. Flipping the new default to `true`, or deleting the `&& simCeiling != null` gate, goes green. | Medium | ~15 lines, 2 cases | Add (a) flag `false` + a sim reporting `loadedAccountsDataSize` -> no data-size ix in *either* the sim tx or the returned ixs; (b) flag `false` + an explicit limit -> explicit still emitted. |
| 2 | **The same exposure remains in two client-side builders.** `packages/distributor-oracle/src/client.ts:548` (its own comment reads `// unsigned txs`) and `packages/voter-stake-registry-hooks/src/hooks/useSplitPosition.ts:150` build txs a browser wallet signs, and both still derive the limit. The PR body names the class ("any tx we return unsigned to a wallet"); the fix covers only blockchain-api. | Medium | 2 one-line additions | Set the flag at both, or state the deferral in the PR body. |
| 3 | **Design alternative worth a deliberate answer.** The failing limit was 2,424,832 bytes; Lighthouse needs ~434 KiB more. Flooring the wallet-signed request *up* to `DEFAULT_LOADED_ACCOUNTS_DATA_SIZE_LIMIT` (16 MiB) would absorb ~13.6 MiB of appended guards and keep the 75% saving that constant's own comment advertises. As written the PR forfeits SIMD-0553 sizing across essentially the whole user-facing API surface. | Low-Medium | Design | Answer it explicitly. "Wallets may append arbitrary guards, so no derived ceiling is safe" is a defensible counter; it just wants to be a stated choice rather than a side effect. |
| 4 | **The batch size probe now over-measures by one ix.** `spl-utils/src/transaction.ts:~925` unconditionally puts `setLoadedAccountsDataSizeLimit` into the probe, under a comment reading "Probe must match the real tx's ix count". With the flag `false` the real tx omits that ix. The direction is conservative (splits early), but in `migrate.ts` `fitGroupsWithinLimit` an over-count can push a group past `MAX_JITO_BUNDLE_TXS` and defer it to the next call. | Low | 1 line | Gate the probe ix on the same condition, or fix the now-false comment. |
| 5 | PR body says "server-signed paths keep the optimization", but `buildVersionedTransaction` accepts `signers` and hardcodes the flag internally, so `reward-contract/procedures/claim.ts:224` and `welcomePacks/procedures/claim.ts:105` (both passing `feePayerWallet`) also lose it. Behavior is right, since those are still wallet-co-signed; the sentence is just imprecise. | Low | Body text | Reword. |

### Five dimensions

- **Security.** Clean. No auth or authz surface. Relaxing a self-imposed resource ceiling cannot admit a tx the runtime would reject, and the 64 MiB runtime default still bounds it.
- **Simplicity.** Clean. One optional boolean threaded through two functions; no layers worth collapsing.
- **DRY.** Minor. The `// Wallet-signed: guard ixs may be appended` comment is copy-pasted at 7 call sites. That repetition is the signal the option wants to be the default for blockchain-api's batch calls (a thin local wrapper) rather than a per-site opt-out that call site 8 will forget.
- **YAGNI.** Clean. Both new params have real consumers; nothing speculative.
- **Coverage.** This is finding 1.

madninja (approved):
All five findings addressed. Approving.

### What I verified, not just read

Measured in a clean worktree at `158c0e053`, not the main checkout: baseline `tests/priority-fees.ts` is 27 passing, matching the body's claim.

I then mutated the guards the *fixes themselves* introduced, since that is the least-reviewed code in any round:

| Mutation | Verdict |
|---|---|
| `withPriorityFees` default `true` -> `false` | **CAUGHT** (7 failing) |
| Drop `&& simCeiling != null` from the derive gate | **CAUGHT** (the new opt-out test) |
| Probe gate `!== false` -> `=== true` | not caught, equivalent (below) |
| Probe gate: drop the `loadedAccountsDataSizeLimit != null` clause | not caught, equivalent (below) |
| Drop `simCeiling != null` from `simValidatesDataSizeCeiling` | not caught, equivalent (below) |

The default being pinned is the important one: that was the substance of finding 1, and it is genuinely closed rather than merely covered by a new test file.

On the three not-caught verdicts, I checked whether they are real gaps rather than logging them:

- **Both probe-gate mutations are unobservable, and I measured why.** The data-size ix costs exactly 8 bytes compiled. The size check adds `64 * signatures.length` on top of a `serialize()` that already includes the signature, so every probe carries a 64-byte cushion that swallows an 8-byte discrepancy. Sweeping `maxTxSize` 1100..1232 across 266 drafts, no draft exceeded its own limit either mutated or clean. So the probe gate is defensive, not load-bearing, and a test for it would be asserting a difference that cannot manifest.
- **`simCeiling != null` in `simValidatesDataSizeCeiling` is redundant** given the derive gate below it. With the flag `false` its only reachable effect is one redundant re-simulation in an error path; no emitted ix changes. Fine to keep as intent documentation.

Also clean: `spl-utils` `tsc -b`, `distributor-oracle` and `voter-stake-registry-hooks` `tsc --noEmit`, and CI at 67 pass / 3 skips (the skips are the devnet-deploy workflow, correctly inapplicable here).

Swept every remaining call site that does not opt out. All are server-signed (helium-admin-cli, crons), a hardware Ledger signer (migration-service), or test code. Nothing wallet-signed is left uncovered.

### Two non-blocking notes

1. The body's "so the probe matches the real ix count" is slightly stronger than the code holds: in the derive path with a *failing* sim, the real tx carries no data-size ix while the probe does. Harmless because of the 64-byte cushion above, but this sentence becomes the squash commit message, so it is worth softening to "omits it when the opt-out guarantees the real tx won't carry one".
2. `formTransaction` does not forward the new flag, unlike `formBulkTransactions`. Every in-repo caller is server-signed so there is no live exposure, but a third-party app following the `getting_started.md` claim pattern would still get a derived limit. Worth a follow-up rather than a change here.

PR #1279, opened 2026-08-29
Fee payer options for update-info and transfers, and verification fixes

madninja:
Three related changes to who pays for a transaction, who is allowed to ask for
one, and what the service verifies before it acts.

## Owner-paid updates (`20ac79e`)

`POST /hotspots/update-info` relays the onboarding server, which co-signs so the
maker burns the DC. A caller asserting on behalf of the owner needs the owner to
pay. `feePayer` defaults to `"maker"` and that path is unchanged; `"owner"`
builds `UpdateIotInfoV0` / `UpdateMobileInfoV0` here, with `payer`,
`dc_fee_payer` and `dc_burner` all the owner, and accounts for the DC burn and
any `mobile_info` resize in the balance check and reported fee. A hotspot whose
leaf delegate differs from its owner is refused with a 409, since the update
requires `owner == delegate == hotspot_owner`.

## Third-party fee payer on transfers (`754e1c1`)

`POST /tokens/transfer` had one `walletAddress` acting as both source authority
and fee payer, so a caller draining a wallet to zero could not use it: the
wallet has no SOL left for the fee or for the recipient's ATA rent. The optional
`feePayer` keeps `walletAddress` as the authority and moves the fee and the ATA
funding to a separate account. Absent, behaviour is unchanged.

Two consequences worth knowing. The transaction then requires **two
signatures**, and the response envelope has no field announcing that — a caller
knows because it sent `feePayer`. And `submit` derives a batch's payer from
account 0, so these transfers are recorded under the fee payer rather than the
wallet. `feePayer` combined with `multisig` is refused: in propose mode the
vault is the authority and the proposing member is already the outer fee payer,
so both roles are taken.

## Verify and authorize before acting (`6be95fd`)

From a security review of the service. Each was confirmed in source.

**The service spent its own signing key on unverified input.** Both claim
endpoints loaded `FEE_PAYER_WALLET_PATH` and signed before checking the
delegate signature they were handed, so any caller could make the service sign
and pay for a transaction the program would reject. The signature is now
verified against the pack owner over the message the program builds, before the
key is loaded, and the approval window is bounded by the program's own
`MAX_CLAIM_APPROVAL_SECONDS`.

**`updateRewardsDestination`, `createSplit` and `deleteSplit` had no ownership
check**, while four sibling endpoints each hand-rolled one. All six now call a
single `assertAssetOwner`. Without it the service assembled a ready-to-sign,
caller-funded transaction redirecting any hotspot's future rewards. The
signature of the true owner was always still required, so this was never direct
theft; it removes the service as a factory for them. `createSplit` and
`deleteSplit` also now read the asset on-chain rather than trusting a database
row that can lag.

**Submission attributed a batch to an unverified fee payer.** `payer` came from
account 0 of the submitted bytes with no signature check, and submission uses
`skipPreflight`, so history rows could be written against any wallet with
attacker-chosen metadata. The fee payer's signature is now verified over the
message. It checks that signature specifically rather than the signature count,
so the two-signature transfers above still pass.

**Three visibility fixes.** The swap output account now appears in the
description, metadata and tag, so two swaps differing only in destination no
longer produce the same tag. `dc mint` names a recipient that differs from the
owner. And the request logger moved from `JSON.stringify(input)` to a
scalar-only allowlist, having been writing bank account and routing numbers to
stdout.

## Coverage

179 unit tests, from a baseline of 122. Guards are mutation-tested one at a
time: 16 of 16 caught for the transfer fee payer, 24 of 30 for owner-paid
updates, 23 of 25 here.

The gaps are stated rather than smoothed over. The six uncaught on owner-paid
updates are handler consumption sites no unit test can reach; an e2e case covers
them and has not been run. The two uncaught here are equivalent mutations,
measured across 264 input combinations with zero behavioural difference rather
than assumed. And because `@/` aliases do not resolve under the unit runner,
the ownership and signature guards are pinned at their call sites by a source
scan, which would not notice a guard called with the wrong arguments.

No part of this has been exercised against a live cluster.

## Also worth a maintainer's decision

A swap `destinationTokenAccount` not owned by the signer is now visible but
still permitted; refusing it is a policy call. `reward-contract/delete.ts`
compares against the reward contract's delegate rather than the cNFT owner, so
it was deliberately left out of the ownership consolidation. And this branch
changes the published `@helium/blockchain-api` client package without a
changeset.

### DIFF
--- a/packages/blockchain-api/src/lib/utils/asset-ownership.ts
+++ b/packages/blockchain-api/src/lib/utils/asset-ownership.ts
@@ -1,0 +1,43 @@
+import type { Asset } from "@helium/spl-utils";
+
+/** The error builder an ownership refusal is raised through. */
+type AssetOwnerErrors = {
+  UNAUTHORIZED: (opts: { message: string }) => Error;
+};
+
+/**
+ * A cNFT's current owner in base58. The asset endpoints hand the field back
+ * either already decoded or as the string it arrived as, so both are accepted.
+ */
+export function assetOwnerAddress(asset: Asset): string {
+  const { owner } = asset.ownership;
+  return typeof owner === "string" ? owner : owner.toBase58();
+}
+
+/**
+ * Refuse a caller that is not the cNFT's owner, and hand back the owner it
+ * checked so the address is derived once.
+ *
+ * `expectedOwner` is the authority the endpoint builds its instruction for: the
+ * caller's own wallet, or the vault of the multisig it proposes through, which
+ * is the owner on chain while the caller is only a member of it. `message` is
+ * the endpoint's own wording, because the endpoints name different things the
+ * wallet failed to own.
+ */
+export function assertAssetOwner({
+  asset,
+  expectedOwner,
+  message,
+  errors,
+}: {
+  asset: Asset;
+  expectedOwner: string;
+  message: string;
+  errors: AssetOwnerErrors;
+}): string {
+  const ownerAddress = assetOwnerAddress(asset);
+  if (ownerAddress !== expectedOwner) {
+    throw errors.UNAUTHORIZED({ message });
+  }
+  return ownerAddress;
+}
--- a/packages/blockchain-api/src/lib/utils/asset-ownership.ts
+++ b/packages/blockchain-api/src/lib/utils/asset-ownership.ts
@@ -1,8 +1,14 @@
-import type { Asset } from "@helium/spl-utils";
+import { getAsset, type Asset } from "@helium/spl-utils";
+import type { PublicKey } from "@solana/web3.js";
 
 /** The error builder an ownership refusal is raised through. */
 type AssetOwnerErrors = {
   UNAUTHORIZED: (opts: { message: string }) => Error;
+};
+
+/** `fetchOwnedAsset` also refuses an asset the endpoint does not know. */
+type OwnedAssetErrors = AssetOwnerErrors & {
+  NOT_FOUND: (opts: { message: string }) => Error;
 };
 
 /**
@@ -40,4 +46,36 @@ export function assertAssetOwner({
     throw errors.UNAUTHORIZED({ message });
   }
   return ownerAddress;
+}
+
+/**
+ * Fetch a cNFT and refuse the caller unless `expectedOwner` owns it, handing
+ * back the asset so the endpoint never pays for a second DAS round-trip.
+ *
+ * `assetEndpoint` comes in rather than being read from the environment here so
+ * this module stays importable under the unit runner, which resolves no `@/`
+ * aliases; endpoints pass `getAssetEndpoint()` from `@/lib/solana`.
+ * `getAssetFn` exists for tests.
+ */
+export async function fetchOwnedAsset({
+  assetEndpoint,
+  assetId,
+  expectedOwner,
+  message,
+  errors,
+  getAssetFn = getAsset,
+}: {
+  assetEndpoint: string;
+  assetId: PublicKey;
+  expectedOwner: string;
+  message: string;
+  errors: OwnedAssetErrors;
+  getAssetFn?: (url: string, assetId: PublicKey) => Promise<Asset | undefined>;
+}): Promise<Asset> {
+  const asset = await getAssetFn(assetEndpoint, assetId);
+  if (!asset) {
+    throw errors.NOT_FOUND({ message: "Asset not found" });
+  }
+  assertAssetOwner({ asset, expectedOwner, message, errors });
+  return asset;
 }
--- a/packages/blockchain-api/src/server/api/routers/data-credits/procedures/mint.ts
+++ b/packages/blockchain-api/src/server/api/routers/data-credits/procedures/mint.ts
@@ -55,7 +55,7 @@ export const mint = publicProcedure.dataCredits.mint.handler(
     const useJito = shouldUseJitoBundle(txs.length, getCluster());
     const txFees = await getTotalTransactionFees(
       connection,
-      txs.map((t) => t.tx)
+      txs.map((t) => t.tx),
     );
     const jitoTipCost = useJito ? getJitoTipAmountLamports() : 0;
 
@@ -66,13 +66,13 @@ export const mint = publicProcedure.dataCredits.mint.handler(
     const recipientDcAta = getAssociatedTokenAddressSync(
       DC_MINT,
       recipientPubkey,
-      true
+      true,
     );
     const recipientDcAtaInfo = await connection.getAccountInfo(recipientDcAta);
     const ataRent = recipientDcAtaInfo ? 0 : RENT_COSTS.ATA;
     const requiredBalance = calculateRequiredBalance(
       txFees + jitoTipCost,
-      ataRent
+      ataRent,
     );
 
     const ownerPubkey = new PublicKey(owner);
@@ -89,6 +89,7 @@ export const mint = publicProcedure.dataCredits.mint.handler(
       userAddress: owner,
       dcAmount: dcAmount || undefined,
       hntAmount: hntAmount || undefined,
+      recipient: recipient && recipient !== owner ? recipient : undefined,
     });
 
     // The credits land on the recipient rather than the signer whenever one is
@@ -135,5 +136,5 @@ export const mint = publicProcedure.dataCredits.mint.handler(
         recipient: recipient || undefined,
       },
     };
-  }
+  },
 );
--- a/packages/blockchain-api/src/server/api/routers/data-credits/procedures/mint.ts
+++ b/packages/blockchain-api/src/server/api/routers/data-credits/procedures/mint.ts
@@ -91,6 +91,15 @@ export const mint = publicProcedure.dataCredits.mint.handler(
       hntAmount: hntAmount || undefined,
     });
 
+    // The credits land on the recipient rather than the signer whenever one is
+    // named, so the signer is told which account before signing for it.
+    const recipientSuffix =
+      recipient && recipient !== owner ? ` to ${recipient}` : "";
+    const description =
+      (dcAmount
+        ? `Mint ${dcAmount} data credits`
+        : `Burn ${hntAmount} HNT bones for data credits`) + recipientSuffix;
+
     const transactions = txs.map((t) => {
       if (t.signers.length > 0) {
         t.tx.sign(t.signers);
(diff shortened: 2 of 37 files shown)


packages/blockchain-api/src/server/api/routers/data-credits/procedures/mint.ts
```diff
@@ -91,6 +91,15 @@ export const mint = publicProcedure.dataCredits.mint.handler(
       hntAmount: hntAmount || undefined,
```
bryzettler, line 91: The tag leaves `recipient` out, so two pending mints of the same amount to different recipients hash to the same tag. Pending batches are unique on `(tag, payer)`, so the second is refused as a duplicate of the first even though it is a different transaction that lands the credits on a different account. That is the same reason the swap tag in this PR gained `destinationTokenAccount`, and the condition mirrors the `recipientSuffix` logic just below.

```suggestion
      hntAmount: hntAmount || undefined,
      recipient: recipient && recipient !== owner ? recipient : undefined,
```

packages/blockchain-api/src/server/api/routers/hotspots/procedures/updateHotspotInfo.ts
```diff
@@ -189,71 +306,187 @@ export const updateHotspotInfo =
...
+            remainingAccounts,
+            update: {
+              network: "iot",
+              location: newLocation,
+              elevation: toOnChainElevation(input.elevation),
+              gain: toOnChainGain(input.gain),
             },
           });
+        } else {
+          const mergedDeploymentInfo = mergeDeploymentInfo(
```
bryzettler, line 351: `mergeDeploymentInfo`'s CBRS-to-CBRS branch returns `cbrsInfoV0` as a bare `radioInfos` array where the struct shape `{ cbrsInfoV0: { radioInfos } }` is expected. Compare `inputToOnboardingDeploymentInfo` above it, which returns `{ cbrsInfoV0: cbrs }` with `cbrs = { radioInfos }`.

The maker path handed this value straight to the onboarding server, so the shape never had to encode locally. The owner path added here feeds it into `estimateMobileResizeRent`, which calls `program.coder.accounts.encode("mobileHotspotInfoV0", ...)`, and a bare array does not encode as that enum variant.

Replacing lines 146-154:

```ts
  if (existingType === "CBRS" && newType === "CBRS" && existing.cbrsInfoV0) {
    const { type: _, ...cbrs } = newInfo;
    return {
      cbrsInfoV0: {
        radioInfos:
          cbrs.radioInfos !== undefined
            ? cbrs.radioInfos
            : existing.cbrsInfoV0.radioInfos,
      },
    };
  }
```

bryzettler (commented):
Two correctness gaps in the new paths (the mint tag missing `recipient`, the CBRS merge shape the owner path now encodes), one unit-contract question on `gain`, and a handful of reuse/efficiency cleanups. Details inline; the two `suggestion` blocks apply as-is.

packages/blockchain-api/src/server/api/routers/hotspots/procedures/updateHotspotInfo.ts
```diff
@@ -189,71 +306,187 @@ export const updateHotspotInfo =
...
+            owner,
+            entityKey,
+            merkleTree: accounts.merkleTree,
+            proofArgs: args,
+            remainingAccounts,
+            update: {
+              network: "iot",
+              location: newLocation,
+              elevation: toOnChainElevation(input.elevation),
+              gain: toOnChainGain(input.gain),
```
bryzettler: The two fee-payer modes give one input field two different units.

This owner path stores `toOnChainGain(input.gain)`, which is `Math.trunc(input.gain * 10)` — tenths of a dBi, as the on-chain `i32` wants. The maker path below forwards `input.gain` unscaled to `onboardingClient.updateIotMetadata`, and `@helium/onboarding`'s own `HotspotOnboardingUtil` multiplies a decimal gain by 10 (`Math.round((decimalGain || 0) * 10.0)`) before calling that same method, so that wire field is already tenths too. The same request body therefore writes a gain 10x apart depending on `feePayer`.

One of the two branches needs to change so `gain` means one thing. Which one depends on what existing callers of this endpoint send. `onboardDataOnlyHotspot.ts` in this service treats its own `gain` input as dBi and scales it (`gain !== undefined ? Math.trunc(gain * 10) : null`, under the comment "On-chain gain is tenths of a dBi"), which agrees with the owner path here.

packages/blockchain-api/src/server/api/routers/hotspots/procedures/createSplit.ts
```diff
@@ -89,6 +91,22 @@ export const createSplit = publicProcedure.hotspots.createSplit.handler(
       });
     }
 
+    // Only the owner may redirect a hotspot's rewards, so the check comes
+    // before anything is looked up or built on the caller's behalf.
+    const asset = await getAsset(
```
bryzettler: The guard fetches the asset here, and then `initializeCompressionRecipient` (line 131) and `updateCompressionDestination` (line 220) each re-fetch the same asset through `proofArgsAndAccounts`. Both SDK functions take a `getAssetFn` override, so the asset already in hand can serve and the two extra DAS round-trips go away. This PR already uses exactly that pattern in `updateHotspotInfo.ts`: "reuse it rather than pay for a second DAS round-trip".

One field added to each of those two calls:

```ts
            getAssetFn: async () => asset,
```

packages/blockchain-api/src/server/api/routers/hotspots/procedures/deleteSplit.ts
```diff
@@ -47,6 +50,22 @@ export const deleteSplit = publicProcedure.hotspots.deleteSplit.handler(
 
     const hotspotName = animalName(hotspotPubkey);
 
+    // Only the owner may remove a hotspot's split, so the check comes before
+    // the split is looked up: an unauthorized caller learns nothing about it.
+    const asset = await getAsset(
```
bryzettler: Same as in `createSplit.ts`: the guard fetches the asset here, then `updateCompressionDestination` (line 136) re-fetches it through `proofArgsAndAccounts`. It takes a `getAssetFn` override, so the asset already in hand can serve.

```ts
        getAssetFn: async () => asset,
```

packages/blockchain-api/src/lib/utils/asset-ownership.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Refuse a caller that is not the cNFT's owner, and hand back the owner it
+ * checked so the address is derived once.
+ *
+ * `expectedOwner` is the authority the endpoint builds its instruction for: the
+ * caller's own wallet, or the vault of the multisig it proposes through, which
+ * is the owner on chain while the caller is only a member of it. `message` is
+ * the endpoint's own wording, because the endpoints name different things the
+ * wallet failed to own.
+ */
+export function assertAssetOwner({
```
bryzettler, line 33: `assertAssetOwner` collapses the owner comparison well, but the step before it is still copied around. `createSplit.ts`, `deleteSplit.ts` and `updateHotspotInfo.ts` each spell out the same triple — `getAsset(...)`, `if (!asset) throw errors.NOT_FOUND({ message: "Asset not found" })`, then `assertAssetOwner(...)` — and `reward-contract/procedures/create.ts` repeats the last two over an asset that reaches it from `proofArgsAndAccounts`. The three `getAsset` sites also disagree on the endpoint fallback: `env.ASSET_ENDPOINT || env.SOLANA_RPC_URL` in the two split procedures, `env.ASSET_ENDPOINT || connection.rpcEndpoint` in `updateHotspotInfo.ts`.

A second helper in this file — one that fetches, throws NOT_FOUND, asserts and returns the asset, say `fetchOwnedAsset` — would collapse each of those sites to a single call and settle the endpoint fallback in one place. `hotspot-cnft.ts` and `updateRewardsDestination.ts` already hold the asset by the time they check, so they would keep calling `assertAssetOwner` directly.

packages/blockchain-api/src/server/api/routers/hotspots/procedures/updateHotspotInfo.ts
```diff
@@ -130,6 +156,101 @@ function mergeDeploymentInfo(
...
+  network,
+  info,
+  newLocation,
+}: {
+  program: HemProgram;
+  network: "iot" | "mobile";
+  info: HotspotInfo;
+  newLocation: BN | null;
+}): Promise<BN> {
+  const config = await (
```
bryzettler, line 241: `locationAssertDcFee` returns `new BN(0)` whenever `newLocation` is null, so this `rewardableEntityConfigV0` fetch is spent for nothing on every owner-paid update that asserts no location — one that only changes gain, elevation or deployment info. Returning early keeps the same zero fee and skips the account fetch.

```suggestion
  if (!newLocation) return new BN(0);

  const config = await (
```

packages/blockchain-api/src/server/api/routers/hotspots/procedures/updateHotspotInfo.ts
```diff
@@ -189,71 +306,187 @@ export const updateHotspotInfo =
...
+            // reuse it rather than pay for a second DAS round-trip.
+            getAssetFn: async () => asset,
+          });
+
+        const newLocation = h3
+          ? new BN(network === "iot" ? h3.iot : h3.mobile, "hex")
+          : null;
+        const info = await getHotspotInfo(provider, entityKey);
+
+        let updateIx;
```
bryzettler: Both branches spell out the whole `buildOwnerPaidUpdateInstruction({...})` call and differ only in `update`. If each branch builds just the payload and the builder is called once after the if/else, the shared proof arguments are spelled once and a future change to them touches one site. The built instruction is unchanged. `OwnerPaidUpdate` joins the existing import from `./update-info-owner` as `type OwnerPaidUpdate,` in the brace list.

Replacing lines 341-386:

```ts
        let update: OwnerPaidUpdate;
        if (input.deviceType === "iot") {
          update = {
            network: "iot",
            location: newLocation,
            elevation: toOnChainElevation(input.elevation),
            gain: toOnChainGain(input.gain),
          };
        } else {
          const mergedDeploymentInfo = mergeDeploymentInfo(
            info.mobile?.deploymentInfo ?? undefined,
            input.deploymentInfo
          );
          update = {
            network: "mobile",
            location: newLocation,
            deploymentInfo: mergedDeploymentInfo ?? null,
          };
          rentLamports = await estimateMobileResizeRent({
            connection,
            program: hemProgram,
            infoKey: hotspotInfoKey("mobile", entityKey),
            current: info.mobile,
            location: newLocation,
            deploymentInfo: withPreservedWifiSerial(
              mergedDeploymentInfo,
              info.mobile?.deploymentInfo ?? undefined
            ),
          });
        }

        const updateIx = await buildOwnerPaidUpdateInstruction({
          program: hemProgram,
          owner,
          entityKey,
          merkleTree: accounts.merkleTree,
          proofArgs: args,
          remainingAccounts,
          update,
        });
```

PR #1286, opened 2026-09-01
fix: size delegation claim bundle CU limits from static table

bryzettler:
Started as one fix for the delegation-claim bundle CU limit and grew into a pass over the governance bundle builders, the batch status/resubmission pipeline, and two swap error paths. Grouped below by area.

## Bundle CU limits (original problem)

Wallet-app users with multi-tx delegation claim backlogs hit `SIMULATION_FAILED` on `governance.claimDelegationRewards`: the Jito bundle died with `InstructionError [3, ProgramFailedToComplete]` on the last claim instruction of the last transaction.

Per-tx CU limits were estimated by simulating each transaction standalone against current chain state (sim × 1.1). The txs execute sequentially in a bundle, and `claim_rewards_v1` has a state-dependent branch: the claim that catches `last_claimed_epoch` up to `current − 1` runs a `clear_recent_proposals_v0` CPI (~9.6k CU). In a standalone sim of a later tx the earlier claims have not run, so the branch never fires and the sim measures the cheap path. In the bundle it fires and the tx runs out of compute. tx[0] is immune because standalone and bundle state agree for the first tx.

- `buildBatchedTransactions` sizes CU limits from the static instruction table (observed max × 2.0) by default, so the limit is state-independent. A bundle transaction whose instruction is missing from the table fails the build (`throwOnMiss`), naming the key; the simulation-failure fallback path still requests `MAX` and logs the key.
- `buildOrSplit` output is capped at the bundle size so a split can never produce a bundle of more than five transactions on mainnet.
- Measured and added table entries for the proxy and vote bundle instructions and the `nft_proxy` config instructions.
- Bundle simulation errors report the failing transaction's own logs and index (`failedTransactionIndex`) instead of a concatenation of every tx's logs.

## Batch status and resubmission

- Batches whose blockhash has expired are marked `expired` instead of being resubmitted until the retry cap while Jito answers "bundle contains an expired blockhash". A transaction's recorded `lastValidBlockHeight` comes from its own blockhash and is stored for every submitted row; submitting a client-crafted bundle the cluster no longer accepts returns `BLOCKHASH_EXPIRED`.
- Status decisions are made from one rule (`decideBatchStatus`) over the cluster's own view: batched `getSignatureStatuses`, block height, and per-blockhash validity, read before the DB transaction opens. The blockhash probe always runs at `confirmed` (at `finalized` it answers false for a merely young blockhash). Writes are compare-and-swap on the status the decision was made from; the batch is then rolled up from the rows as stored, and a terminal batch is never reopened to `pending` by a tick that learned nothing.
- A row whose probe went unanswered and that has no stored lifetime stays `pending` for the stale-batch reaper to bound, instead of being written `expired` on no evidence.
- When Jito cannot be read, the cluster's signature view still decides; only an RPC read failure skips the tick. This keeps a Jito outage from wedging every bundle batch and the `(tag, payer)` submit lock.
- Manual `transactions.resubmit` runs the status check first, like the background job, so a landed batch is persisted and reported as such. Resubmission no longer rewrites the stored `submissionType`.
- The stale-batch reaper reports a failed expiry instead of throwing.

## Governance builders

- `delegatePositions`: ownership, claim-bot, and epoch-info reads are batched; per-position instructions use `accountsStrict` with explicit account builders (`delegation-accounts.ts`, `claim-bot-accounts.ts`) so Anchor's resolvers never run. Tuktuk task ids are reserved from one bitmap read per bundle. Program IDLs are memoized (`anchor-idl-cache.ts`).
- Constant lockups are keyed off `i64::MAX` like the program, not `Number.MAX_SAFE_INTEGER`.
- Preflight rent is priced from program-declared account sizes (`rent.ts`) and covers delegated positions, epoch-info accounts, claim bots, queued tasks, crank rewards, and per-bot prepay. `createPosition` quotes every lamport the bundle charges; the automation prepay is quoted only when a claim bot is actually created (e2e asserts the wallet lands on exactly its rent floor).
- `delegatePositions` judges lockups, expirations and seasons against `registrar.clock_unix_timestamp()` (cluster clock plus `registrar.time_offset`), as the program does.
- Claim epoch-info chunks fetch concurrently; lookup table and blockhash are read once per batched build, with the blockhash read after packing.

## Swap

- `GetQuoteInput` rejects identical input and output mints as a 400 before reaching Jupiter. Jupiter's client-side codes (`CIRCULAR_ARBITRAGE_IS_DISABLED`, `TOKEN_NOT_TRADABLE`) map to `BAD_REQUEST`; rate limiting is classified before its error codes. The swap UI leaves the counterpart token out of each picker.

## Wallet history database load

`transactions.history` calls for a busy wallet ran 2-3 minutes each and lined up one-to-one with every "Monitoring RDS - High CPU Utilization" alarm (~98% CPU for 4 minutes, IOPS flat). `syncWalletHistory` ran one `PendingTransaction.findOne({ signature })` per Helius transaction, up to 1,000 per request, and `pending_transactions.signature` had no index, so each lookup was a sequential scan. The client times out at ~60s and retries, so three copies piled up per spike.

- Migration adds indexes on `pending_transactions(signature)` and `pending_transactions(batch_id)` (the latter backs the history handler's batch join), built `CONCURRENTLY` so the submit path keeps writing; an `INVALID` leftover from a failed build is dropped and rebuilt, and a failure fails the deploy instead of being swallowed.
- `syncWalletHistory` checks a whole page of signatures with one `IN` query instead of one query per transaction.

## Testing

Unit: 271 passing (adds compare-and-swap coverage for `checkAndUpdateBatchStatus`). E2e (surfpool): governance, delegate account resolution, and delegate automation suites added or extended.


packages/blockchain-api/src/lib/utils/batch-status-read.ts
```diff
@@ -0,0 +1,122 @@
...
+  try {
+    [currentBlockHeight, signatureStatuses, blockhashValidity] =
+      await Promise.all([
+        rpc.getBlockHeight({ commitment }),
+        fetchSignatureStatuses(rpc, batchId, open),
+        probeBlockhashValidity(
+          rpc,
+          open.flatMap((tx) => (tx.blockhash ? [tx.blockhash] : [])),
+          commitment,
+        ),
```
madninja, line 75: **Blocker: passing `commitment` into the blockhash probe makes `finalized` report landed transactions as failed.**

`commitment` is a required, client-supplied enum (`GetInputSchema`), and it flows `get.ts:44` -> `checkAndUpdateBatchStatus` -> here. `getBlockHeight` taking it is fine (a lower height errs safe), but `probeBlockhashValidity` is answering a different question: not "is this confirmed to the level I asked for" but "can this still land at all".

Measured on `api.mainnet-beta.solana.com`, blockhash fresh from `getLatestBlockhash({commitment: "confirmed"})`:

| probe | result |
| --- | --- |
| `isBlockhashValid @ processed` | `true` (slot 443770898) |
| `isBlockhashValid @ confirmed` | `true` (slot 443770899) |
| `isBlockhashValid @ finalized` | **`false`** (slot 443770867) |
| control: blockhash fetched *at* finalized, probed at finalized | `true` |

The control matters: the method is not broken at finalized, it returns `false` purely because the blockhash is younger than the finalized bank (~32 slots back). Reproduced twice.

Failure path: a wallet submits a 2-tx claim batch and polls `get(id, "finalized")` a second later. `getSignatureStatuses` returns `confirmationStatus: "confirmed"`, which is not in `LANDED_LEVELS.finalized`, so `decideTransactionStatus` falls through to `isTransactionExpired`; the probe says invalid; both rows are written `expired` and the batch `failed` with `confirmedAt` stamped. The client is told the claim failed, the transactions land seconds later, and neither the background job nor the reaper revisits a non-pending batch. A user retry is then a double claim.

Pre-PR this was self-limiting in the opposite direction, since expiry was `currentBlockHeight > lastValidBlockHeight` and `getBlockHeight({finalized})` returns a *lower* height, making finalized more conservative. The probe inverts that.

Suggest probing at a fixed `"confirmed"` (or `"processed"`) regardless of the caller's commitment, and leaving `getBlockHeight({commitment})` as is.

No in-tree caller passes `finalized` today (only `batch-status.test.ts:228`, with a mocked probe map, which is why the suite is green), so whether this is live depends on the wallet app's call. Worth confirming that before deciding urgency.

packages/blockchain-api/src/lib/utils/transaction-status-checker.ts
```diff
@@ -215,40 +165,23 @@ export async function checkAndUpdateBatchStatus(
...
-        batch.confirmedAt = new Date();
+      const confirmedAt =
+        isTerminal && !batch.confirmedAt ? new Date() : batch.confirmedAt;
+      const [updated] = await TransactionBatch.update(
+        { status: batchStatus, confirmedAt },
+        {
+          where: { id: batch.id, status: batch.status },
+          transaction: dbTransaction,
+        },
+      );
```
madninja: **Blocker: this write can move a terminal batch back to `pending`, and the status it writes is not derived from what the CAS above actually did.**

Two separate problems in this block.

**(a) Terminal to pending.** `decideBatchStatus` seeds `batchStatus` from `jitoBatchStatus` (default `"pending"`) and never from `batch.status`. For a zero-row batch (a leaked reservation the reaper already expired) `transactions.length === 0`, so the all-confirmed branch cannot fire, and `failedCount === 0`, so the failed branch cannot fire. The result is `"pending"`, `batchStatus !== batch.status` is true, and the CAS `where { id, status: 'expired' }` succeeds. The batch is back to `pending`, re-taking the `(tag, payer)` partial unique index, and because this write is not `silent` it resets `updated_at`, which is the reaper's clock. A client polling its own dead batch more often than every 2 minutes keeps a deterministic tag such as `claim_rewards:<wallet>` locked, and that wallet cannot submit.

Suggest refusing a non-terminal write over a terminal status, while keeping expired-to-confirmed healing:

```ts
if (batchStatus === "pending" && batch.status !== "pending") {
  // a terminal batch is not reopened by a tick that learned nothing
} else if (batchStatus !== batch.status) {
  ...
}
```

**(b) The rollup is pre-CAS.** `batchStatus`, `confirmedCount` and `failedCount` all come from the decision computed before the row loop. The row loop then CASes each row and `continue`s on a miss, so the batch status written here can disagree with the rows that actually landed. Two checkers straddling a state change: A decides all-confirmed, B marks row 2 failed first, A's row-2 CAS misses, and A still writes `confirmed` here and stamps `confirmedAt`. The result is a batch reported confirmed containing a failed row.

This is reachable rather than theoretical because there are now four callers: the 30s background job, the reaper, `transactions.get` (every client poll), and `transactions.resubmit`. Cheap fix: recompute the rollup from the post-CAS `pendingTx.status` values, which the loop already maintains.

packages/blockchain-api/src/lib/utils/transaction-status-checker.ts
```diff
@@ -115,98 +82,81 @@ export async function checkJitoBundleStatus(
...
+      // the status the decision was made from makes the write a
+      // compare-and-swap: a row that moved keeps whatever moved it, and the
+      // indexers are told about a landing exactly once.
+      const [updated] = await PendingTransaction.update(
+        {
+          status,
+          ...(status === "confirmed" ? { serializedTransaction: null } : {}),
+        },
+        {
+          where: { id: pendingTx.id, status: pendingTx.status },
```
madninja, line 143: Both CAS guards in this function are untested, and I confirmed it by mutation rather than by reading test names.

| mutation | result |
| --- | --- |
| drop `status: pendingTx.status` here | 267 passing, 0 failing (baseline exactly) |
| drop `status: batch.status` on the batch update | 267 passing, 0 failing (baseline exactly) |

The pure decision layer is genuinely well covered by contrast: five mutations in `batch-status.ts` (no-evidence default, empty-batch guard, settled-status short circuit, placeholder guard, probe-answer branch) all went red. The gap is precisely the consumption site, because no test file imports this module. Same for `transaction-resubmission.ts`, `jito.ts`, `wallet-history-sync.ts`, `transaction-submission.ts`.

Not untestable: `batch-build-shared-reads.test.ts` already establishes the pattern for a module that reads a validated env at load (type-only import, then dynamic import inside `before`), and a fake model recording its `where` clause pins both guards directly. Given the CAS is the concurrency claim this refactor rests on, it seems worth the one test file.

packages/blockchain-api/src/lib/utils/batch-status.ts
```diff
@@ -0,0 +1,199 @@
...
+  if (valid !== undefined) {
+    return !valid;
+  }
+
+  if (
+    transaction.lastValidBlockHeight == null ||
+    currentBlockHeight === undefined
+  ) {
+    return true;
+  }
```
madninja, line 107: On the known trade-off in the PR description: it is a bit worse than described, and there is a third option cheaper than either of the two you weighed.

Worse, because it does not need the submit-time probe to have failed *and* a later probe to fail. `probeBlockhashValidity` catches its own errors and simply omits the blockhash from the map, so a probe failure never reaches `readBatchStatus`'s tick-skip. It lands here as `valid === undefined`, falls through to a null `lastValidBlockHeight`, and returns `true`. The row goes `expired`, the batch rolls up to `failed` (expired is folded into `failedCount`), and the reaper then skips it because `batchStatus !== "pending"`. It also overrides a Jito `Landed` verdict, since `failedCount > 0` beats the `jitoBatchStatus: "confirmed"` seed.

The third option: return `false` here (stay pending) for the specific case of "blockhash present, probe unanswered, no stored lifetime", keeping `true` for the legacy row with no blockhash column at all. That does not risk the permanent wedge you were guarding against, because `reapStalePendingBatches` already owns exactly this case and bounds it at `STALE_PENDING_BATCH_MS` (2 minutes, comfortably past a blockhash lifetime). It already covers Jito batches and zero-row reservations, and the resubmission writes are `silent` so `updated_at` stays its clock. That makes the reaper the single owner of "we can never know" and removes the false-terminal risk.

Worth flagging that the current suite pins the existing default in three tests, so those are a decision to revisit alongside this rather than evidence for it.

packages/spl-utils/src/computeUnitTable.ts
```diff
@@ -411,11 +449,21 @@ const sumTableCu = (pairs: [string, Uint8Array][]): number => {
       INSTRUCTION_CU_TABLE[cuTableKey(programId, data)] ??
       PROGRAM_CU_CEILINGS[programId];
     if (cu === undefined) {
+      console.error(
+        `[computeUnitTable] No CU table entry for ${cuTableKey(
+          programId,
+          data
+        )}; requesting the ${MAX_COMPUTE_UNITS} CU maximum for this transaction`
+      );
       return MAX_COMPUTE_UNITS;
```
madninja: The PR description says a missing table entry is "a named error instead of a silent `MAX`". This is a named *log line*: it is `console.error` followed by `return MAX_COMPUTE_UNITS`, with no throw on any path, and `tests/priority-fees.ts` asserts exactly that behaviour.

That distinction has teeth now that `useTableComputeUnits` defaults to `true`, because as the comment right here says, under SIMD-0553 the requested limit is what gets billed. So a builder that gains an untabled instruction produces a standing 1.4M CU overcharge on every affected transaction, and the only signal is a console line in a server process (the comment concedes Sentry is not available here). No gate can fail on it.

Two related notes:

- `PROGRAM_CU_CEILINGS` short-circuits *before* this log, so a miss on System/Token/ATA/Memo is fully silent.
- The file header still describes the table as what is "used when transaction simulation fails", and `FALLBACK_CU_MARGIN`'s comment still says it is "only hit when live simulation is unavailable". For every `buildBatchedTransactions` caller the table is now the primary path, so both read as the opposite of what happens.

Suggest throwing for the `useTableComputeUnits: true` callers specifically, and rewriting those two comments. I checked that a throw propagates cleanly rather than being swallowed: `buildOrSplit`'s catch filters on `encoding overruns` and `too large to serialize`, so a missing-entry error would not trigger the recursive split.

For what it is worth, I verified the table itself is complete for this diff: all 24 instructions the changed builders emit have entries, and all 16 new or changed discriminators recompute correctly from `sha256("global:<name>")`.

packages/blockchain-api/src/lib/utils/jupiter-errors.ts
```diff
@@ -0,0 +1,70 @@
...
+  // us spamming Sentry with JUPITER_ERROR 500s. Decided before the error codes
+  // because a throttled or failing response can echo the request's own code
+  // back to us, which does not make it the caller's fault.
+  if (status === 429) {
+    return { kind: "RATE_LIMITED" };
+  }
+
+  const errorCode = parseErrorCode(body);
+  const clientMessage =
+    errorCode === undefined ? undefined : CLIENT_ERROR_MESSAGES[errorCode];
```
madninja: **Blocker: this index reaches `Object.prototype`, so an unknown code fails open into a 400.**

`parseErrorCode` correctly narrows to `typeof code === "string"`, but `"constructor"` is a string. Measured against this table:

| `errorCode` in Jupiter's body | lookup returns | passes `!== undefined` | verdict |
| --- | --- | --- | --- |
| `constructor` / `toString` / `valueOf` / `hasOwnProperty` | a function | yes | **BAD_REQUEST (400)** |
| `__proto__` | `Object.prototype` | yes | **BAD_REQUEST (400)** |
| `TOKEN_NOT_TRADABLE` | the message | yes | BAD_REQUEST (400), correct |
| `NOPE` | `undefined` | no | JUPITER_ERROR (500), correct |

So a 500 from Jupiter, a proxy or a WAF carrying `{"errorCode":"constructor"}` tells the client its request is bad and not to retry, and simultaneously suppresses the Sentry report, since `route.ts` captures only `status >= 500`. That is the exact inversion of what this change is for. `errors.BAD_REQUEST({ message })` also receives a function rather than a string.

```suggestion
  const clientMessage =
    errorCode !== undefined && Object.hasOwn(CLIENT_ERROR_MESSAGES, errorCode)
      ? CLIENT_ERROR_MESSAGES[errorCode]
      : undefined;
```

I confirmed this restores the 500 for `constructor` while keeping the 400 for the real codes. Worth a test with `errorCode: "constructor"` asserting `JUPITER_ERROR`, since nothing but a prototype-named key will catch a regression here.

packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
```diff
@@ -0,0 +1,36 @@
...
+        name: "idx_pending_transactions_signature",
+      })
+      .catch(() => {});
+
+    await queryInterface
+      .addIndex("pending_transactions", ["batch_id"], {
+        name: "idx_pending_transactions_batch_id",
+      })
+      .catch(() => {});
+  },
```
madninja, line 44: **Blocker: a plain `CREATE INDEX` blocks writes on `pending_transactions`, and `.catch(() => {})` hides its own failure.**

Measured SQL from the installed sequelize 6.37.7:

```
as written : CREATE INDEX "idx_pending_transactions_signature" ON "pending_transactions" ("signature")
concurrent : CREATE INDEX CONCURRENTLY "idx_pending_transactions_signature" ON "pending_transactions" ("signature")
```

Without `CONCURRENTLY` Postgres takes a SHARE lock: reads are fine, every INSERT/UPDATE/DELETE blocks for the build duration. `docker-entrypoint.sh:34` runs `db:migrate` at container start while the old replicas keep serving, so `transactions.submit` and the status-checker and resubmission writes stall through two index builds, on a table this migration's own docblock describes as large enough to pin the RDS CPU for minutes.

The swallowed error is the half I would fix even if the lock were acceptable: if creation fails on a lock or statement timeout, nothing is logged, `SequelizeMeta` records the migration as applied, the sequential scans continue in production with no signal, and it never retries. The entrypoint has `set -e`, so letting a real error propagate correctly fails the deploy.

```suggestion
  async up(queryInterface) {
    // CONCURRENTLY keeps the index build from holding a write-blocking lock on
    // a table the API writes to on every submit.
    await queryInterface.addIndex("pending_transactions", ["signature"], {
      name: "idx_pending_transactions_signature",
      concurrently: true,
    });

    await queryInterface.addIndex("pending_transactions", ["batch_id"], {
      name: "idx_pending_transactions_batch_id",
      concurrently: true,
    });
  },
```

`concurrently: true` is legal here: sequelize-cli 6.6.5 has no `transaction` reference anywhere in its `lib/`, so migrations are not transaction-wrapped, and `CREATE INDEX CONCURRENTLY` cannot run inside one. One caveat to handle: a failed concurrent build leaves an INVALID index behind, so it is worth a `pg_index.indisvalid` check on re-run rather than relying on `IF NOT EXISTS`, which would skip an invalid index forever.

Separately, `down()` is unguarded while `up()` is, which is the reverse of sibling migration `20260317000000`. `DROP INDEX IF EXISTS` would make the pair symmetric.

Also worth confirming the index shape is what the join needs: I checked the history handler generates `LEFT OUTER JOIN "pending_transactions" AS "transactions" ON "transaction_batches"."id" = "transactions"."batch_id"`, a single-column predicate, so the single-column `(batch_id)` index is exactly right. And `signature` correctly must not be unique, given the `${batchId}-${i}` placeholders and the in-place signature rewrite on resubmit.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/delegation-accounts.ts
```diff
@@ -0,0 +1,217 @@
...
+  proxyConfig: PublicKey;
+  /** `registrar.dao`, as carried by `sub_dao.dao`. */
+  dao: PublicKey;
+  delegatedPosition: PublicKey;
+  subDao: PublicKey;
+  /**
+   * The cluster clock's `unix_timestamp`. The program reads
+   * `registrar.clock_unix_timestamp()`, which adds `registrar.time_offset` —
+   * a test-only dial that is zero on mainnet, so the two agree there.
+   */
```
madninja, line 40: This docstring is accurate today, but it states a mutable on-chain value as though it were an invariant, and the cost of honouring it is one addition.

The Rust is `Clock::get().unwrap().unix_timestamp.checked_add(self.time_offset).unwrap()` (`registrar.rs:30-36`), and the SDK resolvers this replaces do add it (`helium-sub-daos-sdk/src/resolvers.ts:50,98,130,275`). I read the mainnet HNT registrar `BMnWRWZrWqb6JMKznaDqNxWaWAHoaTzVabM6Qwyh3WKz` directly: `time_offset = 0`, so mainnet does agree.

What it costs: `set_time_offset_v0` is a real instruction with 29 samples in the CU table, so it is exercised. Any localnet or e2e run that sets an offset derives the wrong `sub_dao_epoch_info` and fails `ConstraintSeeds`. This PR adds two e2e suites in exactly that area, which is where I would expect it to bite first.

The registrar is already fetched here, so `now.add(info.registrar.timeOffset)` at the `now` derivation (`delegate.ts:103`) closes it and lets the docstring drop the caveat entirely.

packages/blockchain-api/src/server/api/routers/governance/procedures/positions/create.ts
```diff
@@ -348,38 +366,54 @@ export const create = publicProcedure.governance.createPosition.handler(
...
       delegatedPositionRent +
-      claimBotRent +
-      (subDaoMint && automationEnabled ? RENT_COSTS.TUKTUK_TASK : 0); // delegation claim task
+      epochInfoRent +
+      automationRent +
+      claimTaskRent +
+      queuedTaskCrankReward;
 
     const estimatedSolFeeLamports =
       (await getTotalTransactionFees(connection, versionedTransactions)) +
```
madninja, line 420: `estimatedSolFeeLamports` on line 424 adds `PREPAID_TX_FEES` on `automationEnabled` alone:

```ts
(automationEnabled ? PREPAID_TX_FEES * LAMPORTS_PER_SOL : 0);
```

but the transfer that actually spends it (line 273) is only emitted under `subDaoMint && automationEnabled`, and this same block already computes `automates` for every other term. So `automationEnabled: true` with no `subDaoMint` over-quotes 0.01 SOL and can reject a wallet that has enough. Using `automates` here lines it up with the rest.

Direction matters for the rent work generally: an under-quote fails the bundle after the user has approved it, an over-quote blocks a legitimate action. The e2e funds exactly `required` and submits, which catches under-quotes well, but an over-quote still passes. Asserting the post-submit balance lands near `MIN_WALLET_RENT_LAMPORTS` would close that direction, and would also have caught this.

(Anchored a line up because 424 is not in the diff.)

The arithmetic itself checks out, for what it is worth: I reconciled it against a real mainnet `initialize_position_v0` and it is exact to the lamport (mint 1,461,600 + position 2,589,120 + metadata 5,115,600 + 10,000,000 create fee + 2 ATAs at 2,039,280 = 23,244,880 = payer delta minus tx fee). And "no longer double-counts the Jito tip tx fee" seems to correspond to no actual change: `jitoTipCost` is value-identical to the base commit and is the transfer amount, while `getTotalTransactionFees` sums signature fees, so there was no double count to remove.

packages/blockchain-api/src/server/api/routers/governance/procedures/positions/create.ts
```diff
@@ -55,6 +55,14 @@ import {
   toLockupKindArg,
   LockupKindType,
   buildBatchedTransactions,
+  getAutomationRentLamports,
+  getMissingEpochInfoRentLamports,
+  DELEGATED_POSITION_SPACE,
+  DELEGATION_CLAIM_TASK_SPACE,
+  POSITION_SPACE,
+  SUB_DAO_EPOCH_INFO_SPACE,
```
madninja: Imported but never used in this file (it is consumed inside `getMissingEpochInfoRentLamports`). `tsconfig` has no `noUnusedLocals`, so nothing flags it.

```suggestion
```

packages/blockchain-api/src/lib/anchor-idl-cache.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * An IDL only changes when the program is upgraded, so it is read once per
+ * process rather than once per request. Anchor's account resolvers call the
+ * SDKs' `init` on every unresolved account, which without this is one IDL read
+ * per resolver call.
+ *
+ * The tradeoff: a program upgrade is not picked up until the process restarts.
+ */
+export async function fetchCachedIdl(
```
madninja: `fetchCachedIdl` has no consumer outside this file and its own test, which imports only `initCachedProgram`. Suggest keeping it module-private until something needs it.

```suggestion
async function fetchCachedIdl(
```

Noting for the record that the process-lifetime cache tradeoff is already documented two lines up, so I am not raising that as a finding. The one thing worth checking is whether the deploy runbook reflects it, since a program upgrade now silently requires a restart before `accountsStrict` sees a new account.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/build-batched-transactions.ts
```diff
@@ -260,11 +283,24 @@ export async function buildBatchedTransactions({
...
   }
 
+  // The same blockhash goes into every transaction, and it starts ageing the
+  // moment it is read, so it is read after packing rather than before: the
+  // signer gets the full validity window instead of what packing left of it.
+  const { blockhash } = await connection.getLatestBlockhash("finalized");
+
   const built = (
     await Promise.all(
       packedBatches.map(({ instructions, metadata, signers }) =>
```
madninja, line 293: `packedBatches` is capped at `effectiveMaxTxs` (4 on mainnet) at lines 262 and 280-281, but `buildOrSplit` can turn one packed batch into two or more when the oversize path fires, and `built.length` is never re-checked against the cap.

`shouldUseJitoBundle` returns true for any length above 1 with no upper bound, so the tip is appended regardless and the bundle goes out at 6 transactions against Jito's limit of 5. In practice `submit` rejects it first with "Maximum of 5 transactions allowed per batch", so the user gets a confusing 400 and `hasMore` does not tell the client to paginate, because `stoppedEarly` is computed during packing and knows nothing about the split.

Second-order, since the `measureSize` pre-check exists to keep the split from firing at all, but a bounds check after `built` plus setting `hasMore` would make it safe by construction.

packages/blockchain-api/src/lib/utils/transaction-resubmission.ts
```diff
@@ -200,6 +203,73 @@ export async function resubmitTransactionBatch(
...
+  // The cluster's own view of the signatures decides this, not the blockhash
+  // probe alone: a batch that landed in its last valid block carries a dead
+  // blockhash and a confirmed signature, and expiring it on the blockhash would
+  // record a landed batch as expired for good.
+  const connection = new Connection(env.SOLANA_RPC_URL);
+  const decision = await readBatchStatus({
+    rpc: connection,
+    batchId: batch.id,
+    transactions: pendingTransactions,
+  });
```
madninja: This is the second full cluster read of the same batch in one tick: every caller runs `checkAndUpdateBatchStatus` first (which already called `readBatchStatus` at `transaction-status-checker.ts:100`) and returns early unless the decision says every row is still pending. So this re-reads block height, signature statuses and N blockhash probes for up to 5 batches concurrently, every 2s, per replica.

Beyond the RPC cost, it is the only thing that can make the two reads disagree, which is what makes the expiry block below reachable at all. `BatchStatusReadResult` already carries `currentBlockHeight`, so threading the caller's decision in here would remove both the cost and the divergence.

On that expiry block specifically (lines 239-271): `expiredCount > 0` expires *all* pending rows of the batch, so for a parallel or sequential batch whose transactions carry different blockhashes, one sibling's death terminates rows that can still land. And it writes batch status `expired` where `decideBatchStatus` maps the same fact to `failed`, so the PR description's "marked `expired`" does not hold in any path a caller actually takes. Giving `decideBatchStatus` an `expiredCount` and rolling `expiredCount > 0 && confirmedCount === 0` up to `expired` would leave one rule, one label and one read.

madninja (changes requested):
Reviewed the whole diff, then fanned out three domain reviews (batch state machine, governance builders/CU/rent, swap/history/migration) and verified their load-bearing claims myself against live mainnet RPC and a mutation harness on the PR head.

This is careful work. The `check-cu-table` gate fix, the read-before-transaction ordering, and the account-by-account mirroring of the Rust structs are all right. I measured the core CU premise and it holds with margin: at most 8 `claim_rewards_v1` fit one tx against the live LUT, requesting 1,108,272 CU against a real worst case around 442,200, and the Rust confirms the diagnosis (the `clear_recent_proposals_v0` CPI is gated on `current_epoch == last_claimed_epoch + 1`, which an earlier bundle tx advances, so a standalone sim cannot see it).

Requesting changes for four items. Worth saying up front why green CI is not reassuring here: the two most serious defects are both in paths CI never executes.

## Blockers

**1. `transactions.get({commitment: "finalized"})` marks live and already-landed transactions `expired` and the batch `failed`, immediately.** The client's requested confirmation level is also used as the "can this still land" probe level. Measured on mainnet: a blockhash fresh from `getLatestBlockhash({confirmed})` probes `true` at processed and confirmed, `false` at finalized. Control, to rule out a broken method: a blockhash fetched at finalized probes `true` at finalized. So it is age, the finalized bank is about 32 slots back. Details inline on `batch-status-read.ts`.

**2. A terminal batch can be dragged back to `pending`, which permanently holds a wallet's submit tag.** `decideBatchStatus` seeds from `jitoBatchStatus` (default `pending`) and never from `batch.status`, so a zero-row reservation the reaper already expired decides `pending` again and the CAS succeeds over `expired`. The write is not silent, so it also resets the reaper's clock. Inline on `transaction-status-checker.ts`.

**3. Jupiter's error-code lookup fails open into a 400 and silences Sentry.** `CLIENT_ERROR_MESSAGES[errorCode]` is a plain-object index on an upstream-supplied key. Measured: `constructor`, `toString`, `valueOf` and `hasOwnProperty` all return a function and `__proto__` returns `Object.prototype`, every one of which passes `!== undefined`. Committable suggestion inline.

**4. The migration takes a write-blocking lock on the table the incident was about, and swallows its own failure.** Measured SQL from sequelize 6.37.7: `CREATE INDEX "idx_..." ON "pending_transactions" ("signature")`, no `CONCURRENTLY`, so a SHARE lock blocks every INSERT/UPDATE/DELETE for the build. It runs at container start while old replicas are still serving. Separately, `.catch(() => {})` on both indexes means a failure is unlogged while `SequelizeMeta` records success, so the sequential scans this PR exists to remove would continue with no signal and no retry. Committable suggestion inline. `concurrently: true` is legal here because sequelize-cli 6.6.5 does not wrap migrations in a transaction (no `transaction` reference anywhere in its `lib/`).

## Not anchorable to a changed line

**The two new e2e suites never run in CI.** `delegate-account-resolution.test.ts` and `delegate-automation.test.ts` are 353 lines and neither appears in the hand-maintained 17-entry matrix in `.github/workflows/blockchain-api-e2e.yml`. CI runs `test:e2e:file` once per matrix entry and never the `test:e2e` glob, so adding a file is not enough. `data-credits.test.ts` and `transaction-classifier.test.ts` are already orphaned the same way, so this is a class rather than a one-off. Suggest adding both entries plus a check that every `tests/e2e/*.test.ts` appears in the matrix.

**`getCurrentSeason` is missing the upper bound, so an existing refusal cannot fire.** `helpers/get-current-season.ts:12` matches `now.gte(season.start)` only; `ProxyConfigV0::get_current_season` requires `start <= t && end > t` and returns `None`, which `delegate_v0` unwraps. Past the last season's end the TS returns the stale final season, so `if (!seasonEnd) throw BAD_REQUEST` can never fire and the user signs a bundle that panics on chain. Pre-existing and far off (mainnet's last season ends 2029-08-01), but this PR makes it the sole source of the rule, so it is cheap to fix while the area is open.

**`resubmitSingleTransaction` has zero references repo-wide** and carries a second copy of the expiry rule that no longer matches the new one (stored bound only, no probe). Deleting it would make the "one rule" claim true.

## Coverage

The split here is stark, and it is the classic shape where a tested helper says nothing about the feature until its caller is mutated too.

| Layer | Result |
| --- | --- |
| Pure decision functions (`batch-status.ts`, Jupiter classifier, `delegation-accounts`, `computeUnitTable`) | Well tested. 13 of 15 mutations caught, including the 429-ordering case and the `build-transaction.ts` call site |
| Consumption sites in the DB and network modules (row CAS, batch CAS, reaper, expiry path, Jupiter throw block) | No coverage. Every mutation NOT CAUGHT |

Concretely: dropping `status: pendingTx.status` from the row update and `status: batch.status` from the batch update each leave 267 of 267 passing with zero failures. `git grep` over `tests/` finds no reference to `checkAndUpdateBatchStatus`, `resubmitTransactionBatch`, `reapStalePendingBatches` or `checkJitoBundleStatus`. Not untestable: `batch-build-shared-reads.test.ts` already shows the idiom for these modules (type-only import, then dynamic import after env setup), and a fake model that records its `where` clause covers both CAS guards.

Baseline was re-derived immediately before each run (267/0 unit, 28/0 priority-fees), every CAUGHT verdict kept pass+fail at the baseline with at least one failure and zero pending, and every mutation was restored from a copy taken beforehand with the tree confirmed byte-identical.

## Smaller things, inline

`err` is accepted at any commitment while `failed` is permanent; the row read-modify-write at `transaction-resubmission.ts:319` sits outside the CAS and can double-notify the indexers; the submit-time blockhash probe is gated on a client-controlled tag, so the client picks whether the server validates; `buildOrSplit` runs after the 4-tx cap so a split can produce 6 txs past Jito's 5; the mint-pair refine is on `GetQuoteInputSchema` only and does not reach `getInstructions`; `vote_v0`'s effective headroom is 1.14x rather than 2x (entry 199,539 against its own stated max of 350,893) and its cost driver, proposal choice count, is not bounded by the 1,007-vote sample; `genesisEndTs.lte(now)` and the over-quote direction of the rent check are both untested (the first is a NOT CAUGHT mutation); no changeset covers the headline CU-sizing change.

## Verified clean, so it is distinguishable from not-looked-at

All 16 new and changed CU discriminators recomputed from `sha256("global:<name>")` and matched exactly, and the `nft_proxy` program id matches both the IDL's own `address` and existing repo references. All 24 instructions the changed builders emit have table entries, no gap. Account lists check out against the programs: DelegateV0 15/15, CloseDelegationV0 14/14, ChangeDelegationV0 19/19, ClaimRewardsV1 19/19, with seeds and both closing-epoch rules matching. Rent arithmetic reconciles to the lamport against a real mainnet `initialize_position_v0`. `i64::MAX` constant lockups are exact with no 2^53 crossing. Task-id reservation cannot double-assign within a bundle. `getSignatureStatuses`' 256-signature cap is unreachable given the 5-tx batch limit. The `(tag, payer)` index is on `transaction_batches`, so a `failed` batch does release the lock. `signature` correctly must not be unique. The history `IN` query is escaped, bounded, and keyed by a `Set`, so the prototype hazard does not apply there. `payer` is derived from the signature via `verifiedFeePayer` rather than the account list. Nothing moved off localnet, and the `loadBuiltIxNames()` split is the right fix for the trigger/scope gap the committed external IDLs would otherwise have opened.

packages/blockchain-api/src/lib/utils/batch-status-read.ts
bryzettler: Fixed in 0319bc85a. The probe is now fixed at `confirmed` regardless of the caller's commitment, with a comment saying why (`isBlockhashValid` at `finalized` answers false for a blockhash that is merely young, and a false probe expires the row). `readBatchStatus` still reads block height and signature statuses at the caller's commitment. Unit test added: "probes blockhash validity at confirmed even when the caller asked for finalized".

packages/blockchain-api/src/lib/utils/transaction-status-checker.ts
bryzettler: Both points fixed in 0319bc85a.

1. Terminal → pending: the batch write is skipped when the rollup is `pending` and the stored status is not, so a reaped zero-row batch (or one whose rows are still pending) can no longer be pulled back to `pending` by a client poll. Covered by two new tests ("does not reopen a terminal batch from a tick that learned nothing" / "...whose rows are still pending").

2. Rollup vs rows: the batch status is now computed after the row CAS loop, from the rows as stored. A row whose write lost the CAS is reloaded inside the DB transaction, and `rollupBatchStatus` (extracted from `decideBatchStatus`) runs over the resulting statuses. The returned `transactionStatuses`/counts come from the same post-CAS rows. Test: "rolls the batch up from the rows as stored when a row's write lost the race" (cluster says confirmed, another writer failed the row first → batch written `partial`, not `confirmed`).

packages/blockchain-api/src/lib/utils/transaction-status-checker.ts
bryzettler: Added `tests/unit/transaction-status-checker.test.ts` in 0319bc85a. It drives `checkAndUpdateBatchStatus` with stubbed `Connection` reads, `PendingTransaction.findAll/update`, `TransactionBatch.update` and `sequelize.transaction`, and asserts the `where` clauses of both writes (`{ id, status: <status decided from> }`), that a lost row CAS reloads and changes the batch rollup, that a lost batch CAS leaves the in-memory batch untouched, and that a terminal batch is not reopened. Dropping either `status:` guard from a `where` now fails the first test.

packages/blockchain-api/src/lib/utils/batch-status.ts
bryzettler: Agreed and changed in 0319bc85a. When the probe went unanswered and there is no stored lifetime, `isTransactionExpired` now returns `false` (only a row with no blockhash at all, which predates the column, still counts as expired). The row stays pending and the stale-batch reaper bounds it. Since the earlier commit also stores `lastValidBlockHeight` for every submitted row, this case is now limited to legacy rows anyway. The "Known trade-off" section in the PR body is gone; the two unit tests that asserted the old behaviour were rewritten to assert the new one.

packages/spl-utils/src/computeUnitTable.ts
bryzettler: Done in 0319bc85a.

- `sumTableCu`/`tableComputeUnits`/`tableComputeUnitsForInstructions` take `{ throwOnMiss }`. `buildVersionedTransaction` computes the table limit with `throwOnMiss: true` *before* its try/catch when `useTableComputeUnits` is set, so a miss on a bundle transaction fails the build with the key in the message instead of being swallowed into the fallback. The simulation-failure fallback keeps the MAX + console.error behaviour, since there a landed-but-overpriced tx beats no tx.
- Header comment and the `FALLBACK_CU_MARGIN` comment rewritten: the table is the primary sizing for bundle producers and the fallback elsewhere.
- Tests: `tests/priority-fees.ts` "throws on a miss when the caller chose the table over simulation"; `batch-build-shared-reads.test.ts` "refuses to build a bundle transaction the table cannot price".
- PR body updated to say "fails the build" rather than "named error".

On the last point: yes, an under-request from a stale `PROGRAM_CU_CEILINGS` entry or drifted table row is still silent at build time and only shows up on-chain. The throw cannot catch that; `check-cu-table` in CI (drift beyond `FALLBACK_CU_MARGIN`) is the guard for tabled instructions, and the per-program ceilings are sized well above the measured max of the instructions we send through them.

packages/blockchain-api/src/lib/utils/jupiter-errors.ts
bryzettler: Fixed in 0319bc85a exactly as suggested (`Object.hasOwn` guard). Test added: `{"errorCode":"constructor"}` with a 500 now classifies as `JUPITER_ERROR` carrying the raw body.

packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
bryzettler: Rewritten in 0319bc85a:

- Both indexes are created with `concurrently: true` (Sequelize's postgres query generator emits `CREATE INDEX CONCURRENTLY`; sequelize-cli does not wrap migrations in a transaction, which that statement could not run inside anyway).
- No `.catch`: a failure fails `db:migrate` and, via `set -e` in the entrypoint, the deploy.
- Before each create, `pg_index.indisvalid` is checked by name: a valid existing index is kept (idempotent re-run), an `INVALID` leftover from a failed concurrent build is dropped and rebuilt.
- `down` uses `DROP INDEX IF EXISTS`.

I dry-ran `up` against a fake query interface to confirm the statement sequence; there is no postgres in the unit suite.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/delegation-accounts.ts
bryzettler: Applied in 0319bc85a. `delegate.ts` now fetches the registrar before the lockup check and computes `positionNow = clock + registrar.timeOffset` per position; that value feeds the decayed-lockup check, `positionAccountArgs.now`, the expired-delegation filter and `getCurrentSeasonEnd`, matching `create.ts` and the program's `clock_unix_timestamp()`. The `now` docstring in `delegation-accounts.ts` no longer carries the caveat.

packages/blockchain-api/src/server/api/routers/governance/procedures/positions/create.ts
bryzettler: Correct, and fixed in 0319bc85a: the prepay is now gated on `automates` like the transfer it prices. The e2e "quotes exactly the SOL an automated position spends" now also asserts that after the bundle lands the wallet holds exactly `MIN_WALLET_RENT_LAMPORTS`, i.e. the quote and the spend agree to the lamport. The PR body's "no longer double-counts the Jito tip tx fee" line was wrong and has been removed.

packages/blockchain-api/src/server/api/routers/governance/procedures/positions/create.ts
bryzettler: Removed in 0691c572c (the `SUB_DAO_EPOCH_INFO_SPACE` import is gone from `create.ts`).

packages/blockchain-api/src/lib/anchor-idl-cache.ts
bryzettler: `fetchCachedIdl` is module-private as of 0691c572c; only `initCachedProgram` is exported. On the restart-after-upgrade point: it is documented in the module docstring, and there is no deploy runbook in `docs/` to add it to.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/build-batched-transactions.ts
bryzettler: Fixed in 0319bc85a. After `buildOrSplit` results are flattened, `built` is truncated to `effectiveMaxTxs` and `stoppedEarly` is set, so a mainnet bundle is at most `maxTxs` including the tip and the remainder is reported through `hasMore`. The cap applied while packing is kept as the early exit; this is the post-split re-application of the same bound.

packages/blockchain-api/src/lib/utils/transaction-resubmission.ts
bryzettler: Fixed in 0691c572c. `resubmitTransactionBatch` no longer does a second cluster read: `checkAndUpdateBatchStatus` is the only read, and blockhash expiry flows through `decideBatchStatus`, which writes the rows `expired` and rolls the batch up to `failed`. The separate `expired` result flag and the label mismatch are gone; the changeset was updated to match.

madninja (approved):
All four blockers are fixed, and I verified each by mutating the guard the fix introduced rather than by reading the diff, since the previous round's own fixes were the likeliest place for a new hole. Baseline re-derived on `04667298b` immediately beforehand: 271 passing (was 267), typecheck and lint clean.

| Mutation of the new guard | Verdict |
| --- | --- |
| Un-pin the probe back to the caller's commitment | CAUGHT (270+1) |
| Remove the terminal-to-pending skip | CAUGHT (269+2) |
| Remove the lost-CAS row reload | CAUGHT (270+1) |
| Revert the no-evidence default to `expired` | CAUGHT (269+2) |
| `throwOnMiss: true` to `false` | CAUGHT (270+1) |
| Bypass the `Object.hasOwn` guard | CAUGHT (270+1) |
| Disable the post-split bundle cap | NOT CAUGHT (271+0) |
| Drop `registrar.timeOffset` again | NOT CAUGHT (271+0) |

Every CAUGHT verdict kept pass+fail at the 271 baseline with at least one failure and zero pending, so none is a broken-setup pass, and the tree was byte-identical after each.

Two of these are worth calling out as done well rather than just done.

The `throwOnMiss` fix avoided the trap it could easily have created. A new option defaulting to `false` is exactly the shape where a guard becomes decorative, and the thing that makes it real is that `tableComputeUnitLimit` is computed at `build-transaction.ts:88-92`, outside and before the `try`. Inside it, the catch at line 121 would have re-priced a miss at the 1.4M maximum and carried on, which is the behaviour the change set out to remove. The mutation confirms the placement is load-bearing.

The expiry rule reading `return !transaction.blockhash` is the right shape too: the row stays open when it has a blockhash and only a legacy row with no blockhash counts as expired, with the docstring naming the reaper as the mechanism that bounds it instead of asserting a guarantee.

The migration rewrite is solid: `concurrently` on both indexes, no swallowed error, `pg_index.indisvalid` checked per index with an invalid leftover dropped and rebuilt, and `DROP INDEX IF EXISTS` in `down`. Accepting that it has no automated coverage, which seems right for one-shot DDL.

## Follow-ups, none blocking

I applied a "is this worse than `develop`?" test to everything left, and nothing is. Recording them so they are not lost rather than to hold the PR up.

Not addressed, and all three happen to be the findings that had no inline anchor, so I read them as missed rather than declined:

- The e2e matrix is unchanged at 17 entries against 21 test files, so `delegate-account-resolution.test.ts` and `delegate-automation.test.ts` still do not run in CI (`data-credits` and `transaction-classifier` were already in that state). Two lines of YAML, and it is the reason the `timeOffset` fix below has nowhere to be verified, so it is the one I would still suggest doing here.
- `getCurrentSeason` still matches `now.gte(season.start)` only, so `if (!seasonEnd) throw BAD_REQUEST` remains unreachable and a caller past the last season's end signs a bundle that panics in `delegate_v0`. Pre-existing and unchanged in behaviour; mainnet's last season ends 2029-08-01.
- `resubmitSingleTransaction` still has zero references and still carries a second copy of the expiry rule, which this round diverged further.

Newly fixed but unverified, both strictly better than before (each was previously a bug *and* untested):

- The post-split bundle cap at `build-batched-transactions.ts:310` is untested but testable: `batch-build-shared-reads.test.ts` already stubs a connection and inspects `versionedTransactions`, and none of its four tests drives the oversize split or asserts `hasMore`.
- `registrar.timeOffset` has no test anywhere under `tests/` (zero grep hits), and its natural home is one of the two e2e files CI does not run.

One nit in the migration: the `pg_class.relname = :name` lookup is not schema-qualified, so it would match a same-named index in another schema and then read `rows[0]` arbitrarily.

## Two repo-wide gaps this PR surfaced

Neither belongs to this change, but both will keep costing time.

The e2e matrix is a hand-maintained enumeration whose scope (every `tests/e2e/*.test.ts`) is wider than its trigger (17 listed files), and a missing entry renders as nothing at all, so no PR page can show it. A check that every test file appears in the matrix would mechanize the class.

Separately, the `voter-stake-registry` red X on this PR was not the test: it passed with 22 passing, matching this branch's previous run and an unrelated branch, and `Check CU table covers test traffic` passed too, reporting `CU table OK: 22 sampled instruction(s) covered within margin` over 29 sampled `nft_proxy` signatures, which is CI affirmatively validating the new table entries. The job failed in `Post Run ./.github/actions/build-anchor/`, where `hashFiles('**/Cargo.lock')` (`build-anchor/action.yaml:29`) could not read the workspace during post-job cleanup. `Cargo.lock` is tracked, so this is the cache-save racing teardown. It has now failed twice in the last twelve runs of that workflow, and a post-action that fails after a green test trains people to re-run without reading. I re-ran the failed job and it is green, with `Cache Cargo registry + index` succeeding this time.

CI is now 67 passing with 3 skipped by design. Leaving the merge decision to you.

PR #1288, opened 2026-09-01
fanout: attribute vault inflow at the share count that was live

madninja:
Makes the fanout program's share accounting say what it means, and writes the rules down.

## Accounting

`fanout.total_inflow` is the accumulator every voucher watermark is measured against, so an arrival has to be folded into it at the staked-share count in force when it is folded. That fold now lives in `FanoutV0::accrue_inflow` and both callers use it:

- `stake_v0` folds before the new shares join the staked set, and seeds the new voucher's watermark from the accumulator rather than the token account balance. Joining therefore does not dilute the vouchers already staked.
- `distribute_v0` folds through the same helper and drops the snapshot by what it transferred.

The fold stays callable at any share ratio, since `stake_v0` folds before it raises `total_staked_shares` and that is the only way the ratio grows.

`unstake_v0` removes `voucher.shares` from `total_staked_shares`, which is the sum of the live vouchers' shares. The stake account's balance is a separate quantity.

## Bounds

- `stake_v0` refuses a stake that would carry the staked set past `total_shares`, which is fixed at the membership supply the fanout was created with.
- `distribute_v0` requires its recipient account to be something other than the fanout's own token account, so the snapshot it writes reflects a transfer that moved tokens.

Both append errors to the IDL, at 6001 and 6002; 6000 is unchanged. `@helium/idls` changeset included.

## README

The README described a pro-rata split across holders. Revenue is split across *staked* shares: each arrival is scaled by `total_shares / total_staked_shares`, so the staked vouchers take all of it and unstaked shares take none. That is now written down, along with what an unstake forfeits and what it strands, the fixed `total_shares`, and `fanout.authority` being stored at creation and never read again.

## Verification

`cargo test -p fanout --lib` and `tests/fanout.ts` on localnet, plus clippy with the repo's flags, `cargo fmt --check`, and `check-cu-table`.

### DIFF
--- a/programs/fanout/README.md
+++ b/programs/fanout/README.md
@@ -1,8 +1,25 @@
 # fanout
 
-Splits revenue that lands in a shared token account across holders of pro-rata membership NFTs. Holders can stake HNT to mint voucher NFTs, collect distributions against their voucher, and burn the voucher to withdraw their stake.
+Splits revenue that lands in a shared token account across staked membership positions. Holders stake membership tokens to mint a voucher NFT, collect distributions against that voucher, and burn the voucher to withdraw their stake.
 
 Used for multi-party revenue splits where each participant's share is transferable. For small, fixed splits prefer the cheaper [mini-fanout](../mini-fanout).
+
+## How a distribution is sized
+
+`fanout.total_inflow` is a lifetime accumulator scaled to the full share base, `fanout.last_snapshot_amount` is the vault balance already folded into it, and each voucher carries a watermark in the same unit. `distribute_v0` pays a voucher the accumulator's growth since that watermark, times the voucher's fraction of `total_shares`.
+
+Revenue is split across **staked** shares, not across all shares. Every arrival is scaled by `total_shares / total_staked_shares`, so the staked vouchers together take all of it and unstaked shares take none. A voucher holding a tenth of `total_shares` while it is the only staked position receives the whole arrival, not a tenth of it.
+
+## Behaviour to know before building on this
+
+- **Joining does not dilute the vouchers already staked.** `stake_v0` folds the pending balance into the accumulator at the previous staked-share count before the new shares join.
+- **Unstaking forfeits to the vouchers that remain.** A voucher closes without collecting what it is owed, and `unstake_v0` releases that amount back so the next fold pays it to the vouchers still staked. Collect first if you want it yourself.
+- **Revenue that arrives before the first stake goes to the first staker.** With nothing staked there is no share count to scale by, so the balance waits for the first fold. That includes a balance already sitting in the vault when `initialize_fanout_v0` runs. Stake before pointing revenue at the vault.
+- **An arrival is folded only as far as the accumulator has room for.** The snapshot advances by exactly the part that was folded and the remainder stays pending, so the fold is callable at every share ratio rather than refusing once the scaling grows large.
+- **`total_shares` is fixed** at the membership mint's supply when `initialize_fanout_v0` runs, and that supply must be non-zero. Staked shares in aggregate cannot exceed it, and `stake_v0` returns `TooManyShares` for a stake that would. Minting more membership tokens does not raise the cap, though tokens minted later can be staked while the aggregate stays under it.
+- **A stake is at least one share.** `stake_v0` returns `ZeroStake` otherwise, so every voucher in the membership collection is a real position.
+- **`fanout.authority` is stored at creation and never read again.** `initialize_fanout_v0` sends the collection NFT to that account's associated token account; after that, no instruction consults the field. There is no admin, no pause, and no route for tokens to leave the vault except `distribute_v0`, which will not pay into the vault itself.
+- **Dust below one token per distribution is carried** on the voucher in `total_dust` at twelve extra decimal places and paid out once it reaches a whole unit.
 
 SDK: [`@helium/fanout-sdk`](../../packages/fanout-sdk). Metadata: [`fanout-metadata-service`](../../packages/fanout-metadata-service).
 
--- a/programs/fanout/src/instructions/stake_v0.rs
+++ b/programs/fanout/src/instructions/stake_v0.rs
@@ -10,7 +10,7 @@ use anchor_spl::{
   token::{self, Mint, MintTo, Token, TokenAccount, Transfer},
 };
 
-use crate::{fanout_seeds, voucher_seeds, FanoutV0, FanoutVoucherV0};
+use crate::{errors::ErrorCode, fanout_seeds, voucher_seeds, FanoutV0, FanoutVoucherV0};
 
 #[cfg(feature = "devnet")]
 const URL: &str = "https://fanout.nft.test-helium.com";
@@ -143,6 +143,30 @@ impl<'info> StakeV0<'info> {
 }
 
 pub fn handler(ctx: Context<StakeV0>, args: StakeArgsV0) -> Result<()> {
+  // Whatever is already in the vault belongs to the shares staked ahead of this
+  // one, so attribute it before this stake joins the staked set.
+  ctx
+    .accounts
+    .fanout
+    .accrue_inflow(ctx.accounts.token_account.amount)?;
+
+  // Every voucher in the membership collection is a real position.
+  require_gt!(args.amount, 0, ErrorCode::ZeroStake);
+
+  let total_staked_shares = ctx
+    .accounts
+    .fanout
+    .total_staked_shares
+    .checked_add(args.amount)
+    .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
+  // Every share a distribution pays against is one of `total_shares`, which is
+  // fixed at the membership supply the fanout was created with.
+  require_gte!(
+    ctx.accounts.fanout.total_shares,
+    total_staked_shares,
+    ErrorCode::TooManyShares
+  );
+
   // Create voucher
   ctx.accounts.voucher.set_inner(FanoutVoucherV0 {
     fanout: ctx.accounts.fanout.key(),
@@ -150,16 +174,11 @@ pub fn handler(ctx: Context<StakeV0>, args: StakeArgsV0) -> Result<()> {
     total_distributed: 0,
     shares: args.amount,
     stake_account: ctx.accounts.stake_account.key(),
-    total_inflow: ctx.accounts.token_account.amount,
+    total_inflow: ctx.accounts.fanout.total_inflow,
     total_dust: 0,
     bump_seed: ctx.bumps.voucher,
   });
-  ctx.accounts.fanout.total_staked_shares = ctx
-    .accounts
-    .fanout
-    .total_staked_shares
-    .checked_add(args.amount)
-    .unwrap();
+  ctx.accounts.fanout.total_staked_shares = total_staked_shares;
 
   // Stake tokens
   token::transfer(ctx.accounts.stake_ctx(), args.amount)?;
--- a/programs/fanout/src/instructions/unstake_v0.rs
+++ b/programs/fanout/src/instructions/unstake_v0.rs
@@ -1,4 +1,4 @@
-use crate::{state::*, voucher_seeds};
+use crate::{errors::ErrorCode, state::*, voucher_seeds, TWELVE_PREC};
 use anchor_lang::prelude::*;
 use anchor_spl::associated_token::AssociatedToken;
 use anchor_spl::token::{self, Burn, CloseAccount, Mint, Token, TokenAccount, Transfer};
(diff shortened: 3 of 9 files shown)


programs/fanout/src/state.rs
```diff
@@ -17,6 +19,49 @@ pub struct FanoutV0 {
...
+      .checked_mul(u128::from(unstaked_shares))
+      .and_then(|scaled| scaled.checked_div(u128::from(self.total_staked_shares)))
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))? as u64;
+
+    self.total_inflow = self
+      .total_inflow
+      .checked_add(inflow)
+      .and_then(|total| total.checked_add(unstaked_correction))
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
+    self.last_snapshot_amount = curr_balance;
```
bryzettler: The `as u64` on `unstaked_correction` narrows without a check, so a correction wider than `u64` wraps silently: `total_inflow` moves by a wrong amount while `last_snapshot_amount` still advances past the arrival. `inflow + inflow * (total_shares - tss) / tss` is also the same integer as `inflow * total_shares / tss` whenever `inflow` is whole, which is the scaling the doc comment above already describes, so collapsing to the single product both matches the doc and puts the whole result behind one `u64::try_from`.

```suggestion
    let inflow = curr_balance
      .checked_sub(self.last_snapshot_amount)
      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
    let scaled = u128::from(inflow)
      .checked_mul(u128::from(self.total_shares))
      .and_then(|scaled| scaled.checked_div(u128::from(self.total_staked_shares)))
      .and_then(|scaled| u64::try_from(scaled).ok())
      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;

    self.total_inflow = self
      .total_inflow
      .checked_add(scaled)
      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
    self.last_snapshot_amount = curr_balance;
```

programs/fanout/src/state.rs
```diff
@@ -52,3 +97,103 @@ macro_rules! fanout_seeds {
...
+
+  #[test]
+  fn a_fold_stays_callable_at_any_share_ratio() {
+    // `stake_v0` folds before it raises `total_staked_shares`, and raising it is
+    // the only way the ratio grows, so the fold has to remain callable however
+    // small the staked set is.
+    let mut f = fanout(u64::MAX, 1, 0);
+    f.accrue_inflow(2).expect("fold stays callable");
+    assert_eq!(f.last_snapshot_amount, 2);
+  }
```
bryzettler, line 225: This case is exactly the one the bare `as u64` used to wrap on (`2 * u64::MAX / 1`), and asserting only `last_snapshot_amount` pins the truncation as the expected behaviour. Asserting the error and that the snapshot stays at zero pins the property worth having instead: the snapshot never advances past an arrival that was not accumulated. The other cases in this module (500, 100, 60, 400, 201) are unchanged by the collapse.

```suggestion
  #[test]
  fn a_scaled_arrival_wider_than_u64_is_an_error() {
    // 2 * u64::MAX / 1 does not fit the accumulator, so the fold reports it
    // instead of wrapping and advancing the snapshot past a lost arrival.
    let mut f = fanout(u64::MAX, 1, 0);
    assert!(f.accrue_inflow(2).is_err());
    assert_eq!(f.last_snapshot_amount, 0);
  }
```

programs/fanout/src/instructions/stake_v0.rs
```diff
@@ -143,23 +143,39 @@ impl<'info> StakeV0<'info> {
...
+    .accounts
+    .fanout
+    .accrue_inflow(ctx.accounts.token_account.amount)?;
+
+  let total_staked_shares = ctx
+    .accounts
+    .fanout
+    .total_staked_shares
+    .checked_add(args.amount)
+    .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
```
bryzettler, line 161: `require_gte!` admits `args.amount == 0`, which mints a voucher NFT into the membership collection that holds no shares, can never be paid by `distribute_v0`, and can only be cleaned up by unstaking it. Rejecting it here keeps every voucher in the collection a real position.

```suggestion
  require_gt!(args.amount, 0, ErrorCode::ZeroStake);
  let total_staked_shares = ctx
    .accounts
    .fanout
    .total_staked_shares
    .checked_add(args.amount)
    .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
```

This needs a new variant in `programs/fanout/src/errors.rs`, after `InvalidDestination`:

```rust
  #[msg("A stake must be at least one share")]
  ZeroStake,
```

tests/fanout.ts
```diff
@@ -122,6 +180,113 @@ describe("fanout", () => {
...
+          .accountsPartial({ fanout, owner: fanout!, mint: member.mint.publicKey })
+          .rpc({ skipPreflight: false });
+      } catch (e: any) {
+        err = e;
+      }
+
+      expect(err, "distribute to the vault was accepted").to.not.eq(undefined);
+      expect(err.error?.errorCode?.code).to.eq("InvalidDestination");
+
+    });
```
bryzettler: `stake_v0.recipient` is an unchecked `AccountInfo` and `receipt_account` is an `init_if_needed` ATA on it, so the stake can mint the receipt to the fanout PDA directly and the hand-built ATA-create plus SPL transfer is not needed. Funding the vault is dead work too: the `InvalidDestination` constraint fires during account validation, before the handler reads the balance.

```suggestion
    it("refuses a distribution whose destination is the vault", async () => {
      const mint = Keypair.generate();
      // The receipt has to sit in an account owned by the fanout for `owner` to
      // validate, so mint it straight to the fanout PDA.
      await stake(mint, fanout!, 100);

      let err: any;
      try {
        await program.methods
          .distributeV0()
          .accountsPartial({ fanout, owner: fanout!, mint: mint.publicKey })
          .rpc({ skipPreflight: false });
      } catch (e: any) {
        err = e;
      }

      expect(err, "distribute to the vault was accepted").to.not.eq(undefined);
      expect(err.error?.errorCode?.code).to.eq("InvalidDestination");
    });
```

That leaves `createAssociatedTokenAccountInstruction` and `createTransferInstruction` (from `@solana/spl-token`) and `Transaction` (from `@solana/web3.js`) unused at the top of the file; the two import blocks collapse back to the single-line forms they had before.

bryzettler (commented):
Review of the share accounting: one checked-narrowing fix in `accrue_inflow`, a zero-stake guard, two test simplifications, and three notes on stranded balances at init and unstake.

tests/fanout.ts
```diff
@@ -139,25 +304,80 @@ describe("fanout", () => {
...
+        const surplus = 81;
+
+        await createAtaAndMint(provider, membershipMint, surplus);
+        await createAtaAndTransfer(
+          provider,
+          membershipMint,
+          surplus,
+          me,
+          voucher
+        );
```
bryzettler: `createAtaAndMint` already takes a `to` and derives the ATA off-curve, so minting straight into the voucher's stake ATA reaches the same end state in one transaction instead of a mint to the provider wallet followed by an ATA-create-and-transfer.

```suggestion
        await createAtaAndMint(provider, membershipMint, surplus, voucher);
```

programs/fanout/README.md
```diff
@@ -1,11 +1,27 @@
...
+
+## Behaviour to know before building on this
+
+- **Joining does not dilute the vouchers already staked.** `stake_v0` folds the pending balance into the accumulator at the previous staked-share count before the new shares join.
+- **Unstaking forfeits, and part of the forfeit is stranded.** A voucher closes without collecting what it is owed. Whatever the accumulator has not yet folded in goes to the vouchers that remain; whatever it has already folded in stays in the vault, since the accumulator only grows by the balance above `last_snapshot_amount` and no voucher's watermark reaches back over it. The voucher's carried dust goes with the closed account. Collect before unstaking.
+- **Revenue that arrives after creation but before the first stake goes to the first staker.** With nothing staked there is no share count to scale by, so the balance waits for the first fold. A balance already in the vault when `initialize_fanout_v0` runs is different: it becomes the starting value of both the accumulator and the snapshot, so it is never distributed at all. Stake before pointing revenue at the vault.
```
bryzettler: The second half of this bullet documents a balance already in the vault at creation being stranded forever. That comes from `initialize_fanout_v0` seeding `last_snapshot_amount` and `total_inflow` from `token_account.amount`. Now that `accrue_inflow` leaves an arrival pending until something is staked and voucher watermarks seed from `total_inflow`, the init seed's only effect is the strand. Seeding both to zero makes the pre-existing balance the first fold's arrival, and the bullet collapses to one rule:

```rust
    last_snapshot_amount: 0,
    total_inflow: 0,
```

programs/fanout/README.md
```diff
@@ -1,11 +1,27 @@
...
+- **`total_shares` is fixed** at the membership mint's supply when `initialize_fanout_v0` runs. Staked shares in aggregate cannot exceed it, and `stake_v0` returns `TooManyShares` for a stake that would. Minting more membership tokens does not raise the cap, though tokens minted later can be staked while the aggregate stays under it.
```
bryzettler: With `TooManyShares` now enforced, a fanout initialized while `membership_mint.supply == 0` gets `total_shares = 0`, every non-zero `stake_v0` fails forever, and the name-only PDA seed burns the name. A `require_gt!(ctx.accounts.membership_mint.supply, 0, ...)` at the top of the `initialize_fanout_v0` handler fails fast instead.

programs/fanout/src/instructions/unstake_v0.rs
```diff
@@ -102,12 +102,14 @@ impl<'info> UnstakeV0<'info> {
...
   let signer_seeds: &[&[&[u8]]] = &[voucher_seeds!(ctx.accounts.voucher)];
+  // `total_staked_shares` is the sum of the live vouchers' `shares`, so closing
+  // this voucher removes exactly what it contributed.
   ctx.accounts.fanout.total_staked_shares = ctx
     .accounts
     .fanout
     .total_staked_shares
-    .checked_sub(ctx.accounts.stake_account.amount)
-    .unwrap();
+    .checked_sub(ctx.accounts.voucher.shares)
```
bryzettler, line 130: The README now documents that an unstake strands the already-folded, uncollected entitlement in the vault, and `distribute_v0` is the only exit. The accumulator already knows that amount. Lowering the snapshot by it here lets the next fold treat it as a fresh arrival for the vouchers that remain:

```rust
  // Release what this voucher was owed but never collected, so the next fold
  // pays it to the vouchers that remain instead of stranding it in the vault.
  let owed = ctx.accounts.fanout.owed_to(&ctx.accounts.voucher)?;
  ctx.accounts.fanout.last_snapshot_amount = ctx
    .accounts
    .fanout
    .last_snapshot_amount
    .checked_sub(owed)
    .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
```

where `FanoutV0::owed_to` is `(total_inflow - voucher.total_inflow) * voucher.shares / total_shares`, the same expression `distribute_v0` computes inline, so both paths would share it. `owed` never exceeds the snapshot because the snapshot is folded minus paid and entitlements are floor-summed.

programs/fanout/src/state.rs
madninja: Took the collapse to the single product, with one change: the narrowing is a cap rather than a `try_from`.

`stake_v0` folds before it raises `total_staked_shares`, and that is the only instruction that raises it, so a fold that returns `Err` at a low share ratio also refuses the stake that would raise the ratio back. `accrue_inflow` now folds as much of the arrival as the accumulator has room for and advances `last_snapshot_amount` by exactly that much, leaving the remainder pending for a later fold. Nothing truncates and nothing is skipped past.

Checked the identity you cite over 200k random `(total_shares, staked, inflow)` triples: `inflow + inflow*(T-s)/s == inflow*T/s` with no counterexamples, so the collapse is exact.

programs/fanout/src/state.rs
madninja: Agreed the old assertion pinned the truncation. Replaced by two tests that pin the property you name instead, without the fold returning `Err`:

- `a_fold_that_does_not_fit_takes_what_does_and_leaves_the_rest` — one share of `u64::MAX` staked, arrival of 2, one unit folds and `last_snapshot_amount` lands on 1, never past the part that was not accumulated.
- `a_full_accumulator_folds_nothing_and_moves_no_snapshot`.

Both mutation-checked: uncapping the admissible inflow, and jumping the snapshot to `curr_balance`, each redden two tests.

programs/fanout/src/instructions/stake_v0.rs
madninja: Taken, with `ZeroStake` added after `InvalidDestination`. Covered by `refuses a stake of zero shares`, and removing the `require_gt!` reddens it.

tests/fanout.ts
madninja: Taken. Staking straight to the fanout PDA drops the hand-built ATA and transfer, and you are right that the vault funding was dead work since the constraint runs during account validation. 2354ms to 493ms, and it still reddens when the constraint is removed.

tests/fanout.ts
madninja: Taken.

programs/fanout/README.md
madninja: Taken. `initialize_fanout_v0` now seeds both fields to `0`, so a balance already in the vault is the first fold's arrival and the bullet collapses to one rule. Added `pays a balance that was in the vault before creation` to cover it; restoring the old seeding reddens it.

programs/fanout/README.md
madninja: Taken, as `require_gt!(ctx.accounts.membership_mint.supply, 0, ErrorCode::NoShares)` at the top of the handler. First pass had the guard with no test and the mutation came back NOT CAUGHT, so there is now a `refuses a fanout whose membership mint has no supply` test and removing the guard reddens it.

programs/fanout/src/instructions/unstake_v0.rs
madninja: Taken. `unstake_v0` lowers the snapshot by what the closing voucher was owed, and `FanoutV0::owed_to` is now shared with `distribute_v0`, which derives both its payout and its dust remainder from it.

Checked the precondition you assert rather than assuming it: 3000 randomized stake/arrive/distribute/unstake sequences, no case where `owed` exceeded the snapshot, and total left in the vaults dropped. It is a `checked_sub` returning `ArithmeticError` regardless. Covered by `releases an uncollected entitlement to the vouchers that remain`.

bryzettler (approved):

PR #1290, opened 2026-09-02
mini-fanout: record a next pre task only when one occupies the slot

### LINKED ISSUE #1255: mini-fanout: total_shares sums into u32 and can overflow
`distribute_v0` sums share weights into a `u32`:

```rust
let total_shares: u32 = share_indices
  .iter()
  .map(|&i| mini_fanout.shares[i].share.as_u32())
  .sum();
```

`overflow-checks = true` is set for the release profile, so this panics rather than wraps once the
weights total more than `u32::MAX`. The weights are chosen by the fanout owner, so a fanout with
large enough `Share::Share` amounts cannot distribute at all: every attempt aborts.

The blast radius is that one fanout, and it is recoverable — the owner can lower the weights
through `update_mini_fanout_v0` — so this is not urgent. Widening the sum to `u128` is the fix;
`share.as_u128()` is already used a few lines below for the same value.

Worth a test that a fanout whose weights exceed `u32::MAX` still distributes, since nothing
currently exercises that range.

Raised during review of #1254 and deliberately kept out of it: that PR is a lint sweep, and this is
unrelated arithmetic in a program that moves tokens, so it deserves its own change and its own test
rather than riding an approved diff.

madninja:
Four invariants in `mini-fanout`, plus the arithmetic width one of them divides by.

## Free task slots

tuktuk takes one free task account per task a run returns and validates it as it is taken, so a recorded slot is one it accepted. `distribute_v0` recorded both the task slot and the pre task slot regardless of how many tasks it returned.

It now records the pre task slot only when a pre task takes it, keeps the fanout's own key as the sentinel for there being none, and declares `free_tasks` as the number of tasks the next run returns. `schedule_task_v0` declares it the same way — whether a cycle queues a pre task is a property of the fanout, since `pre_task` is fixed when it is created.

A slot lookup that runs past the accounts a run was given returns `MissingFreeTask` rather than indexing out of bounds.

## Who may close what is in the slot

Only a fanout that queues a pre task ever owned the account in that slot, so `update_mini_fanout_v0` and `close_mini_fanout_v0` dequeue there only when it has one. Anything else in that slot belongs to whoever queued it, and a fanout with no pre task is neither able to close it nor trapped behind it.

## One decision, one write

`schedule_impl` decides `next_task` and `next_pre_task`, and `get_task_ix` reads `next_pre_task` back to build the account list the queued task carries, which `distribute_v0` checks against the stored one. The update instructions hand `schedule_impl` a clone, so they carry those two values back to the fanout's own copy rather than recomputing them — which is what holds the queued transaction and the stored state to the same key.

## Share weight width

Share weights are summed into a `u128`, the width the divisor two lines below already uses, so weights totalling more than a `u32` divide rather than abort under the release profile's overflow checks. `Share::as_u32` had no remaining caller. Closes #1255.

## Settling what a share member is owed

A payout whose transfer cannot land is recorded on the share as owed. The fixed branch reads that back on the next distribution; the share branch did not. It is now settled off the top and in order, the way a fixed payout is settled, so the proportional split divides only what is left. Two members on even weights end on the even split whether or not one of them had somewhere to receive at the time.

This changes what existing fanouts pay out: a share member carrying an owed balance receives it on the next distribution, ahead of the proportional split, and the members who had been receiving that share stop.

## Verification

`cargo test -p mini-fanout --lib`, `tests/mini-fanout-bankrun.ts` (10), `tests/mini-fanout.ts` on localnet (10), `check-cu-table`, plus clippy with the repo's flags and `cargo fmt --check`. `@helium/idls` changeset included for the new error.

### DIFF
--- a/packages/helium-admin-cli/src/reschedule-all-mini-fanouts.ts
+++ b/packages/helium-admin-cli/src/reschedule-all-mini-fanouts.ts
@@ -55,7 +55,9 @@ export async function run(args: any = process.argv) {
   // scheduleTaskV0 only accepts a fanout whose next_task/next_pre_task are
   // already the "nothing scheduled" sentinel (pointing at the fanout itself,
   // or legacy: at the program id) -- see its on-chain constraint. Anything
-  // else means a task is genuinely still live, and dequeuing a live task
+  // else is either a live task or a key left over from a free task slot the fanout
+  // recorded but never owned. schedule_task_v0 accepts the latter, since its account is
+  // empty, and normalises it to the sentinel; a live task it refuses. Dequeuing a live task
   // requires a CPI signature from the mini-fanout program's own
   // queue_authority PDA (see queue_authority_seeds! in state.rs), which only
   // update_mini_fanout_v0 / close_mini_fanout_v0 can produce -- both of which
@@ -67,10 +69,31 @@ export async function run(args: any = process.argv) {
     miniFanout: (typeof allMiniFanouts)[number],
     task: PublicKey
   ) => task.equals(program.programId) || task.equals(miniFanout.publicKey);
+  // A recorded key whose account does not exist is a free task slot the fanout claimed
+  // but never owned, which schedule_task_v0 accepts and normalises to the sentinel. Only
+  // a key with an account behind it is a task that is genuinely still live.
+  const claimed = miniFanouts.flatMap((mf) =>
+    [mf.account.nextTask, mf.account.nextPreTask].filter(
+      (task) => !isIdleSentinel(mf, task)
+    )
+  );
+  const existing = new Set<string>();
+  for (let i = 0; i < claimed.length; i += 100) {
+    const chunk = claimed.slice(i, i + 100);
+    const infos = await provider.connection.getMultipleAccountsInfo(chunk);
+    infos.forEach((info, j) => {
+      if (info) {
+        existing.add(chunk[j].toBase58());
+      }
+    });
+  }
+  const isIdle = (
+    miniFanout: (typeof allMiniFanouts)[number],
+    task: PublicKey
+  ) => isIdleSentinel(miniFanout, task) || !existing.has(task.toBase58());
   const idle = miniFanouts.filter((mf) => {
     const ok =
-      isIdleSentinel(mf, mf.account.nextTask) &&
-      isIdleSentinel(mf, mf.account.nextPreTask);
+      isIdle(mf, mf.account.nextTask) && isIdle(mf, mf.account.nextPreTask);
     if (!ok) {
       console.log(
         `Skipping ${mf.publicKey.toBase58()}: still has a live scheduled task; ` +
--- a/programs/mini-fanout/src/instructions/distribute_v0.rs
+++ b/programs/mini-fanout/src/instructions/distribute_v0.rs
@@ -23,9 +23,11 @@ pub struct DistributeV0<'info> {
   pub mini_fanout: Box<Account<'info, MiniFanoutV0>>,
   #[account(mut)]
   pub task_queue: Box<Account<'info, TaskQueueV0>>,
-  /// CHECK: Make sure this is empty, pre task needs to be run before task.
+  /// CHECK: A queued pre task has to run before the distribution it precedes. A fanout that
+  /// queues none has nothing to wait for, and this handler only writes the slot, so the
+  /// account there does not gate it.
   #[account(
-    constraint = next_pre_task.data_is_empty() || next_pre_task.key() == mini_fanout.key() @ ErrorCode::PreTaskNotRun
+    constraint = mini_fanout.pre_task.is_none() || next_pre_task.data_is_empty() || next_pre_task.key() == mini_fanout.key() @ ErrorCode::PreTaskNotRun
   )]
   pub next_pre_task: UncheckedAccount<'info>,
   #[account(mut)]
@@ -41,12 +43,6 @@ pub struct DistributeV0<'info> {
 }
 
 impl Share {
-  pub fn as_u32(&self) -> u32 {
-    match self {
-      Share::Share { amount } => *amount,
-      Share::Fixed { amount: _ } => 0,
-    }
-  }
   pub fn as_u128(&self) -> u128 {
     match self {
       Share::Share { amount } => *amount as u128,
@@ -148,12 +144,28 @@ pub fn handler<'info>(
   }
 
   // 4. Calculate total shares for percent distribution
-  let total_shares: u32 = share_indices
+  let total_shares: u128 = share_indices
     .iter()
-    .map(|&i| mini_fanout.shares[i].share.as_u32())
+    .map(|&i| mini_fanout.shares[i].share.as_u128())
     .sum();
 
-  // 5. Assign share payouts
+  // 5. Settle what a share member is still owed from a distribution whose transfer
+  // could not land, off the top and in order, so the proportional split below divides
+  // only what is left.
+  let mut owed_payouts = vec![0u64; mini_fanout.shares.len()];
+  for &i in &share_indices {
+    let owed = mini_fanout.shares[i].total_owed as u128;
+    if owed == 0 {
+      continue;
+    }
+    let paid = owed.min(remaining);
+    // `owed` came from a u64 and `paid` is clamped to it, so both fit.
+    mini_fanout.shares[i].total_owed = (owed - paid) as u64;
+    owed_payouts[i] = paid as u64;
+    remaining = remaining.saturating_sub(paid);
+  }
+
+  // 6. Assign share payouts
   for &i in &share_indices {
     let share = &mini_fanout.shares[i];
     let share_val = share.share.as_u128();
@@ -162,7 +174,7 @@ pub fn handler<'info>(
       .ok_or_else(|| error!(ErrorCode::ArithmeticError))?
       .checked_mul(DUST_PRECISION)
       .ok_or_else(|| error!(ErrorCode::ArithmeticError))?
-      .checked_div(total_shares as u128)
+      .checked_div(total_shares)
       .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
     let payout =
       u64::try_from(amount / DUST_PRECISION).map_err(|_| error!(ErrorCode::ArithmeticError))?;
@@ -180,6 +192,9 @@ pub fn handler<'info>(
       payouts[i] = payout;
       new_dusts[i] = dust as u128;
     }
+    payouts[i] = payouts[i]
+      .checked_add(owed_payouts[i])
+      .ok_or_else(|| error!(ErrorCode::ArithmeticError))?;
   }
 
   let token_account_info = token_account.to_account_info();
@@ -237,8 +252,28 @@ pub fn handler<'info>(
     });
   }
 
-  mini_fanout.next_task = ctx.remaining_accounts[mini_fanout.shares.len()].key();
-  mini_fanout.next_pre_task = ctx.remaining_accounts[mini_fanout.shares.len() + 1].key();
+  // tuktuk takes one free task account per task returned and validates it as it is
+  // taken, and a validation that fails takes this write with it, so a recorded slot is
+  // one it accepted. The pre task slot is recorded only when a pre task takes it; the
+  // fanout's own key is the sentinel for there being no next pre task.
+  let self_key = mini_fanout.key();
+  let pre_task = mini_fanout.pre_task_to_queue()?;
+  let free_task_base = mini_fanout.shares.len();
+  let free_task_at = |offset: usize| -> Result<Pubkey> {
+    Ok(
+      ctx
+        .remaining_accounts
+        .get(free_task_base + offset)
+        .ok_or_else(|| error!(ErrorCode::MissingFreeTask))?
+        .key(),
+    )
+  };
+
+  mini_fanout.next_task = free_task_at(0)?;
+  mini_fanout.next_pre_task = match pre_task {
+    Some(_) => free_task_at(1)?,
+    None => self_key,
+  };
 
   // Schedule next task via tuktuk CPI if funds available, else set next_task = Pubkey::default()
   let next_time = get_next_time(mini_fanout)?;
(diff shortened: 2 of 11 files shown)


bryzettler (approved):

bryzettler (commented):
Four gaps around the same sentinel (the distribute constraint, the reschedule CLI filter, and two in `update_wallet_delegate_v0`), plus two test cleanups.

programs/mini-fanout/src/instructions/distribute_v0.rs
```diff
@@ -237,8 +250,28 @@ pub fn handler<'info>(
...
+        .get(free_task_base + offset)
+        .ok_or_else(|| error!(ErrorCode::MissingFreeTask))?
+        .key(),
+    )
+  };
+
+  mini_fanout.next_task = free_task_at(0)?;
+  mini_fanout.next_pre_task = match pre_task {
+    Some(_) => free_task_at(1)?,
+    None => self_key,
```
bryzettler, line 275: The `next_pre_task` constraint at line 28 still applies when `mini_fanout.pre_task` is `None`. A fanout with no pre task whose stored `next_pre_task` is a leftover free task slot key that something else now occupies has a non-empty account there, so the constraint fails `PreTaskNotRun` on every run and this `None => self_key` repair never executes. Nothing else in `distribute_v0` reads `next_pre_task` when `pre_task` is `None`, and `update_mini_fanout_v0` / `close_mini_fanout_v0` already skip that slot in that case.

Replace lines 27-29 with:

```rust
  #[account(
    constraint = mini_fanout.pre_task.is_none() || next_pre_task.data_is_empty() || next_pre_task.key() == mini_fanout.key() @ ErrorCode::PreTaskNotRun
  )]
```

programs/mini-fanout/src/instructions/update_wallet_delegate_v0.rs
```diff
@@ -88,30 +88,33 @@ pub fn handler(
     ))?;
   }
 
+  // `schedule_impl` writes `next_task` and `next_pre_task` on the fanout it is given, and
```
bryzettler, line 91: The handler dequeues `next_task` but never `next_pre_task`, then `schedule_impl` queues a new pre task and line 117 overwrites the stored pointer, so a live pre task stays queued with nothing referencing it; `update_mini_fanout_v0.rs` lines 118-135 already do this dequeue.

```suggestion
  // Only a fanout that queues a pre task ever owned the account in that slot, so only it
  // may close one. Anything else there belongs to whoever queued it.
  if mini_fanout.pre_task.is_some()
    && !ctx.accounts.next_pre_task.data_is_empty()
    && ctx.accounts.next_pre_task.key() != mini_fanout.key()
  {
    dequeue_task_v0(CpiContext::new_with_signer(
      ctx.accounts.tuktuk_program.to_account_info(),
      DequeueTaskV0 {
        task_queue: ctx.accounts.task_queue.to_account_info(),
        task: ctx.accounts.next_pre_task.to_account_info(),
        queue_authority: ctx.accounts.queue_authority.to_account_info(),
        rent_refund: ctx.accounts.payer.to_account_info(),
        task_queue_authority: ctx.accounts.task_queue_authority.to_account_info(),
      },
      &[queue_authority_seeds!(mini_fanout)],
    ))?;
  }

  // `schedule_impl` writes `next_task` and `next_pre_task` on the fanout it is given, and
```

programs/mini-fanout/src/instructions/update_wallet_delegate_v0.rs
```diff
@@ -88,30 +88,33 @@ pub fn handler(
     ))?;
   }
```
bryzettler, line 89: `if !ctx.accounts.next_task.data_is_empty()` at line 77 has no sentinel check. `has_one = next_task` forces the fanout itself to be passed as `next_task` whenever the stored key is the fanout key -- a freshly initialized fanout, or one that took `distribute_v0`'s insufficient-lamports branch -- and that account is not empty, so the handler CPIs `dequeue_task_v0` with the fanout as the task and fails with `AccountOwnedByWrongProgram`. `update_mini_fanout_v0.rs` line 105 and `close_mini_fanout_v0.rs` line 98 both include `&& next_task.key() != mini_fanout.key()`.

Replace line 77 with:

```rust
  if !ctx.accounts.next_task.data_is_empty() && ctx.accounts.next_task.key() != mini_fanout.key() {
```

packages/helium-admin-cli/src/reschedule-all-mini-fanouts.ts
```diff
@@ -55,7 +55,9 @@ export async function run(args: any = process.argv) {
   // scheduleTaskV0 only accepts a fanout whose next_task/next_pre_task are
   // already the "nothing scheduled" sentinel (pointing at the fanout itself,
   // or legacy: at the program id) -- see its on-chain constraint. Anything
-  // else means a task is genuinely still live, and dequeuing a live task
+  // else is either a live task or a key left over from a free task slot the fanout
+  // recorded but never owned. schedule_task_v0 accepts the latter, since its account is
+  // empty, and normalises it to the sentinel; a live task it refuses. Dequeuing a live task
```
bryzettler, line 60: The comment says `schedule_task_v0` accepts a leftover `next_pre_task` whose account is empty and normalises it to the sentinel, but the filter below still requires `isIdleSentinel(mf, mf.account.nextPreTask)`, so exactly those fanouts are skipped and logged as still having a live scheduled task.

Replace lines 72-75 with:

```ts
  // A recorded key whose account does not exist is a free task slot the fanout claimed
  // but never owned, which schedule_task_v0 accepts and normalises to the sentinel. Only
  // a key with an account behind it is a task that is genuinely still live.
  const claimed = miniFanouts.flatMap((mf) =>
    [mf.account.nextTask, mf.account.nextPreTask].filter(
      (task) => !isIdleSentinel(mf, task)
    )
  );
  const existing = new Set<string>();
  for (let i = 0; i < claimed.length; i += 100) {
    const chunk = claimed.slice(i, i + 100);
    const infos = await provider.connection.getMultipleAccountsInfo(chunk);
    infos.forEach((info, j) => {
      if (info) {
        existing.add(chunk[j].toBase58());
      }
    });
  }
  const isIdle = (
    miniFanout: (typeof allMiniFanouts)[number],
    task: PublicKey
  ) => isIdleSentinel(miniFanout, task) || !existing.has(task.toBase58());
  const idle = miniFanouts.filter((mf) => {
    const ok =
      isIdle(mf, mf.account.nextTask) && isIdle(mf, mf.account.nextPreTask);
```

tests/mini-fanout-bankrun.ts
```diff
@@ -326,6 +326,93 @@ describe("mini-fanout under bankrun", () => {
...
+    const distribute = await reachDistribution(task, preTask);
+    await distribute();
+
+    const owed = (await program.account.miniFanoutV0.fetch(miniFanout))
+      .shares[1].totalOwed;
+    expect(owed.toNumber()).to.be.greaterThan(0);
+
+    // Give wallet2 somewhere to receive, then reach the next distribution.
+    const wallet2Ata = await ataWith(wallet2.publicKey, 0);
+    const acc = await program.account.miniFanoutV0.fetch(miniFanout);
```
bryzettler: The fanout is fetched twice with nothing in between that changes it, and `owed` is only read once.

```suggestion
    const acc = await program.account.miniFanoutV0.fetch(miniFanout);
    expect(acc.shares[1].totalOwed.toNumber()).to.be.greaterThan(0);

    // Give wallet2 somewhere to receive, then reach the next distribution.
    const wallet2Ata = await ataWith(wallet2.publicKey, 0);
```

tests/mini-fanout-bankrun.ts
```diff
@@ -389,6 +476,155 @@ describe("mini-fanout under bankrun", () => {
...
+    await ataWith(wallet.publicKey, 0);
+    const { miniFanout } = await scheduledFanout({
+      seed: "clears-stale",
+      shares: [shareOf(wallet.publicKey, 100)],
+      preTask: null,
+    });
+
+    // Plant a key no instruction writes: a fanout holding a claim on a slot it does not
+    // own is the state this clears, and only a direct rewrite can produce it.
+    const original = await readAccount(ctx, miniFanout);
```
bryzettler: The plant-a-stale-`next_pre_task` sequence is written out three times (~490-497, ~547-554, ~595-601), differing only in the planted key, and the `updateMiniFanoutV0` reschedule call is written out twice (~502-522, ~604-619). Two describe-scoped helpers, `plantNextPreTask` and `rescheduleViaUpdate`, next to `scheduledFanout` / `reachDistribution` cover both; `plantNextPreTask` decodes the fanout from the bytes it already read instead of fetching a second time.

Add next to `scheduledFanout` / `reachDistribution`:

```ts
  /**
   * Rewrite the fanout's stored `next_pre_task` to `key`. A fanout holding a claim on a
   * slot it does not own is a state no instruction writes, so only a direct rewrite of
   * the account produces it.
   */
  async function plantNextPreTask(miniFanout: PublicKey, key: PublicKey) {
    const original = await readAccount(ctx, miniFanout);
    const before = program.coder.accounts.decode(
      "miniFanoutV0",
      Buffer.from(original!)
    );
    const planted = Buffer.from(original!);
    const at = planted.indexOf(Buffer.from(before.nextPreTask.toBytes()));
    expect(at).to.be.greaterThan(-1);
    planted.set(key.toBytes(), at);
    await overwriteAccountData(ctx, miniFanout, planted);
  }

  /** Reschedule through `update_mini_fanout_v0`, changing nothing but the task ids. */
  async function rescheduleViaUpdate(miniFanout: PublicKey) {
    const { taskBitmap } = await tuktukProgram.account.taskQueueV0.fetch(
      taskQueue
    );
    const [newPreTaskId, newTaskId] = nextAvailableTaskIds(taskBitmap, 2, true);
    await program.methods
      .updateMiniFanoutV0({ shares: null, schedule: null, newTaskId, newPreTaskId })
      .preInstructions([
        ComputeBudgetProgram.setComputeUnitLimit({ units: 1400000 }),
      ])
      .accounts({
        payer: me,
        owner: me,
        miniFanout,
        newTask: taskKey(taskQueue, newTaskId)[0],
        newPreTask: taskKey(taskQueue, newPreTaskId)[0],
        taskRentRefund: me,
      })
      .rpc();
  }
```

Each call site then becomes one line: `await plantNextPreTask(miniFanout, stale);` (lines 490-497), `await rescheduleViaUpdate(miniFanout);` (lines 502-522), `await plantNextPreTask(miniFanout, Keypair.generate().publicKey);` (lines 547-554), `await plantNextPreTask(miniFanout, taskQueue);` (lines 595-601), and `await rescheduleViaUpdate(miniFanout);` (lines 604-619).

programs/mini-fanout/src/instructions/distribute_v0.rs
madninja: Taken. Verified the premise before changing it: the constraint had no `pre_task.is_none()` arm, and `next_pre_task` is only ever written in this handler (the crank-reward early return, and the slot record) and never read, so the account there does not gate a fanout that queues no pre task.

I could not get a test to cover it. `has_one = next_pre_task` means the account the queued transaction carries has to equal the stored key, and `get_task_ix` bakes that key in at queue time, so a fixture needs the same stale key in both places. `compile_transaction` dedups the sentinel onto the fanout's own entry, so patching the task account cannot separate the two. I tried, got `ConstraintHasOne` instead of reaching the constraint, and removed the test rather than keep one that passes for a different reason. The mutation that removes the `is_none()` arm therefore comes back uncaught, and I would rather say so than describe it as equivalent.

programs/mini-fanout/src/instructions/update_wallet_delegate_v0.rs
madninja: Taken, with a test. `has_one = next_task` does force the fanout itself to be passed when the stored key is the sentinel, and the sibling instructions both carry the check. Covered by `changes a delegate on a fanout with nothing scheduled`, which initializes without scheduling so the sentinel is what `initialize_mini_fanout_v0` wrote; removing the added condition reddens it.

packages/helium-admin-cli/src/reschedule-all-mini-fanouts.ts
madninja: Taken. Confirmed `getMultipleAccountsInfo` is on `Connection` before using it. This also fixes an inconsistency I introduced earlier in the branch: I had corrected the comment to say `schedule_task_v0` accepts an empty-account key while leaving the filter rejecting exactly those, so the comment described behaviour the code below it did not implement.

tests/mini-fanout-bankrun.ts
madninja: Taken.

tests/mini-fanout-bankrun.ts
madninja: Taken, both helpers, next to `scheduledFanout` and `reachDistribution`. Decoding from the bytes already read rather than fetching again is a better shape than what I had.

Worth flagging how I applied it: I collapsed the three sequences with a regex, and it swallowed a `const stale = ...` binding that later lines still referenced, so the suite failed on `stale is not defined`. Repaired by hand.

programs/mini-fanout/src/instructions/update_wallet_delegate_v0.rs
madninja: Not taking this one, and the reason is worth recording.

`dequeue_task_v0` has `has_one = rent_refund` on the task, and this handler has no `task_rent_refund` account — it passes `ctx.accounts.payer` for the `next_task` dequeue above, so a pre task dequeue here would pass the same. But the two ways a task gets queued give it different refund addresses: `queue_task_v0:97` sets `rent_refund = payer`, while `run_task_v0:451` sets `rent_refund = task_queue` for a task returned by a run.

Every live queued distribute task on mainnet has the second shape. I read all 48 `MiniFanoutV0` accounts, followed each non-sentinel `next_task` to its `TaskV0`, and got 31 of 31 with `rent_refund == task_queue`, none with a payer. So a dequeue from this handler matches on a fanout whose tasks were queued by `schedule_task_v0` and not on one whose tasks came from a distribution.

That is why `update_mini_fanout_v0` takes a `task_rent_refund` account and `helium-admin-cli`'s `resolveTaskRentRefund` reads the address off the task itself. Giving this handler the same account is a wider change than this PR — a new account is an IDL change and every caller has to pass it — so I would rather do it deliberately than fold it in here.

I added the pre task dequeue as suggested first, wrote a test that changes a delegate after a distribution has run, and it failed `ConstraintHasOne` on `task`. Reverting just the dequeue reproduced the same failure, which is what places the cause in the existing `next_task` dequeue rather than in the addition.

PR #1295, opened 2026-09-03
fix(blockchain-api): review follow-ups for batch expiry, migration race, e2e helpers

bryzettler:
Follow-ups from review of the blockchain-api batch status / index migration work.

## Fixes

- **Finalized poll wrote a landed transaction expired.** A signature the cluster reports confirmed (not yet finalized) is in a block that will not be dropped. It now stays `pending` until it reaches the requested commitment instead of being expiry-checked and rolling its batch up to `failed`, which made a retry a double claim. Any status the cluster returns without an error (`processed` included) stays `pending`; only a signature the cluster has never seen goes through the blockhash check. Unit test added.
- **Migration race between replicas.** Every replica runs `db:migrate` on start. A concurrent build shows `indisvalid=false` mid-build, so a second replica dropped the first one's in-flight build and failed its start. The migration now holds a session advisory lock on one pooled client for the whole `up()`, and drops an invalid leftover with `DROP INDEX CONCURRENTLY` instead of taking ACCESS EXCLUSIVE on `pending_transactions`. The lock is polled with `pg_try_advisory_lock`: a session blocked inside `pg_advisory_lock` holds a snapshot that `CREATE INDEX CONCURRENTLY` waits on, so the blocking form deadlocked one replica. Verified on a scratch database (postgres 14, 3M rows): fresh build, idempotent rerun, forced-invalid rebuild, two concurrent `up()` calls at 0/1000/3000ms stagger both succeed with valid indexes, no lock leaked, `down` cleans up.
- **Matrix-coverage check could not fire on the change it guards.** The e2e workflow now also runs on edits to itself.
- **extend.ts judged on the raw cluster clock.** Now adds the registrar `time_offset`, matching `delegate.ts`. E2e added that dials the registrar past the delegation's expiration and asserts the refusal.
- **Unchecked `surfnet_setAccount` writes.** One exported `setSurfnetAccount` that throws on a non-2xx or an RPC error replaces the seven inline fetches across the tuktuk, governance and proposal e2e helpers, so a failed time-offset reset no longer leaves the registrar dialed forward silently.
- **`getMultipleAccounts` chunking helper** moved from `build-claim-instructions.ts` to `src/lib/utils/get-multiple-accounts.ts`.

## Not done

- **Post-split bundle cap test.** The measurer and the real build compile identical instructions, lookup tables and compute-budget prefix, so no public input makes the real build split where the measurer said it fit. That makes `buildOrSplit`'s recursive split and the post-split cap equally unreachable from public input; both are defensive, and neither has a test. Testing them needs a mock seam on `buildVersionedTransaction` (swc CJS exports are non-configurable, so the test cannot patch it). Left out rather than add an injection parameter for a test alone.

## Verification

- `pnpm test:unit`: 279 passing
- `pnpm typecheck`, `pnpm lint:check`: clean
- `tests/e2e/delegate-automation.test.ts`: 3 passing

https://claude.ai/code/session_01Lsc2KJ3LtSQY4dvAc9g9bu

### DIFF
--- a/packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
+++ b/packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
@@ -20,6 +20,13 @@
  * locks belong to one connection, so the whole migration runs on a single
  * client taken from the pool rather than through sequelize's pooled query.
  *
+ * The lock is polled with pg_try_advisory_lock rather than taken with the
+ * blocking pg_advisory_lock. A session parked inside pg_advisory_lock holds an
+ * open snapshot, and CREATE INDEX CONCURRENTLY waits for exactly those
+ * snapshots before it can finish, so the holder and the waiter deadlock and
+ * postgres kills one of them. Each try is its own short statement, so the
+ * waiter holds no snapshot between attempts.
+ *
  * @type {import('sequelize-cli').Migration}
  */
 const INDEXES = [
@@ -35,7 +42,16 @@ module.exports = {
     const manager = queryInterface.sequelize.connectionManager;
     const client = await manager.getConnection({ type: "write" });
     try {
-      await client.query("SELECT pg_advisory_lock($1)", [ADVISORY_LOCK_KEY]);
+      for (;;) {
+        const { rows } = await client.query(
+          "SELECT pg_try_advisory_lock($1) AS got",
+          [ADVISORY_LOCK_KEY],
+        );
+        if (rows[0].got) {
+          break;
+        }
+        await new Promise((resolve) => setTimeout(resolve, 1000));
+      }
       try {
         for (const { name, column } of INDEXES) {
           // to_regclass resolves through the search path, the same way the
--- a/packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
+++ b/packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
@@ -13,6 +13,13 @@
  * rebuilt. sequelize-cli does not wrap migrations in a transaction, which
  * CREATE INDEX CONCURRENTLY could not run inside anyway.
  *
+ * Every replica runs db:migrate on start, so two can reach this at once. An
+ * index mid-build also shows as invalid, and the second replica would drop the
+ * first one's live build and fail its start. A session advisory lock serialises
+ * them: the second waits, then finds the valid index and keeps it. Session
+ * locks belong to one connection, so the whole migration runs on a single
+ * client taken from the pool rather than through sequelize's pooled query.
+ *
  * @type {import('sequelize-cli').Migration}
  */
 const INDEXES = [
@@ -20,32 +27,53 @@ const INDEXES = [
   { name: "idx_pending_transactions_batch_id", column: "batch_id" },
 ];
 
+/** Arbitrary but fixed; only this migration takes it. */
+const ADVISORY_LOCK_KEY = 20260902;
+
 module.exports = {
   async up(queryInterface) {
-    for (const { name, column } of INDEXES) {
-      // to_regclass resolves through the search path, the same way the
-      // unqualified CREATE INDEX below does, so a same-named index in another
-      // schema is not mistaken for this one. It yields NULL when none exists.
-      const [rows] = await queryInterface.sequelize.query(
-        `SELECT indisvalid FROM pg_index WHERE indexrelid = to_regclass(:name)`,
-        { replacements: { name } },
-      );
-      if (rows.length > 0 && rows[0].indisvalid) {
-        continue;
+    const manager = queryInterface.sequelize.connectionManager;
+    const client = await manager.getConnection({ type: "write" });
+    try {
+      await client.query("SELECT pg_advisory_lock($1)", [ADVISORY_LOCK_KEY]);
+      try {
+        for (const { name, column } of INDEXES) {
+          // to_regclass resolves through the search path, the same way the
+          // unqualified CREATE INDEX below does, so a same-named index in
+          // another schema is not mistaken for this one. It yields NULL when
+          // none exists.
+          const { rows } = await client.query(
+            "SELECT indisvalid FROM pg_index WHERE indexrelid = to_regclass($1)",
+            [name],
+          );
+          if (rows.length > 0 && rows[0].indisvalid) {
+            continue;
+          }
+          if (rows.length > 0) {
+            // CONCURRENTLY so the drop waits for readers instead of taking
+            // ACCESS EXCLUSIVE on the table and queueing the submit path
+            // behind a long read.
+            await client.query(`DROP INDEX CONCURRENTLY IF EXISTS "${name}"`);
+          }
+          await client.query(
+            `CREATE INDEX CONCURRENTLY "${name}" ON "pending_transactions" ("${column}")`,
+          );
+        }
+      } finally {
+        await client.query("SELECT pg_advisory_unlock($1)", [
+          ADVISORY_LOCK_KEY,
+        ]);
       }
-      if (rows.length > 0) {
-        await queryInterface.sequelize.query(`DROP INDEX IF EXISTS "${name}"`);
-      }
-      await queryInterface.addIndex("pending_transactions", [column], {
-        name,
-        concurrently: true,
-      });
+    } finally {
+      manager.releaseConnection(client);
     }
   },
 
   async down(queryInterface) {
     for (const { name } of INDEXES) {
-      await queryInterface.sequelize.query(`DROP INDEX IF EXISTS "${name}"`);
+      await queryInterface.sequelize.query(
+        `DROP INDEX CONCURRENTLY IF EXISTS "${name}"`,
+      );
     }
   },
 };
--- a/packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
+++ b/packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
@@ -42,13 +42,18 @@ module.exports = {
     const manager = queryInterface.sequelize.connectionManager;
     const client = await manager.getConnection({ type: "write" });
     try {
-      for (;;) {
+      for (let attempt = 0; ; attempt++) {
         const { rows } = await client.query(
           "SELECT pg_try_advisory_lock($1) AS got",
           [ADVISORY_LOCK_KEY],
         );
         if (rows[0].got) {
           break;
+        }
+        if (attempt === 0) {
+          console.log(
+            "Waiting for another db:migrate run to finish building the pending_transactions indexes",
+          );
         }
         await new Promise((resolve) => setTimeout(resolve, 1000));
       }
--- a/packages/blockchain-api/src/server/api/routers/governance/procedures/delegation/extend.ts
+++ b/packages/blockchain-api/src/server/api/routers/governance/procedures/delegation/extend.ts
@@ -80,7 +80,10 @@ export const extend = publicProcedure.governance.extendDelegation.handler(
 
     const clock = await connection.getAccountInfo(SYSVAR_CLOCK_PUBKEY);
     const unixTime = clock!.data.readBigInt64LE(8 * 4);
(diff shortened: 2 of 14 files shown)


packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
```diff
@@ -13,39 +13,67 @@
...
-      const [rows] = await queryInterface.sequelize.query(
-        `SELECT indisvalid FROM pg_index WHERE indexrelid = to_regclass(:name)`,
-        { replacements: { name } },
-      );
-      if (rows.length > 0 && rows[0].indisvalid) {
-        continue;
+    const manager = queryInterface.sequelize.connectionManager;
+    const client = await manager.getConnection({ type: "write" });
+    try {
+      await client.query("SELECT pg_advisory_lock($1)", [ADVISORY_LOCK_KEY]);
```
madninja: **Blocker: this does not stop the two-replica race, it converts it into a deadlock, and it still kills one `db:migrate`.**

A session parked inside `SELECT pg_advisory_lock(...)` holds an open snapshot, and `CREATE INDEX CONCURRENTLY` has to wait for exactly those snapshots before it can finish. So the replica that holds the lock waits on the replica waiting for the lock.

Measured against postgres 16 with 3M rows in `pending_transactions`, two `up()` calls on separate connections:

| second replica starts | result |
| --- | --- |
| +0ms | **A fails: deadlock detected** / B ok |
| +1000ms | **A fails: deadlock detected** / B ok |
| +3000ms | A ok / **B fails: deadlock detected** |

The server log names both sides:

```
Process 233 waits for ExclusiveLock on advisory lock [16384,0,20260902,1]; blocked by process 231.
Process 231 waits for ShareLock on virtual transaction 4/6; blocked by process 233.
Process 233: SELECT pg_advisory_lock($1)
Process 231: CREATE INDEX CONCURRENTLY "idx_pending_transactions_signature" ON "pending_transactions" ("signature")
```

I isolated the mechanism separately, without any second migration involved: with one session blocked on an advisory lock held by an idle third session, an unrelated `CREATE INDEX CONCURRENTLY` sat for 4006ms and completed the instant the lock was released.

Not a regression. I ran the same race against the version that merged in #1286 and it deadlocks 3 for 3 as well. So the behaviour is unchanged and the deploy still loses a replica under `set -e`; it is the claim that this fixes it that does not hold.

I think the verification is what made it look fixed. "Waits on a held lock" tests a waiter against an **idle** holder, which is the one arrangement where no `CREATE INDEX CONCURRENTLY` is running, so it exercises the lock rather than the race.

What does work, on the same harness: poll with `pg_try_advisory_lock` so the waiter holds no snapshot between attempts. Each attempt is its own short statement.

```suggestion
      // Polled rather than blocking: a session parked inside pg_advisory_lock
      // holds an open snapshot, and CREATE INDEX CONCURRENTLY waits for exactly
      // those, so the holder and the waiter deadlock. Each attempt is its own
      // statement, so this session holds no snapshot between tries.
      for (;;) {
        const { rows } = await client.query(
          "SELECT pg_try_advisory_lock($1) AS got",
          [ADVISORY_LOCK_KEY],
        );
        if (rows[0].got) {
          break;
        }
        await new Promise((resolve) => setTimeout(resolve, 1000));
      }
```

With that change and nothing else altered, the same three staggers give `A=ok | B=ok`, both indexes valid, and no advisory lock left held.

The other four properties all hold as written, for what it is worth: fresh build, idempotent rerun, forced `indisvalid=false` rebuild, and `down` cleanup were all correct in my run.

packages/blockchain-api/tests/e2e/helpers/surfpool.ts
```diff
@@ -210,3 +211,31 @@ export async function stopSurfpool(timeoutMs = 10_000): Promise<void> {
...
+        address.toBase58(),
+        {
+          data: accountInfo.data.toString("hex"),
+          owner: accountInfo.owner.toBase58(),
+          lamports: accountInfo.lamports,
+        },
+      ],
+    }),
+  });
+  const json = await res.json();
```
madninja: `json.error` is the JSON-RPC error, not the transport one. A non-2xx whose body happens to parse and carries no `error` key still passes silently, which is the case this helper exists to catch.

```suggestion
  if (!res.ok) {
    throw new Error(`setAccount failed: HTTP ${res.status}`);
  }
  const json = await res.json();
```

packages/blockchain-api/src/server/api/routers/governance/procedures/delegation/extend.ts
```diff
@@ -80,7 +80,10 @@ export const extend = publicProcedure.governance.extendDelegation.handler(
 
     const clock = await connection.getAccountInfo(SYSVAR_CLOCK_PUBKEY);
     const unixTime = clock!.data.readBigInt64LE(8 * 4);
-    const now = new BN(Number(unixTime));
+    // The program judges the lockup, expiration and season against
+    // `registrar.clock_unix_timestamp()`, the cluster clock plus the
+    // registrar's time_offset. Zero on mainnet; tests dial it forward.
+    const now = new BN(Number(unixTime)).add(registrar.timeOffset);
```
madninja, line 86: Nothing covers this. Dropping `.add(registrar.timeOffset)` here leaves the unit suite at 279 passing, and there is no e2e for `extendDelegation` against a dialed registrar, so the only thing pinning it is the sibling test on `delegate.ts`.

`setRegistrarTimeOffset` plus the `setSurfnetAccount` helper this PR adds makes it cheap: the same shape as "judges lockup decay on the registrar clock, not the cluster clock", dialing the registrar past `delegatedPositionAcc.expirationTs` and asserting the "Delegation has expired" refusal rather than a bundle the program would reject.

madninja (changes requested):
The finalized-expiry fix is right, and I checked it by driving `decideBatchStatus` on this branch with `commitment: "finalized"` and a blockhash the probe rejects:

| cluster reports | row | batch |
| --- | --- | --- |
| `confirmed` | pending | pending |
| `processed` | pending | pending |
| error | failed | failed |
| never seen | **expired** | failed |

The expiry path survives exactly where it should. Both new guards are covered: deleting the `return "pending"` fails 2 tests, neutering the `err != null` early return fails 3. 279 passing on the branch.

Blocking on the migration: the advisory lock does not prevent the failure it was added for, and I have the postgres log for it. Details in the inline comment.

Two smaller things:

- The body says "Only a `processed` status still goes through the blockhash check". `28831e86` removed that distinction, and `processed` now stays pending too. Worth editing before this squashes.
- On the post-split cap under "Not done": the reasoning holds. `setComputeUnitLimit` and `setComputeUnitPrice` encode fixed-width, so `measureSize` and the real build compile to the same size and `SIZE_MARGIN` covers the rest. But if no input can make `buildOrSplit` split, `buildOrSplit`'s recursive path is equally unreachable, and only the cap is being described that way. Either both are dead or the split can fire and the cap needs the test. Not blocking, but the two should say the same thing.

Verified clean otherwise: the 7-to-1 `surfnet_setAccount` consolidation (only a doc-comment mention of the raw method is left), the `get-multiple-accounts.ts` extraction with no orphaned imports, the workflow `paths:` addition on both triggers, and `extend.ts` fetching the registrar before it derives `now`.

packages/blockchain-api/tests/e2e/helpers/surfpool.ts
```diff
@@ -210,3 +211,34 @@ export async function stopSurfpool(timeoutMs = 10_000): Promise<void> {
...
+    }),
+  });
+  if (!res.ok) {
+    throw new Error(`setAccount failed: HTTP ${res.status}`);
+  }
+  const json = await res.json();
+  if (json.error) {
+    throw new Error(`setAccount failed: ${JSON.stringify(json.error)}`);
+  }
+}
```
bryzettler, line 234: `jsonRpc` a few lines up already builds this exact JSON-RPC envelope, and it carries an `AbortSignal.timeout` that this hand-rolled copy lacks, so a surfpool that accepts the connection but never answers hangs the await forever. Calling it keeps one envelope and gives `setSurfnetAccount` the timeout too.

```suggestion
export async function setSurfnetAccount(
  address: PublicKey,
  accountInfo: { data: Buffer; owner: PublicKey; lamports: number }
): Promise<void> {
  const json = await jsonRpc("surfnet_setAccount", [
    address.toBase58(),
    {
      data: accountInfo.data.toString("hex"),
      owner: accountInfo.owner.toBase58(),
      lamports: accountInfo.lamports,
    },
  ]);
  if (json.error) {
    throw new Error(`setAccount failed: ${JSON.stringify(json.error)}`);
  }
}
```

This depends on one change in `jsonRpc`, which is outside the diff so it cannot be suggested inline: add the HTTP-status check after its `fetch`, before the `res.json()` return, so both callers get it.

```ts
  if (!res.ok) {
    throw new Error(`${method} failed: HTTP ${res.status}`);
  }
```

`surfpoolHealthy` already wraps each of its three `jsonRpc` calls in try/catch, so that throw does not change its behaviour.

packages/blockchain-api/src/lib/migrations/20260902000000-add-pending-transactions-signature-index.js
```diff
@@ -13,39 +13,83 @@
...
+      for (;;) {
+        const { rows } = await client.query(
+          "SELECT pg_try_advisory_lock($1) AS got",
+          [ADVISORY_LOCK_KEY],
+        );
+        if (rows[0].got) {
+          break;
+        }
+        await new Promise((resolve) => setTimeout(resolve, 1000));
       }
```
bryzettler: A run parked on this lock prints nothing until the other replica's `CREATE INDEX CONCURRENTLY` finishes, so its output cannot be told from a hang. Logging once on the first miss names what it is waiting on without spamming a line per second; `20250104000000-update-pending-transactions-table.js` in this directory already reports its progress this way.

```suggestion
      for (let attempt = 0; ; attempt++) {
        const { rows } = await client.query(
          "SELECT pg_try_advisory_lock($1) AS got",
          [ADVISORY_LOCK_KEY],
        );
        if (rows[0].got) {
          break;
        }
        if (attempt === 0) {
          console.log(
            "Waiting for another db:migrate run to finish building the pending_transactions indexes",
          );
        }
        await new Promise((resolve) => setTimeout(resolve, 1000));
      }
```

bryzettler (commented):
Two cleanups; neither blocks. Both applied cleanly against the head with lint, typecheck and unit tests green.

madninja (approved):
All three addressed. I re-ran the migration against a real postgres rather than reading the change, since that was the blocker.

Postgres 16, 3M rows in `pending_transactions`, two `up()` calls on separate connections at four stagger delays:

| second replica starts | result |
| --- | --- |
| +0ms | A ok / B ok, both indexes valid, no lock held |
| +1000ms | A ok / B ok, both indexes valid, no lock held |
| +3000ms | A ok / B ok, both indexes valid, no lock held |
| +8000ms | A ok / B ok, both indexes valid, no lock held |

Same harness that failed 3 for 3 on the blocking lock, and 3 for 3 on the version that merged in #1286. The waiter logs its one line and picks the lock up when the builder releases it. The other four properties still hold on the committed file: fresh build, idempotent rerun, forced `indisvalid=false` rebuild, and `down` leaving nothing behind, each with zero advisory locks after.

The surfpool change went further than the suggestion and is better for it: one `jsonRpc` path with the status guard in it, and `setSurfnetAccount` reduced to its parameters. The three health callers already wrap in `try/catch`, so the new throw does not change what `surfpoolHealthy` reports.

One nit, not worth another round. `HEALTH_REQUEST_TIMEOUT_MS` and the comment above it ("Per-request timeout for health probes") now also govern `surfnet_setAccount`, so a write inherits a 5s abort that nobody chose for it. The value is fine for account writes this size; it is the name and the comment that are narrower than the scope. Worth a rename next time the file is open.

On the new expiry test: it asserts the cause rather than the refusal, which is what makes it worth having. The constant lockup means the decay check cannot fire, the expiry check is the first thing in the handler that can refuse, and the message is unique to it. Without the offset the handler falls through to the `expirationTs.gte(newExpirationTs)` early return and answers with no error at all, so `expect.fail` takes it red. CI's `delegate-automation` job is green on this head.

Checked locally: unit 279 passing, typecheck clean. Body reads correctly now on the `processed` case.

PR #1298, opened 2026-09-04
feat(blockchain-api): delegation claim state in getPositions, and cluster-priced rent after the mainnet Rent change

bryzettler:
## Summary

1. `governance.getPositions` exposes per-position delegation claim state.
2. Every hardcoded rent amount is replaced with a byte size priced from the cluster's Rent sysvar, following the mainnet rent change (3480/2.0x to 6333/1.0x, ~9% cheaper). Two program fixes need releases.

## Delegation claim state in getPositions

Each position gets a `delegation` object (`null` when not delegated): sub-DAO, `lastClaimedEpoch`, `expirationTs`, `claimableEpochCount` (what `claimDelegationRewards` would build) and `unissuedRequiredEpochCount` (what `undelegatePosition` rejects on). The epoch-range and issuance logic lives in a shared `claimable-epochs.ts` helper used by getPositions, claim and undelegate, so they cannot disagree. HNT-era issuance is gated on `DaoEpochInfoV0.doneIssuingRewards`, matching `claim_rewards_v1`; the old per-sub-DAO check let claims fail on-chain with `EpochNotClosed`.

Motivation: web-helium-world's undelegate/claim loop ("claim rewards first" while claim returned nothing).

## Rent from the cluster

**Program fixes (each needs a Squads release)**

- `welcome-pack`: `claim_welcome_pack_v0` refunded excess rent before the bubblegum CPI. Under new rent the refund is nonzero and `rent_refund` is usually a read-only account in that CPI, so claims fail with `UnbalancedInstruction`. The refund now runs after the CPI.
- `mini-fanout`: `distribute_v0` panicked on fanouts funded below the current rent minimum. The surplus check now uses `saturating_sub` and falls through to stop scheduling.

**Constants replaced.** `RENT_COSTS`, `MIN_WALLET_RENT_LAMPORTS`, `BASE_AUTOMATION_RENT`, `AUTOMATION_BOT_FEE`, `DELEGATION_FEE`, `RECIPIENT_RENT`, `ATA_RENT` and friends across blockchain-api, distributor-oracle, automation-hooks and voter-stake-registry-hooks are gone. Byte sizes now live in the SDK of the program that allocates the account (`lazy-distributor-sdk`, `helium-sub-daos-sdk`, `hpl-crons-sdk`) and are priced at call time through a one-hour TTL-cached `getRentLamports`. Mini-fanout task accounts are sized from tuktuk's `queue_task_v0` allocation.

[...]

PR #1299, opened 2026-09-09
feat: governance claim state in getPositions, cluster-priced rent after the mainnet Rent change, and Agave 4.2 tx v1 support

bryzettler:
Supersedes #1298 (same commits, retitled to match the full scope).

## Summary

Three pieces of work, layered in this order:

1. **Governance claim logic in blockchain-api.** `governance.getPositions` exposes per-position delegation claim state, and HNT-era claim eligibility is gated on the same flag `claim_rewards_v1` checks.
2. **Cluster-priced rent after the mainnet Rent change.** Every hardcoded rent amount is replaced with a byte size priced from the cluster's Rent sysvar (mainnet moved from 3480 lamports/byte-year at 2.0x to 6333 at 1.0x, ~9% cheaper). Two program fixes need releases.
3. **Agave 4.2 transaction v1 readiness.** `@solana/web3.js` 1.99.0 and `maxSupportedTransactionVersion: 1` on every transaction and block fetch across the repo.

## 1. Governance claim logic

Each position from `getPositions` gets a `delegation` object (`null` when not delegated): sub-DAO, `lastClaimedEpoch`, `expirationTs`, `claimableEpochCount` (what `claimDelegationRewards` would build), `requiredUnclaimedEpochCount` (what `close_delegation_v0` requires) and `unissuedRequiredEpochCount` (what `undelegatePosition` is still waiting on). The epoch-range and issuance logic lives in a shared `claimable-epochs.ts` helper used by getPositions, claim and undelegate, so they cannot disagree. The range is capped at the current epoch so a cliff that ended earlier today never includes it.

HNT-era issuance is now gated on `DaoEpochInfoV0.doneIssuingRewards`, matching `claim_rewards_v1`. The old per-sub-DAO `rewardsIssuedAt` check flips before the last sub-DAO issues, so claims built in that window failed on-chain with `EpochNotClosed`.

getPositions fetches claim epoch infos once across batches, reads the clock in parallel, and caches IDLs.

Motivation: web-helium-world's undelegate/claim loop ("claim rewards first" while claim returned nothing).

## 2. Rent from the cluster

**Program fixes (each needs a Squads release)**

- `welcome-pack` 0.0.6: `claim_welcome_pack_v0` refunded excess rent before the bubblegum CPI. Under new rent the refund is nonzero and `rent_refund` is usually a read-only account in that CPI, so claims fail with `UnbalancedInstruction`. The refund now runs after the CPI.
- `mini-fanout` 0.1.7: `distribute_v0` panicked on fanouts funded below the current rent minimum. The surplus check now uses `saturating_sub` and falls through to stop scheduling.

**Constants replaced.** `RENT_COSTS`, `MIN_WALLET_RENT_LAMPORTS`, `BASE_AUTOMATION_RENT`, `AUTOMATION_BOT_FEE`, `DELEGATION_FEE`, `RECIPIENT_RENT`, `ATA_RENT` and friends across blockchain-api, distributor-oracle, automation-hooks and voter-stake-registry-hooks are gone. Byte sizes now live in the SDK of the program that allocates the account (`lazy-distributor-sdk`, `helium-sub-daos-sdk`, `hpl-crons-sdk`) and are priced at call time through a one-hour TTL-cached `getRentLamports`. Mini-fanout task accounts are sized from tuktuk's `queue_task_v0` allocation, verified against tasks queued on a surfpool fork.

**Quotes and gates now match what the transactions charge.** Mini-fanout counts the HNT ATA and both queued tasks' `min_crank_reward`. Welcome pack costs `max(pack rent, gift + fanout escrow)` and skips `UserWelcomePacksV0` rent when it exists. The cost math is shared between `estimateCreationCost` and the create procedures. `createPosition` quotes `DelegatedPositionV0` rent and prices ATAs from the cluster. A failed rent lookup blocks the quote instead of quoting zero.

**Hooks.** `usePositionsFees` charges `DelegatedPositionV0` rent per position that still needs one. Both it and `useAutomateHotspotClaims` expose `loading` and treat unloaded rent as loading, so an empty wallet or missing provider cannot pass the balance gate.

**Test harness.** Surfpool seeds the old Rent sysvar, so `ensureSurfpool` copies the real one in after boot. The welcome-pack e2e funds its mainnet test pack to the exact claim cost until the program fix ships (marked TODO). The createPosition preflight test reads the charged fee from balance deltas, since surfpool omits the priority fee from `meta.fee`.

Not changed: data-only onboarding `new_tree_fee_lamports` is on-chain state and over-collects ~0.10 SOL per tree under new rent; lowering it is an `update_data_only_v0` call.

## 3. Agave 4.2 transaction v1

Agave 4.2 introduces version 1 transactions. `@solana/web3.js` 1.99.0 adds `MessageV1` decode, and every `getTransaction`, `getTransactions`, `getParsedTransaction(s)`, `getBlock` and Helius `getTransactionsForAddress` call now passes `maxSupportedTransactionVersion: 1`. Touched: spl-utils, blockchain-api, account-postgres-sink-service, asset-ownership-service, recent-helium-transactions-service, migration-service, helium-admin-cli, and the hpl-crons test. The sink service's integrity check derives its 24h lookback from block time instead of a hardcoded 400ms slot.

## Also in the branch

- `ecc-sig-verifier` docker build accepts the expired bullseye apt index.
- `voter-stake-registry-hooks` avoids a bigint literal that the ES2016 CJS build rejects (the CI failure on #1298).

## Testing

- blockchain-api: `tsc`, lint, `test:unit` (293 passing). `tsc -b` clean for every touched SDK and hooks package.
- `cargo test -p mini-fanout` 5 passing; mini-fanout bankrun 11 passing.
- e2e on a surfpool mainnet fork: estimate-creation-cost 3/3, reward-contract 19/19, welcome-pack 3/3, automation 13 passing, governance 41/41.
- Mini-fanout e2e can flake with `Allocate: account ... already in use` when live mainnet takes a task id the fork thought free. Environment, not code.


madninja (approved):

utils/ecc-sig-verifier/Dockerfile
```diff
@@ -16,7 +16,9 @@ RUN --mount=type=cache,id=cargo-registry-1.68,target=/usr/local/cargo/registry \
     cp /app/target/release/ecc-sig-verifier /app/ecc-sig-verifier
 
 FROM debian:bullseye-slim
-RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
+# bullseye is EOL and its security repo index has expired. The builder's
+# libssl1.1 keeps the runtime on bullseye, so accept the expired index.
+RUN apt-get -o Acquire::Check-Valid-Until=false update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
```
madninja: **Blocker: this does not fix the CI failure.** I reproduced both directions locally with `docker build --platform linux/amd64`.

On `develop`'s Dockerfile: `E: Release file for .../bullseye-security/InRelease is expired (invalid since 1d 18h)`. So the diagnosis is right.

With this line: apt gets past the index and then fails on the pool. From the failing job's log:

```
E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/o/openssl/openssl_1.1.1w-0+deb11u8_amd64.deb  404  Not Found
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

bullseye is being moved to `archive.debian.org`, and `Acquire::Check-Valid-Until=false` does nothing about a removed pool. My local build of this exact Dockerfile passed, so which mirror edge you land on decides the outcome. That makes it a flake generator rather than a fix.

Cheapest durable fix is to drop apt from the runtime stage entirely, since the `rust:1.68.2` builder is bullseye and already ships the certs:

```dockerfile
FROM debian:bullseye-slim
COPY --from=builder /etc/ssl/certs /etc/ssl/certs
```

Repointing `sources.list` at `archive.debian.org` also works, but keeps the network dependency for something the builder already has.

packages/spl-utils/src/transaction.ts
```diff
@@ -417,7 +417,7 @@ export async function sendAndConfirmWithRetry(
     if (confirmation.err) {
       const tx = await connection.getTransaction(txid, {
         commitment: "confirmed",
-        maxSupportedTransactionVersion: 0,
+        maxSupportedTransactionVersion: 1,
```
madninja, line 420: `@helium/spl-utils` publishes with `"@solana/web3.js": "^1.91.1"` (package.json:44), but these four call sites now pass `maxSupportedTransactionVersion: 1`. Only 1.99.0 accepts `1` in its response validator; 1.98.x and earlier validate the returned `version` against `0 | "legacy"` and throw on anything else.

The root `pnpm.overrides` pin to `1.99.0` only covers this monorepo. A consumer that already has 1.98.x hoisted dedupes to it, and the first genuinely-v1 transaction throws out of the send path, not just a logging path.

Raise the floor to `^1.99.0` in `packages/spl-utils/package.json` and `packages/helium-admin-cli/package.json`, which is the other published package whose code now passes `1`.

packages/voter-stake-registry-hooks/src/hooks/usePositionFees.ts
```diff
@@ -41,29 +43,54 @@ export const usePositionsFees = ({
...
+    rent,
+  ]);
 
   const prepaidTxFees = automationEnabled
     ? (numPositions - numDelegationClaimBots) * PREPAID_TX_FEES
     : 0;
   const totalFees = rentFee + prepaidTxFees;
 
+  // A failed rent lookup must block rather than quote 0 rent.
   const insufficientBalance =
```
madninja, line 92: When `provider?.connection` is falsy the async returns `undefined` rather than pending or throwing, so `react-async-hook` settles to `{loading: false, result: undefined, error: undefined}`. Then `rent?.bot ?? 0` and `rent?.delegatedPosition ?? 0` make `rentFee` exactly 0, and with `automationEnabled` false `totalFees` is 0, so this comparison is `userLamports < 0` and `insufficientBalance` is false.

Net effect: the UI shows "Rent: 0 SOL" with the button enabled, while not reporting loading. `HeliumVsrStateProvider` only builds `provider` once both connection and wallet are set, so any render before the adapter finishes connecting hits this.

`useAutomateHotspotClaims` got a `rentPending = !rent && !rentError` guard for exactly this hazard in this same PR; this hook did not. `loadingSol` is also not folded into the returned `loading`.

packages/voter-stake-registry-hooks/src/hooks/useCreatePosition.ts
```diff
@@ -53,9 +53,14 @@ export const useCreatePosition = ({
   );
   const queryClient = useQueryClient();
 
-  const { rentFee, prepaidTxFees, insufficientBalance } = usePositionFees({
+  const {
+    rentFee,
```
madninja, line 65: `isDelegated: false` maps to `numDelegatedPositions: 0, numPositions: 1`, and the new formula `(numPositions - numDelegatedPositions) * rent.delegatedPosition` therefore always adds one `DelegatedPositionV0` rent (~0.0026 SOL).

But `delegateV0` is only emitted inside `if (subDao)` at line 164. A plain lock-up with no subDao is quoted rent it will never pay, and a wallet holding between `prepaidTxFees` and `prepaidTxFees + 0.0026` is blocked from an operation it can afford.

The old expression evaluated to 0 in both branches, so this is new here. The formula is correct for `useDelegatePositions`; it is the `isDelegated: false` caller that needs the subDao condition.

packages/account-postgres-sink-service/src/utils/integrityCheckProgramAccounts.ts
```diff
@@ -72,37 +72,18 @@ export const integrityCheckProgramAccounts = async ({
...
-    ) {
-      blockTime24HoursAgo = await getBlockTimeWithRetry({
-        slot: attemptSlot,
-        provider,
-      });
-
-      if (blockTime24HoursAgo) {
-        break;
-      }
+    const snapshotBlockTime = await getBlockTimeWithRetry({
```
madninja: Replacing the 10-candidate loop with a single probe makes the whole integrity check fail on one unavailable slot.

`getBlockTimeWithRetry` throws `"Max retries reached. Unable to fetch block time."` on exhaustion, so the `if (!snapshotBlockTime)` guard below only catches a null return; the throw propagates and kills `performIntegrityCheck` for that program for the cycle. Its skipped-slot recovery only walks forward (`slot + 1`) and only when the error text contains both "slot" and "was skipped", so a lagging replica answering something else is not recovered from.

Probing the finalized tip is also the worst case for the forward walk, since `slot + 1` may not exist yet. The old code probed a slot roughly 24h old, where walking forward always had somewhere to go.

Deriving the lookback from chain time instead of a 400ms slot count is the right change; it is the loss of retry breadth that is the regression.

packages/account-postgres-sink-service/src/utils/integrityCheckProgramAccounts.ts
```diff
@@ -72,37 +72,18 @@ export const integrityCheckProgramAccounts = async ({
...
+    if (!snapshotBlockTime) {
+      throw new Error(`Unable to get blocktime for slot ${snapshotSlot}`);
     }
 
-    if (!blockTime24HoursAgo) {
-      throw new Error("Unable to get any blocktime in the last 24 hours");
-    }
+    // Derive the lookback from chain time rather than a slot count; slot
+    // duration is no longer a fixed 400ms on mainnet.
+    const blockTime24HoursAgo = snapshotBlockTime - 24 * 60 * 60;
```
madninja: Window length now equals the schedule length. The example config runs `"10 0 * * *"`, so consecutive runs have exactly zero overlap: a late start, a restart, or chain time drifting relative to wall clock leaves a band of transactions that no run ever examines, which is the one thing this check exists to prevent.

The old slot-count form carried whatever slack real slot duration gave it, so the overlap was accidental rather than designed. Worth making it explicit: subtract 25 or 26 hours.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/claimable-epochs.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  const closeRequiresThroughEpoch = Math.min(
+    isDecayed && isCliff ? decayedEpoch - 1 : currentEpoch - 1,
+    expirationCap - 1,
+  );
+
+  const lastClaimedEpoch = delegatedPosition.lastClaimedEpoch.toNumber();
+  const startEpoch = lastClaimedEpoch + 1;
+  const bitmapWindowEnd = lastClaimedEpoch + 129;
+  const rawEndEpoch = Math.min(
```
madninja, line 101: `closeRequiresThroughEpoch` above gates on `isDecayed && isCliff`, but `rawEndEpoch` here gates on `isDecayed` alone. For a decayed non-cliff lockup (kind `none`) the two disagree: `rawEndEpoch` stops at `decayedEpoch + 1` while close still requires claims through `currentEpoch - 1`.

The epochs in `(decayedEpoch, currentEpoch - 1]` are then never enumerated into `unclaimedEpochs`, so `requiredUnclaimedEpochCount` reads 0, no claim is built, and `undelegate` builds a `close_delegation_v0` whose `assert!(last_claimed_epoch >= to_claim_to_epoch)` fails on chain.

The asymmetry is pre-existing, but this PR makes this function the single source of truth behind a new public API field, so it is the moment to close it.

Separately, `isDecayed` uses `lockup.endTs.lte(unixNow)` where `to_claim_to_epoch` uses `lockup_end_ts < curr_ts`. One second wide, but they should match.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/claimable-epochs.ts
```diff
@@ -0,0 +1,274 @@
...
+   */
+  requiredUnclaimedEpochCount: number;
+  /**
+   * Unclaimed epochs close_delegation_v0 requires that are not issued yet:
+   * what undelegatePosition rejects with BAD_REQUEST.
+   */
+  unissuedRequiredEpochCount: number;
+}
+
+export const summarizeClaimableEpochs = (
```
madninja, line 263: `summarizeClaimableEpochs` iterates `range.unclaimedEpochs`, which stops at `min(rawEndEpoch, lastClaimedEpoch + 129)`. So `requiredUnclaimedEpochCount` and `unissuedRequiredEpochCount` are both capped at 128.

A delegation 400 epochs behind reports 128, not 399, while the schema describes `requiredUnclaimedEpochCount` as the count `undelegatePosition` must claim before closing. The zero versus non-zero contract still holds, which is what gating undelegate needs; the magnitude does not, which is what a progress indicator would need. Worth saying so in the schema description if the cap is intended.

programs/welcome-pack/src/instructions/claim_welcome_pack_v0.rs
```diff
@@ -272,6 +260,23 @@ pub fn handler<'info>(
     &[welcome_pack_seeds!(welcome_pack)],
   )?;
 
+  // Refund excess rent only after update_compression_destination_v0 and the
+  // bubblegum transfer. The runtime syncs only a CPI's own accounts back before
+  // the call, so a welcome_pack debit paired with a credit to a rent_refund
+  // that the CPI does not pass leaves the caller unbalanced and fails with
+  // UnbalancedInstruction.
+  let rent_refunded_amount = welcome_pack
```
madninja, line 268: The reorder itself is correct. I traced both CPIs it now sits behind, and neither `update_compression_destination_v0` nor the bubblegum transfer touches `welcome_pack`'s lamports, so the amount computed here is byte-identical to what the old position computed.

What is missing is coverage. The e2e now calls `fundPackToExactClaimCost`, whose own comment says it exists so the refund branch stays dead, and nothing else drives `rent_refunded_amount > 0`. Moving this block back above `update_compression_destination_v0`, which reintroduces the production `UnbalancedInstruction`, leaves the suite green.

The branch is not exotic: per `initialize_welcome_pack_v0.rs:197-201` a pack is topped up to `sol_amount + fanout_cost` inclusive of its own rent, so any pack whose init rent exceeded that sum, or that `resize_to_fit` topped up, claims with a nonzero refund. A test that funds above exact cost and asserts `rent_refund`'s balance rose would pin it.

programs/mini-fanout/src/instructions/distribute_v0.rs
```diff
@@ -237,7 +237,11 @@ pub fn handler<'info>(
 
   // Pay min crank reward to task_queue from mini_fanout, if available
   let min_rent_exempt = Rent::get()?.minimum_balance(mini_fanout_info.data_len());
-  if mini_fanout_info.lamports() - min_rent_exempt >= ctx.accounts.task_queue.min_crank_reward * 2 {
+  // saturating_sub: accounts funded before a Rent sysvar increase can sit
+  // below the current minimum, and a bare subtraction would panic.
+  if mini_fanout_info.lamports().saturating_sub(min_rent_exempt)
```
madninja: This changes the failure mode, not just the panic. With `overflow-checks = true` the old subtraction panicked when the fanout sat below the current rent minimum: nothing committed, the tuktuk task stayed queued, and the fanout self-healed on the next top-up.

Now it falls into the `else` branch, writes the `next_task`/`next_pre_task` sentinel and returns no tasks. Tuktuk frees the task, the fanout stops distributing permanently, and topping it up does not restart it; only an off-chain `schedule_task_v0` does.

If that is the intent, the two paths into the `else` are worth distinguishing with a `msg!`, because "below rent exempt" and "cannot afford the crank reward" want different operator responses.

Two smaller things at the same site. `min_crank_reward * 2` is still an unchecked multiply in three places immediately around the subtraction that was just hardened. And when `min_crank_reward` is 0 the guard reads `0 >= 0` and passes, so a below-rent-exempt fanout takes the pay-the-crank-reward branch under a check whose comment says "if available".

madninja (commented):
## Review

Blocking issue first: **`Test Docker Builds (utils/ecc-sig-verifier)` is failing, and the Dockerfile change in this PR does not fix it.** Detail in the inline comment. Everything else below is non-blocking.

I re-derived the byte sizes rather than reading the comments, and they check out: `size_of::<DelegatedPositionV0>()` = 176 (the u128 forces 16-byte alignment on a 170-byte sum), `PositionV0` = 176, `DelegationClaimBotV0::INIT_SPACE` = 138, `size_of::<TaskV0>()` = 216. That last one makes `miniFanoutDistTaskSpace(2)` = 645 and `miniFanoutPreTaskSpace(97)` = 453, which are exactly the two mainnet accounts the comment cites. `bitmapWindowEnd = lastClaimedEpoch + 129` matches `set_claimed`'s `epoch > last_claimed_epoch + 128` rejection, and the `doneIssuingRewards && delegationRewardsIssued > 0` gate matches `claim_rewards_v1`'s two account constraints exactly. The v0/v1 instruction choice and `isEpochInfoIssued` read the same discriminator, so they cannot disagree. `createTtlCache` does not cache rejections.

### Findings

| # | Where | Sev |
|---|---|---|
| 0 | `utils/ecc-sig-verifier/Dockerfile` | blocker |
| 1 | `spl-utils` / `helium-admin-cli` published web3.js floor | high |
| 2 | `usePositionFees` settles not-loading with 0 rent | high |
| 3 | `useCreatePosition` charges delegation rent unconditionally | med |
| 4 | integrity check single block-time probe | med |
| 5 | integrity check lookback has zero overlap | med |
| 6 | decayed non-cliff lockup: required epochs never enumerated | med |
| 7 | welcome-pack refund reorder has no coverage | med |
| 8 | `saturating_sub` turns a retryable panic into a permanent stop | med |
| 9 | unbounded PDA derivation in `fetchDelegations` | med |
| 10 | `claimableEpochCount` does not model the claim builder's caps | med |
| 11 | required-epoch counts truncated at the bitmap window | med |
| 12 | `reward-contract/create.ts` task queue read moved earlier | low |
| 13 | `getOracleCount` memoized with no TTL | low |
| 14 | `rentPending` latches on when the provider is absent | low |
| 15 | automation e2e assertion computes its expectation from the code under test | low |

Security: no auth path is touched; the only identity input is `walletPubkey`, already authenticated upstream. DRY: the shared `claimable-epochs.ts` is a genuine de-duplication. YAGNI: every new export has a real consumer outside its tests. Coverage is the weak dimension, and #7 and #15 are where.

packages/blockchain-api/src/server/api/routers/governance/procedures/positions/list.ts
```diff
@@ -19,6 +42,85 @@ const getPositionsIpRateLimiter = createRateLimiter({
...
+    return getClaimableEpochRange({
+      lockup: p.account.lockup,
+      delegatedPosition: delegation,
+      unixNow,
+    });
+  });
+
+  // Positions on the same sub-DAO share sub-DAO epoch infos, and every
+  // position shares the DAO epoch infos; read each once.
+  const entries: { subDao: PublicKey; epoch: number }[] = [];
```
madninja, line 99: `entries` gets one element per (position, unclaimed epoch) pair with no cap, and each position contributes up to 128 of them.

`fetchEpochInfos` dedupes into unique keys before deriving, so the common case where positions share epochs is cheap. The bad case is many positions with disjoint claim histories: unique (subDao, epoch) pairs then scale with positions times epochs, and both `subDaoEpochInfoKey` and `daoEpochInfoKey` are `findProgramAddressSync`, which grinds bumps synchronously on the request thread. A few hundred positions with non-overlapping histories is tens of thousands of derivations plus hundreds of `getMultipleAccountsInfo` calls, on an endpoint that previously did none of this.

There is an IP rate limiter in front, which bounds the request rate but not the cost of one request. Worth capping the epochs summarised per position, since the field's consumers only need the count.

packages/blockchain-api-client/src/schemas/governance.ts
```diff
@@ -277,6 +277,40 @@ const RelinquishAllVotesMetadataSchema = z.object({
...
+    .describe(
+      "Latest epoch through which rewards are contiguously claimed; the first candidate epoch is lastClaimedEpoch + 1",
+    ),
+  expirationTs: z
+    .number()
+    .int()
+    .describe(
+      "Raw on-chain delegation expiration unix timestamp; 0 means a legacy delegation with no expiration",
+    ),
+  claimableEpochCount: z
```
madninja, line 294: This describes the field as what `claimDelegationRewards` would emit right now, but `getPositions` does not model two caps the claim builder applies.

`buildClaimInstructions` sets `hitCap` when any position's `rawEndEpoch > bitmapWindowEnd` and then `break`s out of the position loop, and it separately stops at `MAX_TXS_PER_CALL * 128 = 640` epochs. `fetchDelegations` applies neither.

Concretely: a wallet with position A holding 200 unclaimed epochs and position B holding 5 issued unclaimed epochs. `getPositions` reports 5 for B; the claim endpoint emits A's 128 and never reaches B. A UI driven off this field shows a claimable position that the claim call silently ignores. Either apply the caps here or reword the description to "in the claimable range" without the emit claim.

packages/blockchain-api/src/server/api/routers/reward-contract/procedures/create.ts
```diff
@@ -113,27 +115,82 @@ export const create = publicProcedure.rewardContract.create.handler(
...
 
+    const welcomePackProgram = hasClaimable ? await init(provider) : undefined;
+    const uwpAcc = welcomePackProgram
+      ? await welcomePackProgram.account.userWelcomePacksV0.fetchNullable(
+          userWelcomePacksKey(new PublicKey(delegateWalletAddress))[0],
+        )
+      : null;
+    const taskQueueId = TASK_QUEUE_ID;
+    const tuktukProgram = hasClaimable ? undefined : await initTuktuk(provider);
+    const taskQueueAcc =
```
madninja: Two things moved here that are worth a second look.

`taskQueueAcc` moved from just before `nextAvailableTaskIds` to ahead of the balance check and all the asset and instruction building, which widens the window in which another caller takes the task id the bitmap said was free. The PR description reports exactly that flake in the mini-fanout e2e.

Separately, `taskQueueId` went from `new PublicKey(process.env.HPL_CRONS_TASK_QUEUE!)`, which throws when the env is unset, to `TASK_QUEUE_ID`, which falls back to the hardcoded mainnet queue. On devnet with the env unset this now builds against mainnet's queue instead of failing loudly. Every other call site already uses the fallback constant, so this is arguably alignment rather than a regression; flagging it so it is a decision rather than a side effect.

packages/distributor-oracle/src/server.ts
```diff
@@ -133,6 +136,30 @@ export class OracleServer {
...
+    const value =
+      this.ldProgram.provider.connection.getMinimumBalanceForRentExemption(
+        space
+      );
+    this.rentCache.set(space, { value, expiresAt: now + 60 * 60 * 1000 });
+    value.catch(() => this.rentCache.delete(space));
+    return value;
+  }
+
+  private getOracleCount(): Promise<number> {
```
madninja, line 161: `oracleCount` is memoized for the process lifetime and only cleared on rejection, while `minRent` immediately above carries a deliberate 1 hour TTL.

Add an oracle to the lazy distributor and `recipientSpace(n)` stays 8 bytes short in every running oracle process until it restarts, so `neededBalance` under-quotes and a claim that should have been refused builds a transaction that fails on chain. Cheap fix is to route it through the same TTL cache as the rent lookups.

packages/automation-hooks/src/hooks/useAutomateHotspotClaims.ts
```diff
@@ -191,17 +195,33 @@ export const useAutomateHotspotClaims = ({
...
+      )
+    );
+    return {
+      walletMin,
+      recipient,
+      ata: ataRent,
+      baseAutomation: base.reduce((sum, lamports) => sum + lamports, 0),
+    };
+  }, [provider?.connection]);
+  const rentPending = !rent && !rentError;
```
madninja: When `provider?.connection` is falsy the async resolves `undefined`, so `rent` and `rentError` are both undefined and `rentPending` latches true and never clears.

That makes `loading` permanently true, and it hardwires `insufficientSol` to false, because `!rentPending` gates the comparison. A consumer gating on `insufficientSol` alone reads "sufficient SOL" while `rentFee`, `recipientFee` and `totalSolNeeded` are all quoted with rent at 0.

Same root cause as the `usePositionFees` comment, opposite symptom: there the hook settles not-loading with 0 rent, here it never settles at all. Both come from returning `undefined` for "no connection yet" instead of keeping the hook pending.

packages/blockchain-api/tests/e2e/automation.test.ts
```diff
@@ -95,7 +96,8 @@ describe("automation endpoints", () => {
       expect(result.isOutOfSol).to.equal(false);
       expect(result.currentSchedule).to.be.undefined;
       expect(result.rentFee).to.equal(
-        BASE_AUTOMATION_RENT + TASK_RETURN_ACCOUNT_SIZE
+        (await getBaseAutomationRentLamports(connection)) / LAMPORTS_PER_SOL +
```
madninja, line 120: This assertion computes its expectation by calling the same helper the handler calls, so it can only ever agree with the implementation. A wrong byte size in `entityClaimCronSpaces`, which is exactly what this PR introduces, stays green.

The previous form pinned a literal, which is why it would have caught a size change. Better still is to assert against the account sizes read back from the chain after the automation is created, so the test measures the artifact rather than re-running the model.

utils/ecc-sig-verifier/Dockerfile
madninja: Correction on the status, not the substance: the job was re-run after I posted and it passed, so CI on this PR is now fully green. Downgrading this from a blocker to a note.

What is still true is that the pool fetch is mirror dependent. The same Dockerfile 404ed on one Fastly edge and succeeded on another within the same hour, so this will keep failing intermittently as bullseye finishes moving to `archive.debian.org`. The `COPY --from=builder /etc/ssl/certs /etc/ssl/certs` form removes the network dependency entirely and is worth taking whenever this file is next touched.

bryzettler: Agreed. Dropped apt from the runtime stage and copied `/etc/ssl/certs` from the bullseye-based builder, so the runtime stage no longer depends on the bullseye mirrors at all.

packages/spl-utils/src/transaction.ts
bryzettler: Bumped `@solana/web3.js` to `^1.99.0` in `spl-utils` and `helium-admin-cli` package.json. The root `pnpm.overrides` already pins 1.99.0, so the lockfile is unchanged; the published range now matches what the code needs.

packages/voter-stake-registry-hooks/src/hooks/usePositionFees.ts
bryzettler: Fixed. `loading` is now `loadingSol || (!rent && !error)`, so a missing provider is treated as pending rather than settling with rent at 0. `insufficientBalance` only compares once `loading` is false.

packages/voter-stake-registry-hooks/src/hooks/useCreatePosition.ts
bryzettler: Fixed. `useCreatePosition` takes a `delegating` prop (defaults to `automationEnabled`, since automation implies delegation) and passes `isDelegated: !delegating` to `usePositionFees`. An undelegated lock-up no longer quotes or blocks on `DelegatedPositionV0` rent.

packages/account-postgres-sink-service/src/utils/integrityCheckProgramAccounts.ts
bryzettler: Restored the multi-slot probe: it now walks up to 10 slots backwards from the finalized slot, catching per-slot failures, and only throws if none of them return a block time.

packages/account-postgres-sink-service/src/utils/integrityCheckProgramAccounts.ts
bryzettler: Lookback is now 25 hours (`snapshotBlockTime - 25 * 60 * 60`) so consecutive daily runs overlap by an hour and a late start or restart cannot leave a gap.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/claimable-epochs.ts
bryzettler: Fixed. Both `closeRequiresThroughEpoch` and `rawEndEpoch` now gate on `isDecayed && isCliff`, and `isDecayed` uses `endTs < now` to match `lockup_end_ts < curr_ts` in the program. Added unit tests for a decayed `none` lockup (must enumerate through yesterday) and a cliff ending exactly now (not yet decayed).

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/claimable-epochs.ts
bryzettler: Documented rather than changed: the 128-epoch bitmap window is a program limit (`claimed_epochs_bitmap`), so the counts cannot exceed it any more than a claim can. Updated the `ClaimableEpochSummary` doc comment and the client schema descriptions to state that every count is capped at 128 and that a position needing more must claim and re-query.

programs/welcome-pack/src/instructions/claim_welcome_pack_v0.rs
bryzettler: Added coverage in `tests/welcome-pack.ts`: a nested case initializes the pack with a distinct `rent_refund` wallet, tops the pack up above its exact claim cost, claims, and asserts `rent_refund`'s balance rose by exactly the top-up. Moving the refund back ahead of the bubblegum CPI now fails this test with `UnbalancedInstruction`.

programs/mini-fanout/src/instructions/distribute_v0.rs
bryzettler: Changed to an explicit error instead of silently converging: `require_gte!(lamports, min_rent_exempt, BelowRentExempt)` so a fanout below the current rent minimum errors and the task stays queued for a retry after a top-up, rather than being permanently unscheduled. Crank rewards use `checked_mul` and the affordability check subtracts only after the require, so no overflow panic path remains.

packages/blockchain-api/src/server/api/routers/governance/procedures/positions/list.ts
bryzettler: Left as is, with reasoning. `subDaoEpochInfoKey` depends only on `(subDao, epoch)` and `daoEpochInfoKey` only on `epoch`, and `fetchEpochInfos` dedupes on those keys before deriving. So unique derivations are bounded by sub-DAOs (2) times epochs that have ever existed under delegation (~1.1k), not by positions times epochs: about 3.3k derivations and ~30 `getMultipleAccountsInfo` batches in the worst case regardless of how many positions or how disjoint their histories are. That is the same ceiling a single position with a long history hits today. Happy to add a per-position cap if you would still rather bound it, but it would not change the worst case.

packages/blockchain-api-client/src/schemas/governance.ts
bryzettler: Reworded the schema descriptions. `claimableEpochCount` is now described as the issued, unclaimed epochs inside the position's 128-epoch bitmap window, with a note that `claimDelegationRewards` batches and caps per call, so it is not a per-call emit count. `requiredUnclaimedEpochCount` and `unissuedRequiredEpochCount` note the same 128 cap.

packages/blockchain-api/src/server/api/routers/reward-contract/procedures/create.ts
bryzettler: Narrowed the early fetch to just `minCrankReward` for the rent math, and `taskQueueAcc` is fetched again immediately before `nextAvailableTaskIds`, so the bitmap window is back to what it was. On the env fallback: kept `TASK_QUEUE_ID` deliberately for alignment with the other call sites, as you noted.

packages/distributor-oracle/src/server.ts
bryzettler: Done. Replaced the two ad hoc caches with one `cachedHourly(key, load)` helper backed by a TTL map; `minRent` and `getOracleCount` both go through it, so oracle count now refreshes hourly too.

packages/automation-hooks/src/hooks/useAutomateHotspotClaims.ts
bryzettler: Same fix as `usePositionFees`: `rentPending` is now computed from `!rent && !rentError`, which is true only until the provider connects and the quote resolves, so it does not latch. Removed the redundant `loadingRent`. `insufficientSol` stays false while pending on purpose to avoid a flash; consumers must gate on `loading`, and the hook comment now says so.

packages/blockchain-api/tests/e2e/automation.test.ts
bryzettler: Added an e2e case that reads the accounts the cron actually allocates after scheduling: `UserCronJobsV0`, `CronJobV0`, the name mapping, the schedule task and the `task_return_account_1` PDA. It asserts each account's on-chain `data.length` (or lamports, for the return account) equals the constant the quote uses, so a wrong byte size in the helper now fails against the chain rather than against itself.

It immediately caught one: `ENTITY_CLAIM_SCHEDULE_TASK_SPACE` was 738, measured from a live cron job's schedule task. tuktuk sizes a TaskV0 by its compiled transaction, and the task `initialize_cron_job_v0` queues is 800 bytes; 738 is what `queue_cron_tasks` requeues it at after the first run, paid from the cron's funding rather than by the wallet. The constant is now 800 with that documented, and the e2e passes against a mainnet fork.

packages/blockchain-api/tests/unit/claimable-epochs.test.ts
```diff
@@ -0,0 +1,286 @@
...
+    expect(r.unclaimedEpochs[0]).to.eq(endEpoch - 1);
+    expect(r.unclaimedEpochs[r.unclaimedEpochs.length - 1]).to.eq(
+      CURRENT_EPOCH - 1
+    );
+    expect(
+      summarizeClaimableEpochs(r, allIssued).requiredUnclaimedEpochCount
+    ).to.eq(r.unclaimedEpochs.length);
+  });
+
+  it("treats a cliff ending exactly now as not yet decayed, like close_delegation_v0", () => {
```
madninja: This test cannot fail on the change it is named for.

`.lte` and `.lt` only differ when `endTs === unixNow`, and in exactly that case `decayedEpoch === currentEpoch`. So with `isDecayedCliff` true you get `rawEndEpoch = min(currentEpoch, decayedEpoch + 1) = currentEpoch` and `closeRequiresThroughEpoch = decayedEpoch - 1 = currentEpoch - 1`; with it false you get `rawEndEpoch = currentEpoch` and `closeRequiresThroughEpoch = currentEpoch - 1`. Identical. I worked this fixture through both branches and every assertion here matches either way, so reverting `.lt` to `.lte` leaves it green.

Which also means the `.lt` change is a no-op in every reachable state. That is fine as an alignment with `close_delegation_v0.rs` and worth keeping, but the test is asserting the general boundary rather than the thing its name claims. Either rename it to what it actually pins, or drop it, since the case is already covered by the `UNIX_NOW - 1` test above.

packages/blockchain-api/tests/e2e/automation.test.ts
```diff
@@ -199,6 +222,42 @@ describe("automation endpoints", () => {
...
+          cronJob,
+          cronJobNameMappingKey(authority, ENTITY_CLAIM_CRON_NAME)[0],
+          cronJobAcc.nextScheduleTask,
+          taskReturnAccount1,
+        ]);
+
+      expect(userCronJobs?.data.length).to.equal(USER_CRON_JOBS_SPACE);
+      expect(cronJobInfo?.data.length).to.equal(cronJobSpace(DAILY_CRON.length));
+      expect(nameMapping?.data.length).to.equal(CRON_JOB_NAME_MAPPING_SPACE);
+      expect(scheduleTask?.data.length).to.equal(
```
madninja, line 259: This is the right shape, measuring the accounts rather than re-running the model, and it is what makes the 738 to 800 correction stick.

Two things worth writing into the test. It has an implicit precondition: per the SDK comment next to `ENTITY_CLAIM_SCHEDULE_TASK_SPACE`, `nextScheduleTask` is only 800 bytes before the cron has fired, after which `queue_cron_tasks` requeues it at 738. True on a fork with a daily schedule, but the failure mode if that ever shifts is someone "correcting" 800 back to 738 and quietly restoring the under-quote, so the reason belongs here and not only in the SDK.

Second, `cronJobSpace(DAILY_CRON.length)` pins the formula at a schedule length of 11 while the quote itself uses `MAX_PRESET_SCHEDULE_LEN` of 15. The formula is the risky part so this is the right thing to assert, but it does mean the specific value the quote uses is not covered. Worth a line saying 15 is a deliberate upper bound.

madninja (commented):
## Re-review of the fixes

Went through all 11 fix commits. Nine findings are properly fixed, three were answered by correcting the description rather than the behavior (fine), and one is untouched. Resolving the threads that are settled and replying on the rest.

Two of the fixes are better than what I asked for. The block-time probe now walks *backwards* from the finalized tip, which is the right direction since every slot behind the tip already exists, and the supporting one-liner that stops the recursive retry resetting `maxSlotIncrement` to its default is a real bug I did not spot. And `distribute_v0` failing with a named `BelowRentExempt` beats the `msg!` I suggested.

I verified the Dockerfile change by building the image rather than trusting a green build, since a build passing says nothing about whether the binary runs. Detail on that thread.

Five new things from the fixes, three of them inline below and two as replies on the threads they belong to. The one that matters is the `delegating` default on `useCreatePosition`.

**CI: `governance` is red, and it is not these fixes.** It passed at `35e7e30` and fails at `e5fb794`, and the whole delta between those two commits is one line in `packages/account-postgres-sink-service/src/utils/getBlockTimeWithRetry.ts`, a package the governance e2e never loads. The panic is `extend_expiration_ts_v0.rs:188`, `checked_sub(...).unwrap()` on `vehnt_in_closing_positions`, in a program this PR does not touch, against sub-DAO epoch-info state pulled from a live mainnet fork. Fork drift. I have re-run the job to confirm.

Worth splitting out separately: that `.unwrap()` is reachable from real mainnet state, and a panic is a poor way to find that out.

packages/voter-stake-registry-hooks/src/hooks/useCreatePosition.ts
madninja: Fixed for the delegating case, but the default flips the error to the unsafe direction.

`isDelegated: !delegating` resolves correctly, since `isDelegated` means "already has a `DelegatedPositionV0`". The issue is `delegating = automationEnabled`. Before, an undelegated create was over-quoted by ~0.0026 SOL, which is annoying but safe. Now a caller that creates a delegated position *without* automation, and does not pass `delegating`, is under-quoted, and the transaction fails on chain for insufficient funds.

`useCreatePosition` has no consumers in this repo, so every existing caller picks up the new default silently. Defaulting `delegating` to `true` keeps the fix and puts the error back on the over-quote side; a caller that knows it will not delegate opts out explicitly.

Leaving this open for that.

programs/mini-fanout/src/instructions/distribute_v0.rs
madninja: This is a better fix than what I suggested. The named `BelowRentExempt` beats a `msg!`, `checked_mul` closes the unchecked multiply, and appending the variant last keeps every existing error code stable.

I checked the claim in the comment rather than taking it, since "the task stays queued and the fanout resumes on its own" is the kind of guarantee that is easy to assert and easy to be wrong about. It holds: `run_task_v0` carries `close = rent_refund` on the task account, so a failing CPI reverts the close along with everything else and the task is not consumed.

One thing left. `BelowRentExempt` has no test anywhere in the tree, so reverting `require_gte!` back to the `saturating_sub` form leaves the suite green. The bankrun suite already has what this needs: `setAccount` the fanout below the rent minimum and assert the error. That is the one guard these fixes created with nothing behind it.

Leaving this open for the test.

packages/automation-hooks/src/hooks/useAutomateHotspotClaims.ts
madninja: The comment is a fair answer for the missing-provider case, and telling consumers to gate on `loading` as well as `insufficientSol` is the honest framing.

Dropping `loadingRent` from `loading` was not free, though. `rentPending` covers the first load, but not a refetch: when `provider.connection` changes, `rent` still holds the previous value, so `rentPending` is false and the hook reports not-loading while the new quote is in flight. `loadingRent` covered exactly that. Rent moves about once a year so the exposure is small, but if the intent was to drop a redundant term it is not quite redundant.

Leaving this open as a judgement call rather than a defect.

packages/blockchain-api/src/server/api/routers/governance/procedures/positions/list.ts
madninja: Still open, and no change here in the fix commits. Flagging so it is a decision rather than an oversight: either cap the epochs summarised per position, or say explicitly that the IP rate limiter in front is considered sufficient given how few wallets hold enough positions with disjoint histories to make this expensive.

utils/ecc-sig-verifier/Dockerfile
madninja: Confirmed fixed, and I checked it by building the image rather than trusting a green build, since the build passing says nothing about whether the binary runs. Dropping the apt install also drops the `openssl` package it was pulling in, so the thing to rule out was a runtime `libssl.so.1.1` failure that CI would never see.

`ldd` on the copied binary shows only `libgcc_s`, `libpthread`, `libm`, `libdl` and `libc`. No OpenSSL at all, so `reqwest` is dead weight in a 221-line `main.rs` and the CA bundle was never load-bearing. `ca-certificates.crt` lands as a real 201KB file and the hashed symlinks are relative to `/etc/ssl/certs`, so they resolve rather than dangle.

One latent note, not actionable now: `/usr/lib/ssl`, Debian's `OPENSSLDIR`, does not exist in `bullseye-slim`, so anything added later that does need OpenSSL verification would want `SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt`.

packages/spl-utils/src/transaction.ts
madninja: Fixed. Both `spl-utils` and `helium-admin-cli` now floor at `^1.99.0`, so a consumer cannot resolve a version that rejects `1` in its response validator.

packages/voter-stake-registry-hooks/src/hooks/usePositionFees.ts
madninja: Fixed. `loading = loadingSol || (!rent && !error)` and `insufficientBalance` now gates on `loading` rather than `loadingSol` alone, so the missing-provider case reports loading instead of quoting zero rent and passing the gate.

packages/account-postgres-sink-service/src/utils/integrityCheckProgramAccounts.ts
madninja: Fixed, and better than what I suggested. Walking backwards from the finalized tip is the right direction, since every slot behind the tip already exists, where forward from the tip may not yet.

The supporting one-liner in `getBlockTimeWithRetry` is a real bug I did not spot: the recursive retry dropped `maxSlotIncrement`, so it silently reset to the default on every retry. Passing `maxSlotIncrement: 0` from here would not have stuck without it.

packages/account-postgres-sink-service/src/utils/integrityCheckProgramAccounts.ts
madninja: Fixed. 25h against a daily schedule gives an hour of overlap, so a late start or clock drift cannot leave a band nothing examines.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/claimable-epochs.ts
madninja: Fixed. `isDecayedCliff` now gates both `closeRequiresThroughEpoch` and `rawEndEpoch`, so a decayed non-cliff lockup keeps enumerating the epochs close still requires.

I checked the new unit test actually catches the regression rather than just passing: with the old `isDecayed` the `none` fixture gives `rawEndEpoch = CURRENT_EPOCH - 29` against an assertion of `CURRENT_EPOCH`, so it goes red. That is real coverage.

The separate `.lte` to `.lt` change and the test named for it are a different matter; comment on that test.

packages/blockchain-api/src/server/api/routers/governance/procedures/helpers/claimable-epochs.ts
madninja: Addressed by documenting the cap on both the interface and the schema rather than changing the behavior, which is the right call: zero versus non-zero is what gates undelegate and that is exact. Anyone wanting a progress count now knows what they are reading.

programs/welcome-pack/src/instructions/claim_welcome_pack_v0.rs
madninja: Fixed with a test that genuinely pins it. Asserting `rent_refund` rose by exactly the top-up means both removing the refund and moving it back ahead of the bubblegum CPI go red, which is what this needed.

The mocha ordering works out too: the nested suite's `before` sets `rentRefund` before the outer `beforeEach` builds the pack, and the outer suite's own tests run before the nested suite, so the sibling tests keep `PublicKey.default`.

packages/blockchain-api-client/src/schemas/governance.ts
madninja: Addressed. Dropping the "would emit a claim instruction" phrasing and naming the batching and per-call cap removes the contradiction. Documenting rather than applying the caps here is the right trade, since applying them would make the field depend on the other positions in the wallet.

packages/blockchain-api/src/server/api/routers/reward-contract/procedures/create.ts
madninja: Fixed. `minCrankReward` early, task bitmap re-read immediately before `nextAvailableTaskIds`, so the race window is back to what it was.

I checked both non-null assertions land where the value exists: `minCrankReward!` is inside the `else` branch and the late `tuktukProgram!` fetch is inside the mini-fanout path, both of which only run when `hasClaimable` is false.

The `TASK_QUEUE_ID` half was left as the fallback constant, which matches every other call site in the repo. Fine by me; noting it so it reads as a decision.

packages/distributor-oracle/src/server.ts
madninja: Fixed. One `cachedHourly` behind both, so the oracle count can no longer outlive a change to the oracle set while the rent next to it refreshes.

packages/blockchain-api/tests/e2e/automation.test.ts
madninja: Addressed the right way, by adding a test that measures the on-chain account lengths against the SDK constants instead of re-running the model. That is what makes the 738 to 800 correction stick, and it is why the wrong byte size surfaced at all.

The original assertion here is now redundant rather than load-bearing, so I am not going to push on deleting it. Two smaller notes on the new test in a separate comment.

madninja:
Correcting my call on the `governance` failure: I said fork drift and to re-run. The re-run failed identically (attempt 3, same `extend_expiration_ts_v0.rs:188` panic), so "transient, re-run it" was wrong.

The real mechanism is time of day, and `048c850` has it right. `EPOCH_LENGTH` is 86400, so an epoch is a UTC day, and `clockTimestamp + 3600` crosses into the next epoch whenever the job runs after 23:00 UTC. That makes the rewritten expiration name a different sub-DAO epoch info than the one delegation recorded the correction on, and the `checked_sub` underflows. The run timestamps line up exactly: green at 15:21, 19:40 and 20:18 UTC, red at 23:34 and again at 23:46 on the re-run.

What I had right is only the negative half: the delta between the last green commit and the red one was one line in `account-postgres-sink-service`, so the fix commits could not have caused it. I over-read that into "environmental, therefore transient", when it was a latent test bug that any run in the last hour of a UTC day would have hit.

Two things still worth considering, neither blocking:

The cap narrows the window rather than closing it. `Math.min(clockTimestamp + 3600, epochStart + EPOCH_LENGTH - 1)` shrinks the extension as the clock approaches midnight, and in the final seconds of a UTC day `shortExpiration` collapses to roughly `clockTimestamp`, which is not much of an extension to be asserting on. Surfpool can set the clock, so pinning it for this test would take the time dependence out entirely instead of moving it from the last hour to the last few seconds.

And the program is still the thing that panics. A test now steers around `extend_expiration_ts_v0` subtracting a correction from an epoch info that never recorded one, but the `.unwrap()` on that `checked_sub` is still there. The test reached the state by writing `expiration_ts` directly, which is not a route mainnet has, so this is not evidence of a live bug. It is evidence that the failure mode is a panic rather than a returned error, which is a poor way to find out. Worth its own issue rather than anything in this PR.

packages/voter-stake-registry-hooks/src/hooks/useCreatePosition.ts
madninja: Fixed. `delegating = true` puts the error back on the over-quote side, and the docstring now says why, so a caller that locks without delegating opts out explicitly rather than being the silent default.

programs/mini-fanout/src/instructions/distribute_v0.rs
madninja: Fixed, and this is the strongest of the new tests. It pins the error code and number (I checked 6013 against the enum ordering, `BelowRentExempt` is the 14th variant), asserts nothing committed, asserts the task is still queued rather than consumed, and then tops up and asserts the same task runs and pays out.

That last assertion is the one that matters: it pins the "resumes on its own once it is topped up" claim by demonstrating it, rather than restating it in a comment. Reverting `require_gte!` to the `saturating_sub` form now fails this.

packages/blockchain-api/tests/unit/claimable-epochs.test.ts
madninja: Deleting it is the right call. The `UNIX_NOW - 1` case above already covers the boundary, and the `.lt` alignment with `close_delegation_v0.rs` stands on its own without a test that could not have failed on it.

packages/automation-hooks/src/hooks/useAutomateHotspotClaims.ts
madninja: Fixed. `rentPending = loadingRent || (!rent && !rentError)` closes the refetch window.

Worth noting you also applied the same gate to `usePositionsFees`, which had the identical latent gap. I had only flagged it here.

packages/blockchain-api/tests/e2e/automation.test.ts
madninja: Both preconditions are now written down, and the schedule-task one explains the trap rather than just the fact: measuring after the cron has fired would "correct" the SDK constant from 800 down to 738 and quietly restore the under-quote this test exists to catch. That is the note that keeps the test from being deleted by someone who thinks it is wrong.

packages/blockchain-api/src/server/api/routers/governance/procedures/positions/list.ts
madninja: Documented rather than capped, and the reasoning is better than my finding. The dedupe bounds unique `(subDao, epoch)` pairs by sub-DAOs times epochs-since-delegation regardless of position count, so it does not scale with positions the way I said. My "hundreds of positions, tens of thousands of derivations" was wrong on that point. A few thousand derivations behind a rate limiter is a fine place to leave it.

packages/blockchain-api/src/server/api/routers/reward-contract/procedures/estimateCreationCost.ts
```diff
@@ -11,66 +12,123 @@ import {
...
         }
+        // With more than one recipient, initialize_welcome_pack_v0 escrows the
+        // future fanout's rent, its HNT ATA rent and FANOUT_FUNDING_AMOUNT
+        // alongside the gift.
+        const hasFanout = recipients.length > 1;
+        const { packCost } = getWelcomePackCost({
+          welcomePackRent,
+          fanoutRent,
+          ataRent,
+          giftLamports: recipientGift.toNumber(),
```
madninja, line 100: Non-blocking, noting it rather than asking for a change.

Sharing the cost rule is right and I checked it is value-identical to the three copies it replaced. It does move this site from all-BN to `number` arithmetic, though, so `recipientGift.toNumber()` here and `new BN(packCost)` below now throw a `RangeError` above 2^53 lamports where the BN path did not. That is roughly 9M SOL of gift, so it is not a real input, but it turns what would have been a clean `INSUFFICIENT_FUNDS` into a 500.

Aligning with the rest of `balance-validation`, which already works in `number` lamports, is the right direction. If you ever want the edge closed, clamping in `resolveTokenAmountInput` is the place, not here.

madninja (approved):
## Approving

All 15 findings from the first pass and all five from the re-review are addressed, and CI is green on `900079f` (69 checks passing, only the three deliberately-skipped deploy jobs). Resolved every thread.

Three of the fixes are better than what I asked for:

The `BelowRentExempt` bankrun test does not just assert the error. It asserts nothing committed, asserts the task is still queued rather than consumed, then tops the fanout up and asserts the same task runs and pays out. That last step demonstrates the "resumes on its own" property instead of restating it in a comment, which is the difference between a test and a claim.

The block-time probe walking backwards from the finalized tip is the right direction rather than the one I suggested, and the `maxSlotIncrement` propagation bug it depends on was a real defect I had not spotted.

And on `list.ts`, the reasoning in the comment is better than my finding: the dedupe bounds unique `(subDao, epoch)` pairs by sub-DAOs times epochs-since-delegation regardless of position count, so it does not scale with positions the way I claimed.

I also re-derived the byte sizes independently rather than reading the comments: `size_of::<DelegatedPositionV0>()` and `PositionV0` at 176, `DelegationClaimBotV0::INIT_SPACE` at 138, `size_of::<TaskV0>()` at 216, which reproduces the 645- and 453-byte mainnet tasks the comments cite. The `welcomePackSpace` and `miniFanoutSpace` formulas match the programs field for field, and the `getWelcomePackCost` extraction in `900079f` is value-identical at all three call sites.

One non-blocking note inline on `estimateCreationCost.ts`.

For the record, two calls I got wrong along the way: I read the `governance` failure as transient fork drift when it was a deterministic time-of-day bug in the test, and I overstated the `list.ts` fan-out. Both corrected above and in the thread history.

PR #1304, opened 2026-09-14
fix: reject migrate requests whose source wallet is the service fee payer

bryzettler:
## Why

Both migrate routes (`blockchain-api` `migration.migrate` and `migration-service` `POST /ledger/migrate`) build a batch that moves everything from a caller-supplied source wallet to a caller-supplied destination, then sign every tx as fee payer. Nothing stopped a caller from naming the fee payer itself as the source. The service would then sign a sweep of its own balance to the caller's destination. This is how the migration wallet was drained on 2026-09-13.

## What

- Reject a source equal to the fee payer on both routes.
- Reject a destination equal to the fee payer on both routes. The fee payer also signs the `to` side of `ledgerTransferPositionV0`, so assets parked there could never leave through the route once the source guard bars that wallet.
- `migration-service` gains the `from != to` and on-curve destination checks `blockchain-api` already had. All four checks live in a pure `validateMigrateWallets` with unit tests.
- The `/ledger/migrate` handler moves into `registerLedgerMigrate` with the fee payer and tx builder injected, so a `fastify.inject()` test pins that the guard is actually applied at the call site, not just correct in isolation. `index.ts` starts listening at import time, so the route could not be tested there.
- New CI job runs the `migration-service` unit tests.

## Test changes

- The `blockchain-api` e2e migration suite used the same keypair as source and fee payer, which the guard now rejects. The suite generates a dedicated fee payer.
- Separating the two keys exposed that the split-hotspot flow returns a fee-payer-only tx group (new fanout init + fund + schedule). The signer assertion now pins each tx's signer set to exactly `["source"]` or `[]` and requires both kinds to be present.
- `signAndSubmitTransactionData` only signs txs that list the caller's key as a required signer, and throws if no tx in the batch did. Before this PR it signed unconditionally, so any fully server-signed batch already threw; the new throw keeps that failure visible instead of letting the batch submit silently.

## Rides along

- `welcome-pack.test.ts`: the pack pre-funding stopgap is removed. Its TODO said to remove it once the welcome-pack program with the refund moved after the bubblegum CPI reached mainnet. welcome-pack 0.0.6 and mini-fanout 0.1.7 shipped to mainnet on 2026-09-11, which made that condition true. The claimer's post-claim balance under the deployed programs is 5545080 with or without the stopgap.
- `tests/sus.ts`: the token-account test targeted the HNT ATA of the default pubkey on devnet. That account was created on 2026-09-13, so the idempotent create became a no-op and the rent charge went unobserved. The test now uses a fresh owner.


madninja (commented):
Guard is in the right place in both routes and each one compares against the key that actually signs (`feePayerKeypair` here; `provider.wallet`/`wallet` both derive from `ANCHOR_WALLET` in `migration-service/src/solana.ts`). Two things to sort before this lands, plus inline hardening suggestions.

**1. The e2e migration tests use the fee payer as their source, so 7 of them will now 400.**

`payer = loadKeypairFromEnv()` reads `TEST_WALLET_KEYPAIR_PATH` (`tests/e2e/helpers/wallet.ts:31`), and `blockchain-api-e2e.yml:239,241` point both `TEST_WALLET_KEYPAIR_PATH` and `FEE_PAYER_WALLET_PATH` at the same `testing_key.json`. `migration.test.ts:92` also writes `payer.secretKey` to the fee-payer path directly. So every test calling `migrate` with `sourceWallet: payer.publicKey.toBase58()` (7 occurrences in `tests/e2e/migration.test.ts`) now hits the new `BAD_REQUEST`.

`TEST_WALLET_2_KEYPAIR_PATH` / `loadKeypair2FromEnv()` are already wired in CI, so the cheap fix is to make the migration suite's source wallet key 2 and leave key 1 as the fee payer. I have not run the suite locally (needs surfpool) — this is from the fixture and workflow wiring, so worth confirming against the matrix once `build-idls` clears.

**2. Nothing asserts either guard.**

Worth a regression test in both packages, since the whole value of this change is that it stays. In `blockchain-api`, once (1) is fixed, an e2e case asserting `BAD_REQUEST` for `sourceWallet === feePayer` is a few lines. `migration-service` has no test setup at all today, so its guard is unexercised — at minimum a unit test over the route handler, or lift the predicate into `ledger.ts` where it can be tested without a server. Check by deleting either `if` and confirming something goes red.

**3. Optional, structural.** Both routes sign unconditionally once built (`migrate.ts:1013`; `ledger.ts:227-236` signs whenever the service wallet appears as any signer). The property being protected is "the service never contributes the source authority's signature", and the pubkey comparison covers that for both current routes. An assert at the signing site, rather than at the input, would also cover a future route that takes a different caller-supplied authority. Not needed for this PR.

packages/blockchain-api/src/server/api/routers/migration/procedures/migrate.ts
```diff
@@ -174,6 +174,13 @@ export const migrate = publicProcedure.migration.migrate.handler(
 
     const feePayerKeypair = loadKeypair(env.FEE_PAYER_WALLET_PATH);
     const feePayer = feePayerKeypair.publicKey;
+    // Every returned tx is signed by the fee payer. A source equal to the fee
+    // payer would make it sign a sweep of its own balance to the destination.
+    if (sourcePubkey.equals(feePayer)) {
+      throw errors.BAD_REQUEST({
+        message: "Source wallet must not be the fee payer",
+      });
+    }
```
madninja, line 183: Suggest making this symmetric with the destination. The fee payer also signs the `to` side of `ledgerTransferPositionV0`, so a destination equal to the fee payer parks assets — locked governance positions included — in a wallet that the source guard above then bars as a source, leaving nothing able to move them back out through this route.

```suggestion
    // Every returned tx is signed by the fee payer. A source equal to the fee
    // payer would make it sign a sweep of its own balance to the destination.
    if (sourcePubkey.equals(feePayer)) {
      throw errors.BAD_REQUEST({
        message: "Source wallet must not be the fee payer",
      });
    }
    // The fee payer also signs the `to` side of ledgerTransferPositionV0, so a
    // destination equal to it parks assets in a wallet the guard above bars as
    // a source. Keep the route symmetric.
    if (destPubkey.equals(feePayer)) {
      throw errors.BAD_REQUEST({
        message: "Destination wallet must not be the fee payer",
      });
    }
```

packages/migration-service/src/index.ts
```diff
@@ -283,6 +283,11 @@ server.post<{
   if (request.body.attestation !== ATTESTATION) {
     return reply.code(400).send({ error: "Invalid attestation" });
   }
+  // The service signs every tx it returns as fee payer. With `from` set to
+  // its own wallet it would sign a sweep of its own balance to `to`.
+  if (from.equals(provider.wallet.publicKey)) {
+    return reply.code(400).send({ error: "Invalid source wallet" });
+  }
```
madninja: Same destination symmetry as the `blockchain-api` route, and while here this endpoint is missing that route's two other input checks: `from !== to` and an on-curve destination. `PublicKey` is already imported.

Note the `ATTESTATION` comparison above is a constant in this repo, so it documents intent rather than restricting callers — these input checks are the only validation on the route.

```suggestion
  // The service signs every tx it returns as fee payer. With `from` set to
  // its own wallet it would sign a sweep of its own balance to `to`.
  if (from.equals(provider.wallet.publicKey)) {
    return reply.code(400).send({ error: "Invalid source wallet" });
  }
  // The service also signs the `to` side of ledgerTransferPositionV0, so a
  // destination equal to its own wallet parks assets in a wallet the guard
  // above bars as a source.
  if (to.equals(provider.wallet.publicKey)) {
    return reply.code(400).send({ error: "Invalid destination wallet" });
  }
  if (from.equals(to)) {
    return reply
      .code(400)
      .send({ error: "Source and destination must differ" });
  }
  // Off-curve destinations can never sign, so anything sent there — locked
  // positions included — is unrecoverable.
  if (!PublicKey.isOnCurve(to.toBytes())) {
    return reply.code(400).send({ error: "Destination must be on-curve" });
  }
```

madninja:
Confirmed on the matrix — the `migration` e2e job is red: **8 passing, 8 failing**, every failure `Error: Source wallet must not be the fee payer`.

[Run 34850728009, job `migration`](https://github.com/helium/helium-program-library/actions/runs/34850728009/job/104001760434):

| # | Test | Line |
|---|---|---|
| 1 | migrates SOL to destination | `migration.test.ts:101` |
| 2 | migrates SPL token (USDC) to destination | `:140` |
| 3 | dedupes duplicate mints in tokens[] | `:319` |
| 4 | migrates a VSR governance position to destination | `:359` |
| 5 | warns when a migrated position is delegated and proxied | `:428` |
| 6 | migrates hotspot in welcome pack (closes pack and transfers) | `:531` |
| 7 | migrates hotspot (no split) returns correct transaction data | `:564` |
| 8 | migrates hotspot with split to destination | `:620` |

Correcting the count in my review above: I said 7, from a literal grep for `sourceWallet: payer.publicKey.toBase58()`. It is 8 — one case reaches the same keypair through a local (`sourceWallet: source` at `:306`, feeding the `:319` call), so the grep undercounted.

Every other e2e job in the matrix passed, so this is isolated to the fixture wiring rather than anything behavioural: `payer` and the fee payer are the same keypair under CI (`blockchain-api-e2e.yml:239,241` both point at `testing_key.json`). Pointing the migration suite's source at `loadKeypair2FromEnv()` / `TEST_WALLET_2_KEYPAIR_PATH` should clear all eight, and leaves key 1 as the fee payer so the guard still has a wallet to reject.

madninja (commented):
All four points addressed, CI green across the board — `migration` e2e, the new `Test Migration Service` job, and both welcome-pack jobs all pass. The `validate.ts` extraction is the right shape: each unit test varies one input and asserts the specific message rather than just "was refused", and the off-curve case asserts `isOnCurve === false` before asserting the refusal so it cannot pass vacuously.

Four follow-ups, in order of what I'd want changed. The first two are about the test edits the fix itself made, which is where I'd look hardest.

**1. The split-hotspot signer assertion is now materially weaker than what it replaced.** Detail inline at `migration.test.ts:872`.

**2. `signAndSubmitTransactionData` now skips silently.** Detail inline at `helpers/tx.ts:42`. Taken together with (1), that test would pass with no source signature on any transaction.

**3. The `validateMigrateWallets` call site is unpinned.** Detail inline at `migration-service/src/index.ts:287`.

**4. Three unrelated changes ride along**, and the PR body is empty so nothing records why. The `sus.ts` fresh-owner change is a genuine improvement — the old `PublicKey.default` owner meant the idempotent create could no-op and the rent charge went unobserved — but it has nothing to do with this route. The welcome-pack stopgap removal is the one I'd want a sentence on: the TODO's stated condition for removing it was the reordered welcome-pack program reaching mainnet, and the new hardcoded `5545080` bakes in the refund branch being live. Either say what made that condition true, or split it out. Worth a PR body either way, and an issue link if there is one.

packages/blockchain-api/tests/e2e/migration.test.ts
```diff
@@ -823,15 +862,20 @@ describe("migration", () => {
...
     );
     expect(migrationTxs.length).to.be.greaterThan(0);
     for (const t of migrationTxs) {
       expect(t.metadata?.type).to.equal("migration");
-      expect(t.metadata?.signers).to.deep.equal(["source"]);
+      expect(t.metadata?.signers).to.not.include("destination");
     }
+    expect(
+      migrationTxs.some((t: any) => t.metadata?.signers?.includes("source"))
+    ).to.equal(true);
```
madninja: This drops from an exact per-transaction claim to a tolerant aggregate one. Before, every migration tx had to be exactly `["source"]`. Now the loop only rules out `"destination"`, and a single `some()` covers `"source"` across the whole batch — so a regression that drops the source signer from the cNFT transfer still passes as long as any one transaction (the old-fanout close, say) still requires it.

The loosening is partly justified: with a distinct fee payer the new-fanout group really is fee-payer-only, so `deep.equal(["source"])` cannot hold for all of them. But the fix for "one shape is now different" is to pin the shapes, not to weaken to an existential. Something like:

```suggestion
    expect(migrationTxs.length).to.be.greaterThan(0);
    for (const t of migrationTxs) {
      expect(t.metadata?.type).to.equal("migration");
    }
    // Pin the per-tx signer sets rather than an aggregate: an existential over
    // the batch still passes if a regression drops "source" from the cNFT
    // transfer, which is the signature that matters here.
    const signerSets = migrationTxs
      .map((t: any) => (t.metadata?.signers ?? []).join(","))
      .sort();
    expect(signerSets).to.deep.equal(["", "source", "source"]);
```

Numbers are a guess at the current batch shape — worth printing `signerSets` once and pasting in what it actually is. The property to keep is that the count of source-signed transactions is fixed, so losing one goes red.

packages/blockchain-api/tests/e2e/helpers/tx.ts
```diff
@@ -33,7 +33,15 @@ export async function signAndSubmitTransactionData(
...
-    tx.sign([signer]);
+    // Only sign txs that list this key as a required signer; a server-signed
+    // fee-payer-only tx in the batch would otherwise make web3.js throw.
+    const requiredSigners = tx.message.staticAccountKeys.slice(
+      0,
+      tx.message.header.numRequiredSignatures
+    );
+    if (requiredSigners.some((k) => k.equals(signer.publicKey))) {
+      tx.sign([signer]);
+    }
```
madninja, line 46: Correct for the fee-payer-only transaction this PR introduces into the batch, but it turns a throw into a pass for a case worth catching: a transaction that *should* have required this signer and no longer does is now skipped in silence rather than raising. This helper backs 20+ e2e suites, so that applies well beyond migration.

Cheap fix is to make the skip observable:

```suggestion
    // Only sign txs that list this key as a required signer; a server-signed
    // fee-payer-only tx in the batch would otherwise make web3.js throw.
    const requiredSigners = tx.message.staticAccountKeys.slice(
      0,
      tx.message.header.numRequiredSignatures
    );
    if (requiredSigners.some((k) => k.equals(signer.publicKey))) {
      tx.sign([signer]);
      signedCount++;
    }
```

with a `let signedCount = 0` before the loop and an assert after it that at least one transaction in the batch was signed — otherwise a batch that needs no client signature at all submits happily and the test reads as coverage.

packages/migration-service/src/index.ts
```diff
@@ -283,6 +284,10 @@ server.post<{
   if (request.body.attestation !== ATTESTATION) {
     return reply.code(400).send({ error: "Invalid attestation" });
   }
+  const invalid = validateMigrateWallets(from, to, provider.wallet.publicKey);
+  if (invalid) {
+    return reply.code(400).send({ error: invalid });
+  }
```
madninja: The predicate has five unit tests; this call has none. Deleting the `if (invalid)` body — or changing it to ignore the return — leaves the new `Test Migration Service` job green, because nothing exercises the route itself. That is the usual shape where a decision is extracted, tested, and then discarded at the point of use.

Not necessarily worth a server harness for four lines. But it is the difference between "the rule is correct" and "the rule is applied", so I'd rather it were a recorded choice than an assumption. If a route test is too much, `fastify.inject()` against the registered handler would cover it without a listening socket or an RPC.

packages/blockchain-api/tests/e2e/migration.test.ts
madninja: Replacing my earlier suggestion, which hardcoded a batch shape I had guessed at. This version is strictly stronger than what's here now and needs no counts:

```suggestion
    expect(migrationTxs.length).to.be.greaterThan(0);
    // Pin each tx's signer set rather than an aggregate. An existential over
    // the batch still passes when a regression drops "source" from the cNFT
    // transfer, as long as some other tx in the batch still requires it.
    // Every tx here is either source-signed or fee-payer-only, and both kinds
    // occur, so assert exactly that.
    const signerSets = migrationTxs.map((t: any) =>
      (t.metadata?.signers ?? []).join(",")
    );
    for (const t of migrationTxs) {
      expect(t.metadata?.type).to.equal("migration");
    }
    for (const s of signerSets) {
      expect(s).to.be.oneOf(["source", ""]);
    }
    expect(signerSets).to.include("source");
    expect(signerSets).to.include("");
```

Why it is stronger than `not.include("destination")` plus a `some()`:

- it rules out every unexpected signer set, not just `"destination"` — `"source,destination"`, a new third label, or a tx that requires the destination alone all fail now
- `include("")` pins that the fee-payer-only group still exists, which is the thing that made the old `deep.equal(["source"])` untenable and is currently asserted nowhere
- `include("source")` keeps the existing guarantee

It joins to a string so `oneOf` compares by value; chai's `oneOf` does not deep-compare arrays, which is why this is not written over the arrays directly.

packages/blockchain-api/tests/e2e/helpers/tx.ts
madninja, line 46: Self-contained version of what I suggested last time — the earlier fragment referenced a `signedCount` declared outside the hunk, which would not apply cleanly.

The case worth catching is a batch where *no* transaction required the signer the caller passed in. Today that submits happily and the test reads as coverage.

```suggestion
    // Only sign txs that list this key as a required signer; a server-signed
    // fee-payer-only tx in the batch would otherwise make web3.js throw.
    const requiredSigners = tx.message.staticAccountKeys.slice(
      0,
      tx.message.header.numRequiredSignatures
    );
    if (requiredSigners.some((k) => k.equals(signer.publicKey))) {
      tx.sign([signer]);
      signedAny = true;
    }
```

with `let signedAny = false;` alongside the `signatures` declaration at line 25, and after the loop:

```ts
  if (txData.transactions.length > 0 && !signedAny) {
    throw new Error(
      `signAndSubmitTransactionData: no tx required ${signer.publicKey.toBase58()}; ` +
        "the batch was entirely server-signed, which is usually a regression in " +
        "which signatures the route asks for"
    );
  }
```

One thing to check before taking it: this helper backs 21 e2e files, so it is worth confirming none of them legitimately passes a signer for a fully server-signed batch. If one does, the throw belongs in `migration.test.ts` instead of the shared helper.

madninja (commented):
Second pass on the same head (`3ddd5632`) — no new commits since my last review, so this is me making the previous round's suggestions actually applyable rather than reacting to new work. Three concrete replacements inline, and one correction to something I got wrong last time.

**Correction:** I suggested `fastify.inject()` for the `/ledger/migrate` call site. `server` is a module-level `const` in `migration-service/src/index.ts:72` and is not exported, so that does not work as written — details and a workable version inline.

Nothing here is a blocker. The guard, the extraction and the unit tests are right; these are all about making the surrounding test edits fail when they should.

packages/migration-service/src/index.ts
madninja: Correcting myself: I suggested `fastify.inject()` here, but `server` is declared `const server: FastifyInstance = Fastify({...})` at `index.ts:72` and never exported, so there is nothing for a test to reach.

It is a small change to make it reachable:

```suggestion
  const invalid = validateMigrateWallets(from, to, provider.wallet.publicKey);
  if (invalid) {
    return reply.code(400).send({ error: invalid });
  }
```

(unchanged — the suggestion block is just to anchor the note)

Add `export` to the `server` declaration and a test can do:

```ts
import { server } from "../src/index";

const res = await server.inject({
  method: "POST",
  url: "/ledger/migrate",
  payload: { from: feePayer, to: other, attestation: ATTESTATION },
});
expect(res.statusCode).to.equal(400);
expect(res.json().error).to.equal("Invalid source wallet");
```

Two import-time requirements to be aware of, both satisfiable in the mocha setup without a network: `solana.ts` reads `ANCHOR_WALLET` and will throw if the file is absent, and `index.ts` constructs a `pg` `Pool` — that one is lazy and does not connect until a query, so it is fine. `ATTESTATION` would need exporting too.

Worth it or not is your call: it pins four lines of wiring. But right now deleting the `if (invalid)` body leaves the new `Test Migration Service` job green, so the rule is verified and its application is not.

madninja (approved):
Approving at `3ddd563`. Nothing outstanding is behavioural — every open point from my earlier rounds is test-strength, and none of it needs to gate this.

Verified before approving:

- CI green at this head: 70 passing, 3 skipping (the release jobs, correctly skipped). `migration` e2e, `Test Migration Service`, and both welcome-pack jobs all pass.
- Both guards in `migrate.ts` sit above `createSolanaConnection` and well above the only signing site (`tx.sign([feePayerKeypair])`), so no build path reaches a signature before the checks.
- `migrate.ts` and `getHotspots.ts` are the only procedures in this router, and `migrate` is the only one that signs.
- `validateMigrateWallets` is compared against the key that actually signs on both sides: `feePayerKeypair.publicKey` here, and `provider.wallet`/`wallet` both deriving from `ANCHOR_WALLET` in `migration-service/src/solana.ts`.
- With both guards in place the service wallet can be neither side of `ledgerTransferPositionV0`, which requires `from` and `to` to sign and the approver to be the service key.

Open, non-blocking, happy for these to be a follow-up or dropped on your call:

1. The split-hotspot signer assertion is weaker than the one it replaced (suggestion on `migration.test.ts:871`).
2. `signAndSubmitTransactionData` skips signing silently (suggestion on `helpers/tx.ts:36`).
3. The `validateMigrateWallets` call site is unpinned — deleting the `if (invalid)` body leaves the new job green (note on `migration-service/src/index.ts:287`).
4. A PR body would help, given this changes input validation on a signing route; plus an issue link if there is one, and a line on why the welcome-pack stopgap removal rides along.

Separate from this PR, worth someone asking the same question of the two other routes that load `FEE_PAYER_WALLET_PATH` — `welcomePacks/procedures/claim.ts` and `reward-contract/procedures/claim.ts`. Neither takes this route's "move everything from X to Y" shape, so I would not widen this PR for it, but "which authorities can a caller steer the fee payer's signature toward" has only been asked of `migrate` so far.

packages/migration-service/src/ledgerMigrateRoute.ts
```diff
@@ -0,0 +1,37 @@
...
+  server: FastifyInstance,
+  deps: {
+    feePayer: PublicKey;
+    getMigrateTransactions: (
+      from: PublicKey,
+      to: PublicKey,
+    ) => Promise<VersionedTransaction[]>;
+  },
+) => {
+  server.post<{
```
madninja, line 28: Making `feePayer` a parameter is what let the route become testable, and it is the right shape — but it also means the value the guard compares against is now supplied by the caller rather than read from `provider.wallet`. Worth asking the follow-up: if a caller passed the worst value this parameter admits, what is the guard still worth?

`PublicKey.default` (all zeros) is the realistic bad value — a missing env, a partially-initialised provider, a second registration written from memory. With that passed in, `from.equals(deps.feePayer)` is false for the real service wallet and the self-referential source request is accepted again, with every unit test still green because they inject their own `feePayer` and never exercise the wiring.

Today this is one production call site (`index.ts:277`) passing the correct value, and `feePayer` is required with no default, so it takes an active mistake. Still, one line closes the class:

```suggestion
// Registered separately from index.ts so the route can be exercised with
// fastify.inject() in a unit test; index.ts starts listening at import time.
export const registerLedgerMigrate = (
  server: FastifyInstance,
  deps: {
    feePayer: PublicKey;
    getMigrateTransactions: (
      from: PublicKey,
      to: PublicKey,
    ) => Promise<VersionedTransaction[]>;
  },
) => {
  // The guard below is only as good as the key it compares against, and that
  // key now arrives from the caller. An all-zero default would compare against
  // a key nobody holds and accept the request the guard exists to refuse.
  if (deps.feePayer.equals(PublicKey.default)) {
    throw new Error(
      "registerLedgerMigrate: feePayer must be the service wallet",
    );
  }
  server.post<{
```

It fails at registration rather than per-request, so a misconfigured service will not start instead of serving a route whose check cannot fire. A test for it is two lines:

```ts
it("refuses to register without a real fee payer", () => {
  expect(() =>
    registerLedgerMigrate(Fastify(), {
      feePayer: PublicKey.default,
      getMigrateTransactions: async () => [],
    })
  ).to.throw(/feePayer/);
});
```

madninja (commented):
Both commits look right, and the route test is better than what I asked for — it asserts `calls.length === 0` on each refusal, so it pins that the guard *short-circuits* rather than merely changing the response body, and it carries a positive control. That is the property worth testing.

**I was wrong in my last suggestion and you were right to drop it.** I had `expect(signerSets).to.include("")`, asserting a fee-payer-only tx must exist in the split batch. As your comment at `migration.test.ts:866` says, the batcher packs groups greedily, so whether that group lands in its own tx depends on instruction sizes — my assertion would have tied the test to packing behaviour and broken on any size change. What's left (`oneOf(["source", ""])` per tx plus `include("source")`) is still strictly stronger than the `not.include("destination")` + `some()` it replaced, which was the point.

Also confirms the check I flagged on the `signedAny` throw: it is green across all 21 e2e files that use the helper, so no suite legitimately passes a signer for a fully server-signed batch.

One new finding, inline — it is about the seam my own suggestion asked you to open, so treat it as me cleaning up after myself rather than a new demand.

CI on `c06d552`: 69 pass, 3 skipping, 1 pending (`Test Anchor Contracts (tests/helium-sub-daos.ts)`, unrelated to this diff). `migration` and `Test Migration Service` both pass. My approval stands once that last job lands.

PR #1306, opened 2026-09-14
fix: require the running tuktuk remote task in set_current_rewards_v1

bryzettler:
## Summary

- \`set_current_rewards_v1\`'s \`RemoteTaskTransactionV0\` path now reads the running tuktuk task from remaining accounts and requires it to be a \`RemoteV0\` task whose signer is the selected oracle. Uses the \`tuktuk-program\` crate's exports; a unit test pins the \`run_task_v0\` discriminator and task account index.
- \`set_current_rewards_wrapper_v2\` forwards its remaining accounts to the lazy distributor.
- The distributor oracle passes the task as a read-only remaining account when building the claim transaction.
- \`update-account-circuit-breaker\` accepts a threshold of 0.

## Tests

New \`with a tuktuk remote task\` block in \`tests/lazy-distributor.ts\`: one case runs a signed reward through a real remote task, three cover payloads presented under a compiled task and expect \`InvalidRemoteTask\`.

## Rollout

Deploy the distributor oracle before the lazy-distributor program. The extra remaining account is ignored by the current on-chain code, so the server can go first safely. The program requires it once deployed.


madninja (commented):
Read the program, server and test diffs, and checked the guard against tuktuk `run_task_v0.rs` on `main`.

What holds: the branch now requires the account at `run_task_v0`'s task position to be present in remaining accounts, tuktuk-owned, a `TaskV0`, and `RemoteV0` with `signer` equal to `lazy_distributor.oracles[oracle_index].oracle`. For a `RemoteV0` task tuktuk verifies the ed25519 instruction against that signer, recomputes `hash(task key, queued_at, remaining accounts)` and executes the compiled transaction from the signed message, so once this check passes the `current_rewards` argument came from a message the oracle signed for this task. The routes I tried and could not get through: a compiled task on a caller-owned queue (fails `NotRemoteTaskFromSigner` equivalent here), a `RemoteV0` task on a caller-owned queue with the oracle as signer (needs a fresh oracle signature over that task's own key and `queued_at`, which is `Clock`-set), passing the signed-for task as a remaining account while running a different task (key mismatch, and test 3 pins exactly this), and calling the wrapper or `set_current_rewards_v1` top-level (program id check). The approver relaxation from `len() == 1` to `!is_empty()` keeps `[0] == approver && is_signer`, so no new privilege there.

Rollout claim checked: current on-chain `set_current_rewards_wrapper_v2` forwards only `vec![approver]` and current `set_current_rewards_v1` requires exactly one remaining account, so an extra account handed to the wrapper today is dropped before the CPI. Server first is safe.

Findings:

1. **The wrapper's forwarding is untested, and it is the production path.** All four new tests call `setCurrentRewardsV1` directly with no approver configured. Mainnet claims go `run_task_v0 -> set_current_rewards_wrapper_v2 -> set_current_rewards_v1` with `[approver, task]` as remaining accounts. Reverting the new `remaining_accounts.extend(...)` line to `vec![approver]` leaves this suite green, while every legitimate claim would fail `InvalidRemoteTask` after deploy. A positive test through `setCurrentRewardsWrapperV2` on a lazy distributor with `approver` set would cover the consumption site.

2. **Shape constants duplicate helium/tuktuk#109.** `RUN_TASK_V0_DISCRIMINATOR`, `RUN_TASK_V0_TASK_ACCOUNT` and the pin test are the same as `verify_running_remote_task` in that PR. Fine to land inline now since 0.4.1 is unpublished; worth a follow-up to swap to the helper so the rule lives in one crate.

3. **The ed25519 parse in this branch is now redundant.** With the task check in place, tuktuk has already verified the same instruction against the same signer. Keeping it is harmless; dropping it removes a second copy of the check that can drift. Either way, a one-line comment saying which check is authoritative would help the next reader.

4. **Coverage note on the guard itself.** Tests pin: task absent, task present but compiled, task present but not the running one. Not pinned: `task_info.owner == tuktuk::ID` and the `TaskV0` discriminator. Neither is reachable from a test (the key must equal the running task's, which the runtime resolves to the real account), so this is a record, not a request.

5. Nit: the block comment in the branch narrates the scenario it defends against. The invariant reads more durably as "the running task must be a `RemoteV0` task signed by this oracle; tuktuk binds the signed message to that task and its accounts".

Outside this PR, same as noted on the tuktuk side: a per-transfer cap on the account-windowed breaker, or escrow thresholds sized to a few claims rather than a day, would bound any future single-transaction path, and an alert on a single escrow outflow above the largest legitimate claim would fire on the first transaction instead of when the window fills.

madninja:
Re-read after the update. The wrapper path is now covered: `tests/distributor-oracle.ts` runs `setCurrentRewardsWrapperV2` under a real `RemoteV0` task with the approver-configured lazy distributor and asserts the resulting `current_rewards`, so reverting the wrapper's remaining-account forwarding would fail that test. The branch comment now names tuktuk's check as authoritative. Nothing further from me on the program side.

One CI item: `Check @helium/idls Changeset` fails because the new `InvalidRemoteTask` error variant changes the generated lazy-distributor IDL, and the only changeset on the branch is for `@helium/distributor-oracle`. The check needs a second changeset bumping `@helium/idls` (patch is enough).

PR #1307, opened 2026-09-14
fix(hsd): price delegated positions from the delegated amount

madninja:
## Summary

A delegation is sized from the HNT deposited when it was made (`DelegatedPositionV0::hnt_amount`), and `sub_dao.vehnt_delegated` and each epoch's `vehnt_at_epoch_start` are built from that amount. Every veHNT figure this program derives for a delegated position now computes on that amount through `delegated_position_view`, which pairs the delegated amount with the live lockup and genesis fields:

- `claim_rewards_v1`, `claim_rewards_v0` and `temp_claim_buggy_rewards`: the position's share of an epoch.
- `close_delegation_v0`: the veHNT and fall rate removed from the sub-DAO, matching what delegation added.
- `extend_expiration_ts_v0`: the old and new veHNT info.

`delegate_v0` keeps reading the live position: it writes the snapshot and the sub-DAO totals from the same value in the same instruction. `change_delegation_v0` closes on the old snapshot and re-delegates on the live amount.

A delegated position whose live deposit falls below the delegated amount is paid the share it was counted for until its delegation ends, which is the share the totals were built from.

`delegated_position_view` builds the value without heap allocation (`recent_proposals` is left empty; no veHNT calculation reads it).

## Mainnet state

All 11,960 delegated positions currently have a live deposit equal to their delegated amount, so no total needs correcting at deploy.

## Tests

- Unit test on `delegated_position_view`, pinning the amount, lockup kind, start and end, genesis end and mint config index.
- `tests/helium-sub-daos.ts`: closing a delegation leaves the sub-DAO at exactly the remaining position's veHNT and fall rate, and a claim pays the epoch share computed on the delegated amount, in both cases with the position's live deposit differing from the delegated amount. The claim tests share one proposal-setup helper.

## Deploy

helium-sub-daos 0.2.49. No IDL change.


programs/helium-sub-daos/src/utils.rs
```diff
@@ -32,6 +32,53 @@ impl OrArithError<SignedPreciseNumber> for Option<SignedPreciseNumber> {
...
+    };
+    let delegated_position = DelegatedPositionV0 {
+      hnt_amount: 1_000,
+      ..Default::default()
+    };
+    let view = delegated_position_view(&position, &delegated_position);
+    assert_eq!(view.amount_deposited_native, 1_000);
+    assert_eq!(view.genesis_end, 7);
+    assert_eq!(view.voting_mint_config_idx, 2);
+  }
```
bryzettler, line 89: The test name promises the live lockup is read, but with `..Default::default()` on both inputs a view built from `Lockup::default()` passes too. Setting a non-default lockup and asserting `view.lockup.end_ts` pins it.

```suggestion
  use voter_stake_registry::state::Lockup;

  use super::*;

  #[test]
  fn reads_the_delegated_amount_and_the_live_lockup() {
    let position = PositionV0 {
      amount_deposited_native: 11_000,
      lockup: Lockup {
        start_ts: 3,
        end_ts: 5,
        kind: LockupKind::Cliff,
      },
      genesis_end: 7,
      voting_mint_config_idx: 2,
      ..Default::default()
    };
    let delegated_position = DelegatedPositionV0 {
      hnt_amount: 1_000,
      ..Default::default()
    };
    let view = delegated_position_view(&position, &delegated_position);
    assert_eq!(view.amount_deposited_native, 1_000);
    assert_eq!(view.lockup.end_ts, 5);
    assert_eq!(view.genesis_end, 7);
    assert_eq!(view.voting_mint_config_idx, 2);
  }
```

bryzettler (commented):
Two small follow-ups: pin the lockup in the new unit test, and route the one remaining delegated-claim handler through `delegated_position_view`.

programs/helium-sub-daos/src/utils.rs
```diff
@@ -32,6 +32,53 @@ impl OrArithError<SignedPreciseNumber> for Option<SignedPreciseNumber> {
 
 pub const EPOCH_LENGTH: i64 = 24 * 60 * 60;
 
+/// The position as its delegation counts it. A delegation is sized from the HNT deposited when it
+/// was made (`DelegatedPositionV0::hnt_amount`); `sub_dao.vehnt_delegated` and each epoch's
+/// `vehnt_at_epoch_start` are built from that amount, and every veHNT figure for a delegated
+/// position is computed on it. Lockup and genesis fields are read from the live position.
+/// `recent_proposals` is not an input to any veHNT calculation and is left empty.
+pub fn delegated_position_view(
```
bryzettler, line 40: This doc comment says "every veHNT figure for a delegated position is computed on it", but one delegated-claim path is not converted yet: `programs/helium-sub-daos/src/instructions/delegation/temp_claim_buggy_rewards.rs` line 124 still calls `position.voting_power(...)` on the live position, so it prices the claim off `PositionV0::amount_deposited_native` instead of `DelegatedPositionV0::hnt_amount`. Its handler has the same `position` / `delegated_position` bindings as `claim_rewards_v0.rs`, so the same change applies.

```rust
// temp_claim_buggy_rewards.rs line 15
use crate::{
  current_epoch, delegated_position_view, error::ErrorCode, state::*, ClaimRewardsArgsV0, TESTING,
};

// temp_claim_buggy_rewards.rs lines 124-127
  let delegated_vehnt_at_epoch = delegated_position_view(position, delegated_position)
    .voting_power(
      voting_mint_config,
      ctx.accounts.sub_dao_epoch_info.start_ts(),
    )?;
```

programs/helium-sub-daos/src/utils.rs
madninja: Applied in the follow-up commit.

programs/helium-sub-daos/src/utils.rs
madninja: Applied in the follow-up commit.

bryzettler (approved):
LGTM

PR #1310, opened 2026-09-14
fix(dca): pin the remote task signer and url for DCA tasks

madninja:
A DCA's remote task is signed by `dca_signer` and served from `dca_url`, and the transaction it carries signs as the task queue's custom signers. Both now name the pinned DCA service: `tuktuk-dca`'s `initialize_dca_v0` and its nested variant require `args.dca_signer == DCA_SIGNER` and `args.dca_url == DCA_URL`, and `dc-auto-top`'s `initialize_auto_top_off_v0` applies the same two checks to the values it stores and later hands to `tuktuk-dca`. `slippage_bps_from_oracle` must be below 10000, so a swap always has a price floor. `update_auto_top_off_v0` does not accept either field, so it needs no check.

The url check is exact equality. Both live mainnet accounts store exactly `DCA_URL`, and `packages/tuktuk-dca-service` serves one route, so a sub-path has no consumer.

`DCA_SIGNER` is now `dcauByvWqZMRAhPr7Qoadag4XqVS75JyR5zsvqWPWJS`, read off the two live mainnet `AutoTopOffV0` accounts `EsNPf1Beanrb96PmPqLD7V2RGhNrv4UV4nwJamF6ncr3` and `Bvge1RMm2qfqsG4ijxv3ULNYzSosiaKuu8D2XPNpm2W6` (program `topqqzQZroCyRrgyM5zVq6xkFDVnfF13iixSjajydgU`) via `getAccountInfo` against `https://solana-rpc.web.helium.io`, decoding the `repr(C)` layout at offset 760. Both accounts carry that key and both carry `dca_url` `https://tuktuk-dca.web.helium.io/dca`, which matches the `POST /dca/:dcaKey` route `packages/tuktuk-dca-service` serves. `DCA_URL` gains the `/dca` segment for that reason, on both the mainnet and devnet branches; no devnet account carries the current `AutoTopOffV0` layout, so no stored devnet url constrains that branch.

The guards run at creation and on every HNT refill: `top_off_hnt_v0` CPIs `initialize_dca_nested_v0` with the stored `dca_signer` and `dca_url`, so a stored value that does not match fails the refill run (which then does not reschedule). Only `check_repay_v0` re-queues from stored values without re-checking them. Every live account already stores the pinned values: zero `DcaV0` accounts exist, and both `AutoTopOffV0` accounts match on signer and url. A stored value that does not match can only be changed by closing the account and re-initializing it through `helium-admin setup-dc-auto-topoff` with explicit `--dcaUrl` / `--dcaSigner`.

`dc-auto-top`'s `devnet` feature now propagates to `tuktuk-dca`, so a devnet build of `dc-auto-top` pins the devnet url rather than the mainnet one. Before a devnet deploy, confirm the devnet DCA service signs with `DCA_SIGNER`; a devnet service on a different key would be refused at initialize.

The args are validated rather than removed: `packages/helium-admin-cli` and the test suites pass both fields, so dropping them from `InitializeDcaArgsV0` and `InitializeAutoTopOffArgsV0` would change the instruction shape for every caller, where validating changes only the error set.

Tests: `tests/dc-auto-topoff-bankrun.ts` seeds an `AutoTopOffV0` with an unpinned signer and with a look-alike url and asserts the HNT refill run reverts with `InvalidDcaSigner` and `InvalidDcaUrl`; its existing positive path creates the DCA with the pinned values. `tests/tuktuk-dca.ts` asserts `InvalidDcaSigner`, `InvalidDcaUrl` and `InvalidSlippage` on `initialize_dca_v0`, and `tests/dc-auto-topoff.ts` asserts the first two on `initialize_auto_top_off_v0`, each case against a fresh router key so the guard is the only thing that can reject it. A `TESTING` build pins the signer to a fixed keypair the suites share, so the suites now use one DCA server port.


madninja:
Reviewed the program diff. The pin is enforced once, in `initialize_dca_impl`, so both the direct and nested paths get it, and dc-auto-top reuses the same constant and predicate rather than a second copy. The url predicate handles the look-alike host and the 128-byte field width, and the unit tests cover both. The on-chain signer and `/dca` url were verified against both live top-off accounts, which also caught the url constant missing its path segment.

Two things for the author or a reviewer with devnet context:

1. `DCA_SIGNER` has no devnet branch, so a `--features devnet` build pins the mainnet service key. If the test-helium DCA service signs with a different key, devnet DCA creation fails until a devnet constant is added.
2. The pin applies at creation. Existing `DcaV0` accounts keep the signer and url they were created with, and `check_repay_v0` re-queues from those stored values. Worth enumerating live `DcaV0` accounts once before deploy to confirm they all carry the pinned values.

The two localnet suites (`tuktuk-dca.ts`, `dc-auto-topoff.ts`) carrying six of the new negative cases were not run locally; they rest on the CI matrix. The bankrun suite and the Rust unit tests ran and the signer, url, and boundary guards were each mutation-checked against them.

bryzettler (commented):
Code looks right; one correction to the description.

The PR body says `check_repay_v0` and `top_off_hnt_v0` re-queue from the stored values without re-checking them. `top_off_hnt_v0` CPIs `initialize_dca_nested_v0` with the stored `dca_signer` and `dca_url` (`programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs:374-375`), so tuktuk-dca's new `InvalidDcaSigner` / `InvalidDcaUrl` guards run on every HNT refill. The two bankrun cases in `tests/dc-auto-topoff-bankrun.ts` rely on exactly that: they seed a mismatched account and assert the crank reverts. So a stored value that does not match fails the refill run rather than being re-queued unchecked, and a reverted run does not reschedule (see the comment above the CPI in `top_off_hnt_v0.rs`). Only `check_repay_v0` is guard-free. Worth restating that sentence so the risk paragraph matches the code.

madninja:
Body corrected: the refill path re-runs the guards through the nested CPI; only check_repay_v0 re-queues from stored values unchecked.

.changeset/pin-dca-remote-signer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@helium/idls": patch
+---
+
+Pin the DCA remote task's signer and url. tuktuk-dca and dc-auto-top now reject a `dca_signer` other than the production DCA signer, a `dca_url` that does not address the pinned DCA service, and slippage of a whole 100%. Adds the `InvalidDcaSigner`, `InvalidDcaUrl` and `InvalidSlippage` error codes.
```
bryzettler: The check is `args.dca_url == DCA_URL`, exact equality, so a url that only extends the pinned one with a sub-path is rejected too. "does not address" reads looser than the code is.

```suggestion
Pin the DCA remote task's signer and url. tuktuk-dca and dc-auto-top now reject a `dca_signer` other than the production DCA signer, a `dca_url` other than the pinned DCA service url, and slippage of a whole 100%. Adds the `InvalidDcaSigner`, `InvalidDcaUrl` and `InvalidSlippage` error codes.
```

bryzettler (commented):
Two small things: the changeset describes the url check more loosely than the code, and the rejection assertion is copied into three test files.

tests/dc-auto-topoff.ts
```diff
@@ -312,6 +313,76 @@ describe("dc-auto-topoff", () => {
...
+      [{ dcaSigner: Keypair.generate().publicKey }, "InvalidDcaSigner"],
+      [{ dcaUrl: `${DCA_TEST_URL}.other.example` }, "InvalidDcaUrl"],
+    ] as const) {
+      let caught: any = null;
+      try {
+        await initializeWith(overrides);
+      } catch (e: any) {
+        caught = e;
+      }
+      expect(caught, `expected ${errorName}, got no error`).to.not.be.null;
```
bryzettler: This catch-and-assert block is added three times in the PR: here, as `expectRejection` in tests/tuktuk-dca.ts, and as the tail of `expectDcaRejected` in tests/dc-auto-topoff-bankrun.ts. The bankrun copy has already drifted to `.to.not.equal(null)`. One shared helper at tests/utils/expectAnchorError.ts, beside the existing tests/utils/expectBnAccuracy.ts, would replace all three:

```ts
import { expect } from "chai";

export const expectAnchorError = async (
  promise: Promise<unknown>,
  errorName: string
) => {
  let caught: any = null;
  try {
    await promise;
  } catch (e: any) {
    caught = e;
  }
  expect(caught, `expected ${errorName}, got no error`).to.not.be.null;
  expect(
    `${caught}${JSON.stringify(caught.logs ?? [])}`,
    `expected ${errorName}`
  ).to.include(errorName);
};
```

This loop body then becomes `await expectAnchorError(initializeWith(overrides), errorName);`, the bankrun tail becomes `await expectAnchorError(crank(hntTask), errorName);`, and tests/tuktuk-dca.ts imports it in place of `expectRejection`.

.changeset/pin-dca-remote-signer.md
madninja: Applied in the follow-up commit.

tests/dc-auto-topoff.ts
madninja: Applied in the follow-up commit.

bryzettler (approved):

PR #1311, opened 2026-09-14
fix(vsr): zero marker weight when a marker has no choices

madninja:
A `VoteMarkerV0` caches the voting weight its votes were cast at, so that a
position whose voting power decays keeps a consistent weight for as long as it
holds votes on a proposal. The invariant that makes the cache safe is that a
marker with no choices carries no weight: once the marker holds no votes, the
next vote recomputes the weight from the position.

`VoteMarkerV0::sync_weight` states that rule in one place. It runs after choices
are removed (`relinquish_vote_v1`, `proxied_relinquish_vote_v0`), after the
recount in `count_proxy_vote_v0` replaces the choices from the proxy marker, and
before a choice is added (`vote_v0`, `proxied_vote_v0`), so a marker that holds
no votes takes a fresh weight on its next vote whatever state it was in before.

Tests: two unit tests on `sync_weight` cover the empty and non-empty cases. The
existing v0 and v1 relinquish tests assert the on-chain `weight` is 0 once the
choices are empty. The anchor suite gains `(v1) clears the marker weight when
the recount leaves no choices`, which votes through a proxy, relinquishes,
recounts and asserts the on-chain `weight` field is 0, then advances the
registrar clock, votes again and asserts the new weight is smaller than the
first and matches the weight the proposal recorded.

Mainnet: none of the 25,266 vote markers currently holds a weight with no
choices.


madninja:
Reviewed the diff and the test. The marker rule now lives in one place and every site that can empty choices calls it; the removal CPIs in the two relinquish paths still read the weight before it is cleared, which is the order the tally needs. The anchor test asserts the cause: the marker weight is zero after the recount, and the proposal tally equals the fresh weight after a later vote. Note for the suite: the test moves the registrar time offset and restores it at both ends, so it depends on that shared fixture state.

programs/voter-stake-registry/src/instructions/count_proxy_vote_v0.rs
```diff
@@ -118,6 +118,7 @@ pub fn handler(ctx: Context<CountProxyVoteV0>) -> Result<()> {
   };
   marker.weight = weight;
   marker.choices = ctx.accounts.proxy_marker.choices.clone();
+  marker.sync_weight();
```
bryzettler, line 122: This `sync_weight()` runs after `marker.choices` is replaced, but `weight` was already read from `marker.weight` at line 111. A marker that arrives with `weight > 0` and no choices still passes that old weight to the votes this recount adds, since by line 121 the choices are non-empty again. `vote_v0` and `proxied_vote_v0` sync before they read the weight; doing the same here keeps the read consistent.

Lines 107-118 become:

```rust
  // Calculate voting weight
  let voting_mint_config =
    &ctx.accounts.registrar.voting_mints[usize::from(ctx.accounts.position.voting_mint_config_idx)];

  marker.sync_weight();
  let weight = if marker.weight > 0 {
    marker.weight
  } else {
    ctx.accounts.position.voting_power(
      voting_mint_config,
      ctx.accounts.registrar.clock_unix_timestamp(),
    )?
  };
```

bryzettler (commented):
One finding: `count_proxy_vote_v0` reads the marker weight before the new `sync_weight` runs, so the recount path can reuse a stale weight.

programs/voter-stake-registry/src/instructions/count_proxy_vote_v0.rs
madninja: Applied: the recount now syncs before it reads the cached weight, and again after the choices are replaced.

programs/voter-stake-registry/src/instructions/count_proxy_vote_v0.rs
```diff
@@ -108,6 +108,7 @@ pub fn handler(ctx: Context<CountProxyVoteV0>) -> Result<()> {
   let voting_mint_config =
     &ctx.accounts.registrar.voting_mints[usize::from(ctx.accounts.position.voting_mint_config_idx)];
 
+  marker.sync_weight();
   let weight = if marker.weight > 0 {
```
bryzettler: Key the cache on whether the marker holds choices rather than on a non-zero weight: `sync_weight` keeps the two in step, and a cached weight of 0 is a valid value that must be reused, not recomputed.

```suggestion
  let weight = if !marker.choices.is_empty() {
```

bryzettler (commented):
One finding: the three vote handlers decide whether to reuse the cached marker weight with `marker.weight > 0`, but `sync_weight` now ties the weight to the choice list, so the gate can key on `marker.choices.is_empty()` instead. One suggestion, two plain comments where the edit falls outside the diff hunk, and a doc-comment update.

programs/voter-stake-registry/src/instructions/vote_v0.rs
```diff
@@ -86,6 +86,7 @@ pub fn handler(ctx: Context<VoteV0>, args: VoteArgsV0) -> Result<()> {
     VsrError::MaxChoicesExceeded
   );
 
+  marker.sync_weight();
```
bryzettler: Key the cache on whether the marker holds choices rather than on a non-zero weight: `sync_weight` keeps the two in step, and a cached weight of 0 is a valid value that must be reused, not recomputed. Moving the push below the cache write lets the gate read the choices as they were before this vote, so the pre-push `sync_weight()` is no longer needed.

Replace lines 89-107 with:

```rust
  ctx.accounts.position.num_active_votes += 1;

  let voting_mint_config =
    &ctx.accounts.registrar.voting_mints[usize::from(ctx.accounts.position.voting_mint_config_idx)];

  // Use the original voting weight for this nft until all votes removed
  // This prevents inconsistensies with decaying positions
  let weight = if marker.choices.is_empty() {
    ctx.accounts.position.voting_power(
      voting_mint_config,
      ctx.accounts.registrar.clock_unix_timestamp(),
    )?
  } else {
    marker.weight
  };
  marker.weight = weight;
  marker.choices.push(args.choice);
```

programs/voter-stake-registry/src/instructions/proxied_vote_v0.rs
```diff
@@ -98,6 +98,7 @@ pub fn handler(ctx: Context<ProxiedVoteV0>, args: VoteArgsV0) -> Result<()> {
     VsrError::MaxChoicesExceeded
   );
 
+  marker.sync_weight();
```
bryzettler: Key the cache on whether the marker holds choices rather than on a non-zero weight: `sync_weight` keeps the two in step, and a cached weight of 0 is a valid value that must be reused, not recomputed. Moving the push below the cache write lets the gate read the choices as they were before this vote, so the pre-push `sync_weight()` is no longer needed.

Replace lines 101-119 with:

```rust
  ctx.accounts.position.num_active_votes += 1;

  let voting_mint_config =
    &ctx.accounts.registrar.voting_mints[usize::from(ctx.accounts.position.voting_mint_config_idx)];

  // Use the original voting weight for this nft until all votes removed
  // This prevents inconsistensies with decaying positions
  let weight = if marker.choices.is_empty() {
    ctx.accounts.position.voting_power(
      voting_mint_config,
      ctx.accounts.registrar.clock_unix_timestamp(),
    )?
  } else {
    marker.weight
  };
  marker.weight = weight;
  marker.choices.push(args.choice);
```

programs/voter-stake-registry/src/state/marker.rs
```diff
@@ -20,6 +20,17 @@ pub struct VoteMarkerV0 {
   pub rent_refund: Pubkey,
 }
 
+impl VoteMarkerV0 {
+  /// A marker with no choices carries no weight. Runs before a choice is added
+  /// and after choices are removed, so the cached weight belongs to the votes
+  /// the marker holds.
```
bryzettler: Match the doc comment to where `sync_weight` now runs.

```suggestion
  /// A marker with no choices carries no weight. Runs after choices are
  /// removed, so the cached weight belongs to the votes the marker holds and
  /// a vote on an empty marker takes a fresh weight from the position.
```

programs/voter-stake-registry/src/instructions/count_proxy_vote_v0.rs
madninja: Applied; the anchor suite was re-run locally against the change (23 passing).

programs/voter-stake-registry/src/instructions/vote_v0.rs
madninja: Applied; the anchor suite was re-run locally against the change (23 passing).

programs/voter-stake-registry/src/instructions/proxied_vote_v0.rs
madninja: Applied; the anchor suite was re-run locally against the change (23 passing).

programs/voter-stake-registry/src/state/marker.rs
madninja: Applied; the anchor suite was re-run locally against the change (23 passing).

bryzettler (approved):

PR #1312, opened 2026-09-14
fix(hpl-crons): bind queue_end_epoch to the tracker's task queue

madninja:
`EpochTrackerV0` now records the tuktuk task queue whose tasks advance its epoch, and `queue_end_epoch` requires the `task_queue` account it is passed to equal that field, failing with `InvalidTaskQueue` otherwise. The queue's key seeds the payer PDA and the tasks the return value schedules, so the tracker's own field is what decides which queue that is. `init_epoch_tracker` takes the queue and stores it as a `TaskQueueV0` account; `update_epoch_tracker` gains an optional `task_queue` so the tracker's authority can point an existing tracker at one.

Deploy: the mainnet tracker account is 141 bytes with the trailing 60 unused, so the appended 32-byte field needs no resize and reads back as the default key. `queue_end_epoch` therefore fails with `InvalidTaskQueue` after the upgrade until `update_epoch_tracker` names the live queue. The tracker's `authority` is `hprdnjkbziK8NqhThmAn5Gu4XqrBbctX8du4PfJdgvW`, so that key signs the call.

A queued end-epoch task that keeps failing is closed unexecuted by tuktuk once it is older than the queue's `stale_task_age`, so the cron's task chain does not resume on its own: `helium-admin start-cron` re-bootstraps it, and now also points the tracker at the queue it is bootstrapping when the two differ. The same re-bootstrap is what a later queue change requires, because the in-flight reschedule task names the old queue.

The binding constrains which queue may drive the tracker, not how often: any authority on the named queue can still queue the instruction repeatedly.

Testing: `tests/hpl-crons.ts` adds a case that queues `queue_end_epoch` on a second, freshly created task queue and runs it through tuktuk, asserting the run is rejected with `InvalidTaskQueue` and the tracker's epoch is unchanged. A second case walks the deploy sequence: it points the tracker at the default key, asserts the queued task is rejected with `InvalidTaskQueue`, then names the live queue again and asserts the same task runs and the epoch advances by one. Removing the new constraint and rebuilding fails both cases, since both assert `InvalidTaskQueue`; removing `update_epoch_tracker`'s `task_queue` setter fails only the second, while the rest of the suite stays green. `start-cron` points an existing tracker at the queue it bootstraps when the two differ and no `--epoch` update already does so. The existing positive path and the compute-unit bound are unchanged.


madninja:
Reviewed the program diff. The binding is a single `has_one` on the tracker, the field is set at init and updatable only by the tracker authority, and the negative test asserts the specific error. The `helium-admin start-cron` change keeps the bootstrap path correct by construction.

Deploy notes, verified against mainnet just now:

- Tracker `8SU67drs…` is 141 bytes with zeros at 81..113, so no resize is needed and the new field reads as the default key until set. `queue_end_epoch` fails closed from the moment the upgrade lands until `update_epoch_tracker` names the live queue.
- The tracker's `authority` is `hprdnjkbziK8NqhThmAn5Gu4XqrBbctX8du4PfJdgvW`, a plain key rather than the Squads vault. The follow-up call is therefore a single signed transaction from that key, and should go out immediately after the program upgrade so the end-epoch task chain does not lapse. If it does lapse, `helium-admin start-cron` re-bootstraps it.
- `update_epoch_tracker` takes the queue as a raw pubkey, so a mistyped key stalls the cron rather than being rejected. Acceptable given the operator path above; taking it as a tuktuk-owned account would be the stricter shape if you want it.

packages/helium-admin-cli/src/start-cron.ts
```diff
@@ -98,13 +99,31 @@ export async function run(args: any = process.argv) {
...
         })
         .instruction()
     );
   }
   const epochTrackerAcc = await program.account.epochTrackerV0.fetch(
     epochTracker
   );
+  // queue_end_epoch only runs under the queue the tracker names, so point the tracker at the
+  // queue this bootstrap uses before queueing the task.
+  if (!epochTrackerAcc.taskQueue.equals(taskQueue)) {
```
bryzettler: `fetch` throws on a missing account, so on a fresh tracker the init instruction pushed just above is never sent and this block never runs; and when `--epoch` is passed, the update above already sets `taskQueue`, so this block pushes a second identical instruction. `fetchNullable` plus the two extra guards covers both. Outside this hunk, the `queueTaskV0` description further down has to become `` `queue end epoch ${epochTrackerAcc?.epoch ?? "init"}` `` since the account can now be null.

```suggestion
  const epochTrackerAcc = await program.account.epochTrackerV0.fetchNullable(
    epochTracker
  );
  // queue_end_epoch only runs under the queue the tracker names, so point the tracker at the
  // queue this bootstrap uses before queueing the task. A tracker initialised above already
  // names it, and so does the --epoch update.
  if (
    epochTrackerAcc &&
    !argv.epoch &&
    !epochTrackerAcc.taskQueue.equals(taskQueue)
  ) {
```

programs/hpl-crons/src/state.rs
```diff
@@ -7,6 +7,9 @@ pub struct EpochTrackerV0 {
   pub dao: Pubkey,
   pub epoch: u64,
   pub bump_seed: u8,
+  /// The tuktuk task queue that drives this dao's end-epoch automation. Only a task run by
+  /// this queue may advance `epoch`.
```
bryzettler: `update_epoch_tracker` also writes `epoch` under `authority`, so "only a task run by this queue may advance `epoch`" overstates the guard. The queue check gates `queue_end_epoch`, not every write to the field.

```suggestion
  /// The tuktuk task queue that drives this dao's end-epoch automation. Only a task run by
  /// this queue may run `queue_end_epoch`; `update_epoch_tracker` can still set `epoch`.
```

.changeset/pin-end-epoch-task-queue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@helium/idls": patch
```
bryzettler, line 2: `start-cron` in helium-admin-cli changes behaviour in this PR, so it should get its own changelog entry rather than a silent transitive bump.

```suggestion
"@helium/idls": patch
"@helium/helium-admin-cli": patch
```

tests/hpl-crons.ts
```diff
@@ -225,6 +228,241 @@ describe("hpl-crons", () => {
     );
   });
 
+  // The epoch tracker names the one task queue whose tasks advance its epoch. The payer PDA is
+  // seeded by whichever queue the instruction is passed, so the tracker's own field is what
+  // decides which queue that is.
+  it("rejects queue_end_epoch driven by a task queue the tracker does not name", async () => {
```
bryzettler, line 268: The `compileTransaction([queueEndEpoch().accountsStrict({...nine accounts})], [[Buffer.from("helium"), bump]])` + `queueTaskV0({... freeTasks: 2 ...})` block is now written out three times in this file (the existing heap test and both new cases), and the simulate-then-assert-`InvalidTaskQueue` block twice with the same four-line comment. Two local helpers, `queueEndEpochTask({ queue, taskId, payer, bump })` returning the task key and `simulateRunTask(task)` returning the simulation, would let each case show only what differs: which queue drives the task and what the tracker names.

bryzettler (commented):
4 comments: the start-cron re-point never runs on a fresh tracker and duplicates the --epoch update, the task_queue doc comment overstates the guard, the changeset misses helium-admin-cli, and the three queue-and-run blocks in the test could share two helpers. One note on the description: dropping the has_one fails both new tests, not only the first, since the second also asserts InvalidTaskQueue.

packages/helium-admin-cli/src/start-cron.ts
madninja: Applied in the follow-up commit; the suite was re-run locally against the change (5 passing).

programs/hpl-crons/src/state.rs
madninja: Applied in the follow-up commit; the suite was re-run locally against the change (5 passing).

.changeset/pin-end-epoch-task-queue.md
madninja: Applied in the follow-up commit; the suite was re-run locally against the change (5 passing).

tests/hpl-crons.ts
madninja: Applied in the follow-up commit; the suite was re-run locally against the change (5 passing).

tests/hpl-crons.ts
```diff
@@ -225,6 +262,125 @@ describe("hpl-crons", () => {
...
+
+    const sim = await simulateRunTask(otherTask);
+    expect(sim.value.err, "a task from a queue the tracker does not name must not run").to.not.be
+      .null;
+    expect((sim.value.logs ?? []).join("\n")).to.include("InvalidTaskQueue");
+
+    const epochAfter = (
+      await program.account.epochTrackerV0.fetch(epochTracker)
+    ).epoch;
+    expect(epochAfter.toString()).to.equal(epochBefore.toString());
```
bryzettler: `simulateRunTask` goes through `connection.simulateTransaction`, which never commits state, so the epoch can never differ here and this assertion passes regardless of program behaviour.

```suggestion
    const sim = await simulateRunTask(otherTask);
    expect(sim.value.err, "a task from a queue the tracker does not name must not run").to.not.be
      .null;
    expect((sim.value.logs ?? []).join("\n")).to.include("InvalidTaskQueue");
```

tests/hpl-crons.ts
```diff
@@ -225,6 +262,125 @@ describe("hpl-crons", () => {
...
+          otherName
+        )[0],
+      })
+      .rpc();
+    await tuktukProgram.methods
+      .addQueueAuthorityV0()
+      .accounts({ payer: me, queueAuthority: me, taskQueue: otherQueue })
+      .rpc();
+
+    const [epochTracker] = epochTrackerKey(dao);
```
bryzettler: With the epoch assertions gone, `epochTracker` has no remaining use in this test.

```suggestion
```

bryzettler (commented):
The two new tests assert the epoch is unchanged after a simulated run, which never commits state, so those checks pass regardless of the program; suggestions drop them (the PR body's "epoch is unchanged" clause goes with them), plus the heap-regression comment now heads the helper instead of its test.

tests/hpl-crons.ts
```diff
@@ -225,6 +262,125 @@ describe("hpl-crons", () => {
...
+      await program.account.epochTrackerV0.fetch(epochTracker)
+    ).epoch;
+
+    const sim = await simulateRunTask(task);
+    expect(sim.value.err, "the tracker names no queue, so the task must not run").to.not.be
+      .null;
+    expect((sim.value.logs ?? []).join("\n")).to.include("InvalidTaskQueue");
+    expect(
+      (await program.account.epochTrackerV0.fetch(epochTracker)).epoch.toString()
+    ).to.equal(epochBefore.toString());
```
bryzettler: The preceding call only simulates the transaction, so no state is committed and this comparison holds whatever the program does; the later `epochBefore + 1` assertion after the real run is the one that tests something.

```suggestion
```

tests/hpl-crons.ts
```diff
@@ -118,31 +121,38 @@ describe("hpl-crons", () => {
   // SBF heap ("memory allocation failed, out of memory") while serializing the
   // two compiled return transactions. Mirrors the production flow exactly:
   // start-cron.ts queues the bootstrap task, then a crank turner runs it.
-  it("runs queue_end_epoch through tuktuk without exhausting the heap", async () => {
-    const [customWallet, bump] = customSignerKey(taskQueue, [
+  // Compiles queue_end_epoch under `queue`'s custom "helium" signer, which pays the task
```
bryzettler, line 120: These four lines (120-123) describe the test, not the helper, and now sit above `queueEndEpochTask`; move them back so they sit directly above `it("runs queue_end_epoch through tuktuk without exhausting the heap", ...)` on line 206, leaving the helper with only its own two-line comment.

```ts
  // Regression test for the mainnet outage where queue_end_epoch blew the 32KB
  // SBF heap ("memory allocation failed, out of memory") while serializing the
  // two compiled return transactions. Mirrors the production flow exactly:
  // start-cron.ts queues the bootstrap task, then a crank turner runs it.
```

tests/hpl-crons.ts
madninja: Applied in the follow-up commit.

tests/hpl-crons.ts
madninja: Applied in the follow-up commit.

tests/hpl-crons.ts
madninja: Applied in the follow-up commit.

tests/hpl-crons.ts
madninja: Applied in the follow-up commit.

bryzettler (approved):

PR #1313, opened 2026-09-15
fix(dc-auto-top): run the HNT refill only under the task it scheduled

madninja:
`next_task` and `next_hnt_task` hold task queue addresses, and a task queue address is a slot that anyone else on the queue can fill once the task that held it is gone, whether it was dequeued by `update_auto_top_off_v0` or closed as stale. Until now the top off legs would run under whatever task sat at the recorded address, with the DCA destination and the swap payer taken from that task's accounts.

`AutoTopOffV0` now records when each leg is next due, and a leg refuses to run before its own time. Every run advances the time to the slot after the one it ran, so each leg runs once per slot regardless of what occupies the address. `top_off_hnt_v0` also requires `dca_destination_token_account` to be the top off's own `hnt_account`, and pins `dca_custom_signer` to the task queue's `dca_swap_payer` custom signer, which is the one address the instruction moves lamports to.

Layout and the live accounts: `next_task_time` and `next_hnt_task_time` are appended past the last field, so every existing offset is unchanged and the layout test pins both the old offsets and the new ones. The two live top offs were allocated at 1004 bytes and the struct is now 952 plus its discriminator, so they hold both fields without a realloc. They decode with both times at 0, which is the state a top off has before anything is scheduled: the first run of each leg is unbound the way it is today, and it stamps that leg's real time on its way out. No migration step is needed.

Tests are in `tests/dc-auto-topoff-bankrun.ts`. A second task, queued at the recorded address by another of the queue's authorities and carrying a transaction of that authority's own making, is rejected on the time, on a substituted DCA destination, and on a substituted swap payer, each asserted by error name. The DC leg gets a crank fixture built from the cloned dao, delegated data credits and escrow it parses, and is exercised early, on time, and early again at the address the successful run recorded. Each new guard was removed one at a time and the matching test went red.


bryzettler (commented):
Two doc claims to tighten, and one change with a test: bind the next task each leg records to the free task ids tuktuk passed for the run. Suggestions where the diff allows, plain code blocks where the change sits outside the hunks.

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
```diff
@@ -77,8 +80,14 @@ pub struct TopOffHntV0<'info> {
   /// in the Anchor framework yet, so this is the safe approach.
   #[account(address = IX_ID)]
   pub instruction_sysvar: AccountInfo<'info>,
-  /// CHECK: Custom signer for DCA operations
-  #[account(mut)]
+  /// The task queue's own DCA swap payer: it funds the DCA's rent and is refunded when the DCA
+  /// closes, so it is the one address this instruction moves lamports to.
```
bryzettler: The handler also moves lamports to `task_queue` and `auto_top_off`, so describe what the seeds constraint actually guarantees instead.

```suggestion
  /// The task queue's own DCA swap payer: it funds the DCA's rent and is refunded when the DCA
  /// closes, so it is the only account allowed to pay that rent here.
```

programs/dc-auto-top/src/state.rs
```diff
@@ -38,6 +38,11 @@ pub struct AutoTopOffV0 {
   pub dca_interval_seconds: u64,
   pub dca_input_price_oracle: Pubkey,
   pub dca: Pubkey,
+  // When each leg is next due, as recorded by whatever queued the task sitting in `next_task` /
+  // `next_hnt_task`. A leg refuses to run before its own time, and every run advances the time to
+  // the slot after it, so each leg runs at most once per slot whichever task occupies the address.
```
bryzettler: The DC leg's early return, taken when it cannot pay the crank reward, does not advance the time, so the once-per-slot claim holds only for runs that reschedule.

```suggestion
  // When each leg is next due, as recorded by whatever queued the task sitting in `next_task` /
  // `next_hnt_task`. A leg refuses to run before its own time, and every run that reschedules
  // advances the time to the slot after it, so each leg runs at most once per slot whichever
  // task occupies the address.
```

programs/dc-auto-top/src/errors.rs
```diff
@@ -16,4 +16,7 @@ pub enum ErrorCode {
 
   #[msg("Pyth price not found")]
   PythPriceNotFound,
+
+  #[msg("DCA destination must be the top off's HNT account")]
+  InvalidDcaDestination,
```
bryzettler, line 27: Error code for the free task binding added in both legs below.

```suggestion
  #[msg("DCA destination must be the top off's HNT account")]
  InvalidDcaDestination,

  #[msg("Next task must be a free task the run was given")]
  InvalidFreeTask,
```

programs/dc-auto-top/src/instructions/top_off_dc_v0.rs
```diff
@@ -137,6 +137,14 @@ pub fn handler<'info>(
     ctx.accounts.instruction_sysvar.to_account_info(),
     auto_top_off.next_task,
   )?;
+  // The address in `next_task` is a task queue slot, and a slot is reusable once the task it held
+  // is gone. The recorded time is what makes the leg run on its own schedule: it is due once, and
+  // the reschedule below moves it to the next slot.
+  require_gte!(
```
bryzettler, line 151: Bind the recorded next task to the free task id tuktuk passed for this run.

Four changes in this file.

1. Replace line 94 (the `verify_running_in_tuktuk` signature) with:

```rust
pub fn verify_running_in_tuktuk(
  instruction_sysvar: AccountInfo,
  task_id: Pubkey,
) -> Result<Vec<u16>> {
```

2. Replace lines 127-128 (`  Ok(())` and the closing `}`) with:

```rust
  // tuktuk appends one account per free task id after the accounts the task itself names, and
  // these ids are what those appended accounts are checked against.
  Vec::<u16>::try_from_slice(&current_ix.data[8..])
    .map_err(|_| error!(ErrorCode::InvalidCpiContext))
}

pub fn free_task_key(task_queue: &Pubkey, free_task_ids: &[u16], index: usize) -> Result<Pubkey> {
  let id = free_task_ids
    .get(index)
    .ok_or_else(|| error!(ErrorCode::InvalidFreeTask))?;
  Ok(
    Pubkey::find_program_address(
      &[b"task", task_queue.as_ref(), &id.to_le_bytes()],
      &tuktuk::ID,
    )
    .0,
  )
}
```

3. Replace line 136 (`  verify_running_in_tuktuk(`) with:

```rust
  let free_task_ids = verify_running_in_tuktuk(
```

4. Replace line 153 with:

```rust
  require_keys_eq!(
    ctx.remaining_accounts[0].key(),
    free_task_key(&ctx.accounts.task_queue.key(), &free_task_ids, 0)?,
    ErrorCode::InvalidFreeTask
  );
  auto_top_off.next_task = ctx.remaining_accounts[0].key();
```

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
```diff
@@ -127,7 +136,19 @@ pub fn handler<'info>(
     ctx.accounts.instruction_sysvar.to_account_info(),
     auto_top_off.next_hnt_task,
   )?;
+  // The address in `next_hnt_task` is a task queue slot, and a slot is reusable once the task it
+  // held is gone. The recorded time is what makes the leg run on its own schedule: it is due once,
+  // and the reschedule below moves it to the next slot.
+  require_gte!(
```
bryzettler, line 150: Bind the recorded next task to the free task id tuktuk passed for this run.

Four changes in this file.

1. Replace line 94 (the `verify_running_in_tuktuk` signature) with:

```rust
pub fn verify_running_in_tuktuk(
  instruction_sysvar: AccountInfo,
  task_id: Pubkey,
) -> Result<Vec<u16>> {
```

2. Replace lines 127-128 (`  Ok(())` and the closing `}`) with:

```rust
  // tuktuk appends one account per free task id after the accounts the task itself names, and
  // these ids are what those appended accounts are checked against.
  Vec::<u16>::try_from_slice(&current_ix.data[8..])
    .map_err(|_| error!(ErrorCode::InvalidCpiContext))
}

pub fn free_task_key(task_queue: &Pubkey, free_task_ids: &[u16], index: usize) -> Result<Pubkey> {
  let id = free_task_ids
    .get(index)
    .ok_or_else(|| error!(ErrorCode::InvalidFreeTask))?;
  Ok(
    Pubkey::find_program_address(
      &[b"task", task_queue.as_ref(), &id.to_le_bytes()],
      &tuktuk::ID,
    )
    .0,
  )
}
```

3. Replace line 135 (`  verify_running_in_tuktuk(`) with:

```rust
  let free_task_ids = verify_running_in_tuktuk(
```

4. Replace line 147 with the two checks, resolving the index-1 key before indexing so a short account list returns `InvalidFreeTask` rather than panicking:

```rust
  let task_queue_key = ctx.accounts.task_queue.key();
  let dca_free_task = free_task_key(&task_queue_key, &free_task_ids, 1)?;
  require_keys_eq!(
    ctx.remaining_accounts[0].key(),
    free_task_key(&task_queue_key, &free_task_ids, 0)?,
    ErrorCode::InvalidFreeTask
  );
  require_keys_eq!(
    ctx.remaining_accounts[1].key(),
    dca_free_task,
    ErrorCode::InvalidFreeTask
  );
  auto_top_off.next_hnt_task = ctx.remaining_accounts[0].key();
```

tests/dc-auto-topoff-bankrun.ts
```diff
@@ -361,4 +557,143 @@ describe("dc-auto-topoff under bankrun", () => {
...
+
+    it("does not run before the leg is due", async () => {
+      // Every other account is the one the top off itself would have named, so the time is the
+      // only thing left to reject it on.
+      const { task } = await reoccupy();
+      await expectError("TaskNotDue", crank(task));
+    });
+
+    it("does not send the DCA output anywhere but the top off's HNT account", async () => {
+      const destination = await ataWith(hntMint, Keypair.generate().publicKey, 0n);
```
bryzettler, line 635: Cover the free task binding: a task whose account list puts something other than the run's own free task at `remaining_accounts[0]` is rejected.

1. Replace lines 394-400 to give `queueTopOffAt` an optional extra account:

```ts
    swapPayerSeed = "dca_swap_payer",
    extraAccount,
  }: {
    autoTopOff: PublicKey;
    taskId: number;
    destination?: PublicKey;
    swapPayerSeed?: string;
    extraAccount?: PublicKey;
  }) {
```

2. Replace line 421 (`      .instruction();`) so the override is appended ahead of the free tasks:

```ts
      .remainingAccounts(
        extraAccount
          ? [{ pubkey: extraAccount, isSigner: false, isWritable: false }]
          : []
      )
      .instruction();
```

3. Add the test after line 593, inside this describe block:

```ts

    it("does not record a next task that is not the free task tuktuk was given", async () => {
      // The extra account lands ahead of the free tasks tuktuk appends, so remaining_accounts[0]
      // is a live account rather than the free task this run's own ids name.
      const { autoTopOff, hntTaskId } = await autoTopOffWith(
        50_000_000,
        1_000_000_000
      );
      const { nextHntTaskTime } = await program.account.autoTopOffV0.fetch(
        autoTopOff
      );
      await dequeueBothLegs(autoTopOff);
      const task = await queueTopOffAt({
        autoTopOff,
        taskId: hntTaskId,
        extraAccount: autoTopOff,
      });
      await warpTo(ctx, BigInt(nextHntTaskTime.toString()) + 1n);
      await expectError("InvalidFreeTask", crank(task));
    });
```

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
madninja: Applied: the CHECK now describes what the seeds constraint guarantees.

programs/dc-auto-top/src/state.rs
madninja: Applied: the comment now says every run that reschedules advances the time.

programs/dc-auto-top/src/errors.rs
madninja: Applied: InvalidFreeTask added at the end of the enum, no existing variant renumbered.

programs/dc-auto-top/src/instructions/top_off_dc_v0.rs
madninja: Applied: verify_running_in_tuktuk now returns the free task ids and the DC leg requires remaining_accounts[0] to be the PDA id 0 names. free_task_key lives in schedule_task_v0.rs so both legs share one copy.

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
madninja: Applied: both free task keys are resolved before indexing, so a short id list returns InvalidFreeTask rather than panicking.

tests/dc-auto-topoff-bankrun.ts
madninja: Applied: the test is in place and the suite is at 8 passing. Mutation note: removing either require_keys_eq alone leaves this test green, because the extra account shifts both slots and the other check rejects the same fixture; removing both turns it red.

bryzettler (approved):

PR #1315, opened 2026-09-15
fix(dca-service): sign only for the task and queue the DCA names

madninja:
The DCA swap endpoint read `task`, `task_queue` and `task_queued_at` from the request body and never compared them to the `DcaV0` account it fetched. `task_queue` in particular fed `customSignerKey(taskQueue, ["dca_swap_payer"])`, which derives both the lend destination ATA and the Jupiter `userPublicKey`, so the signed transaction could be bound to a queue the DCA does not name.

The invariant now is that the service signs only for the task and queue the DCA account names: `task` must equal `next_task`, `task_queue` must equal `task_queue`, and `task_queued_at` must equal `queued_at`. A mismatch returns 400 with a short reason before any quote, instruction or signature is produced, and the swap payer is derived from the account's task queue rather than the body. The response shape is unchanged.

The comparison lives in `packages/tuktuk-dca-service/src/binding.ts` as `dcaTaskBindingError`, which the handler and the test double under `tests/utils/dca-test-server.ts` both import, so the anchor suites exercise the code the service runs. The test double's `taskQueue` config field is gone now that the queue comes from the account. `tests/dca-server-binding.ts` covers the shared function directly plus one accepting and three rejecting HTTP requests (wrong task, wrong queue, wrong queued_at); deleting the queue comparison turns the matching rejection test red.

The branch is rebased on a develop that already contains #1310, whose `DCA_TEST_PORT` and pinned remote signer the test double now uses.


packages/tuktuk-dca-service/src/index.ts
```diff
@@ -188,8 +188,27 @@ server.post<{
...
+        .status(400)
+        .send({ error: "task_queue does not match dca task_queue" });
+      return;
+    }
+    if (!taskQueuedAt.eq(dcaAccount.queuedAt)) {
+      reply
+        .status(400)
+        .send({ error: "task_queued_at does not match dca queued_at" });
+      return;
+    }
```
bryzettler: Both this handler and `tests/utils/dca-test-server.ts` carry these checks by hand, and `tests/dca-server-binding.ts` only exercises the copy in the test double, so deleting them here leaves the suite green. Move the comparison into one exported function that both import.

```suggestion
    const bindingError = dcaTaskBindingError(
      { task, taskQueue, taskQueuedAt },
      dcaAccount
    );
    if (bindingError) {
      reply.status(400).send({ error: bindingError });
      return;
    }
```

with `import { dcaTaskBindingError } from "./binding";` alongside the `./env` import, and a new `src/binding.ts`:

```ts
import { BN } from "@coral-xyz/anchor";
import { PublicKey } from "@solana/web3.js";

export interface DcaTaskRequest {
  task: PublicKey;
  taskQueue: PublicKey;
  taskQueuedAt: BN;
}

export interface DcaTaskFields {
  nextTask: PublicKey;
  taskQueue: PublicKey;
  queuedAt: BN;
}

// The signature only covers the task and queue this DCA names, so the
// request has to match the on-chain account before anything is built.
// Returns the rejection reason, or null when the request names the DCA's task.
export const dcaTaskBindingError = (
  request: DcaTaskRequest,
  dca: DcaTaskFields,
): string | null => {
  if (!request.task.equals(dca.nextTask)) {
    return "task does not match dca next_task";
  }
  if (!request.taskQueue.equals(dca.taskQueue)) {
    return "task_queue does not match dca task_queue";
  }
  if (!request.taskQueuedAt.eq(dca.queuedAt)) {
    return "task_queued_at does not match dca queued_at";
  }
  return null;
};
```

tests/utils/dca-test-server.ts
```diff
@@ -86,12 +84,32 @@ export async function createDcaServer(
     try {
       const dca = new PublicKey(request.params.dcaKey);
       const task = new PublicKey(request.body.task);
+      const requestTaskQueue = new PublicKey(request.body.task_queue);
```
bryzettler: The `taskQueue` config field this was disambiguating from is removed in this PR.

```suggestion
      const taskQueue = new PublicKey(request.body.task_queue);
```

tests/utils/dca-test-server.ts
```diff
@@ -86,12 +84,32 @@ export async function createDcaServer(
...
+          .status(400)
+          .send({ error: "task_queue does not match dca task_queue" });
+        return;
+      }
+      if (!taskQueuedAt.eq(dcaAccount.queuedAt)) {
+        reply
+          .status(400)
+          .send({ error: "task_queued_at does not match dca queued_at" });
+        return;
+      }
```
bryzettler: Same checks as the service; import the shared function so the suite pins the code the service runs: `import { dcaTaskBindingError } from "../../packages/tuktuk-dca-service/src/binding";`

```suggestion
      const bindingError = dcaTaskBindingError(
        { task, taskQueue, taskQueuedAt },
        dcaAccount
      );
      if (bindingError) {
        reply.status(400).send({ error: bindingError });
        return;
      }
```

bryzettler (commented):
The new guards are only pinned by the copy in the test double; move them into one shared function so the suite covers the service, and add a direct positive case.

tests/dca-server-binding.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  async function post(payload: Record<string, string>) {
+    return server.inject({
+      method: "POST",
+      url: `/dca/${DCA.toBase58()}`,
+      payload,
+    });
+  }
+
+  it("passes the binding checks when the request names the dca's task and queue", async () => {
```
bryzettler, line 75: This test passes on a 500 from the stub provider, so it proves only that no guard fired. Add a direct positive case on the shared function and say why 500 is the expected outcome here.

```suggestion
  it("accepts a request that names the dca's task and queue", () => {
    expect(
      dcaTaskBindingError(
        { task: NEXT_TASK, taskQueue: TASK_QUEUE, taskQueuedAt: QUEUED_AT },
        { nextTask: NEXT_TASK, taskQueue: TASK_QUEUE, queuedAt: QUEUED_AT },
      ),
    ).to.equal(null);
  });

  // The stub provider has no connection, so a request that clears the guards
  // fails at the balance read with a 500; only a 400 here means a guard fired.
  it("passes the binding checks when the request names the dca's task and queue", async () => {
```

This needs `import { dcaTaskBindingError } from "../packages/tuktuk-dca-service/src/binding";` at the top of the file.

packages/tuktuk-dca-service/src/index.ts
madninja: Done in 4c58485 — the handler now calls dcaTaskBindingError from the new src/binding.ts, so the comparisons live in one place.

tests/utils/dca-test-server.ts
madninja: Done in 4c58485 — renamed back to taskQueue now that the config field is gone.

tests/utils/dca-test-server.ts
madninja: Done in 4c58485 — the test double imports dcaTaskBindingError from packages/tuktuk-dca-service/src/binding, so the suite pins the code the service runs.

tests/dca-server-binding.ts
madninja: Done in 4c58485 — added the direct positive case on the shared function and kept the HTTP one with your comment explaining the 500.

bryzettler (approved):

PR #1317, opened 2026-09-15
fix(dca): price the repay floor in output decimals, bind check_repay to its task, and let a dead DCA close

madninja:
Prices the DCA repay floor in the output mint's minor units, binds `check_repay_v0` to the task it belongs to, and gives a DCA whose task has gone a way to close.

Invariants now enforced:

- A DCA's repayment floor is the oracle price ratio scaled by one signed power of ten that carries both the Pyth exponent difference and the difference in the two mints' decimals: `(input_expo - output_expo) + (output_decimals - input_decimals)`. Before, only the exponent term was applied, so a USDC-in/HNT-out DCA measured its floor against a number 100x too small and any repayment cleared it. The product is computed in u128 and converted back with a checked `try_from`, since a 2,000 USDC order overflows a u64 intermediate. Every step is checked, and the handler has no `unwrap` or `expect` left: the price conversions run after the `> 0` check rather than before, the power of ten is a `checked_pow`, and the slippage subtraction is checked.
- `check_repay_v0` only runs inside the DCA's own tuktuk task. It takes the instruction sysvar and calls the same `verify_running_in_tuktuk` check `lend_v0` uses, against `dca.next_task`. The `is_swapping` check moved into the handler body so the binding is what a call from anywhere else hits first, whatever state the DCA is in.
- `check_repay_v0` reads the task for the next order out of `remaining_accounts` with a bounds check and a typed `MissingNextTask` rather than indexing a slice whose length the crank turner chooses.
- `close_dca_v0` in tuktuk-dca no longer parses the recorded task, so a DCA whose task has been run, dequeued or closed as stale can still be closed; the task is only dequeued when the account still holds one, and only when that task's `queued_at` matches the DCA's, so a task queued at a reused id by something else is left in place. The input still goes to the authority's associated account for the input mint and the lamports to the `rent_refund` recorded at creation, so neither destination is caller-chosen.
- dc-auto-top gains `close_dca_v0`, the only route by which a DCA created by the HNT refill leg can be closed: its authority is the top-off PDA. It requires the top-off's own `authority` as a signer, requires the DCA's authority to be that top-off, and sends the returned input to the top-off's own `dca_mint_account` by `has_one`.

Layout: `input_decimals` and `output_decimals` are the two bytes `DcaV0.reserved` held, so every other field sits where it did. No live DCA accounts exist on mainnet, so nothing needs migrating, and a `state.rs` test pins the size and the offsets around the change.

Admin: `helium-admin close-auto-top-off-dca` closes the DCA an auto top off is currently feeding (or a named one), through the top off's authority or its squads vault.

Tests: the fake DCA server no longer multiplies its repayment by 100 to compensate for the missing decimals term, so it returns fair value and the suites measure the real floor. New cases cover a server returning 99% of fair value against a 50 bps floor (refused), a direct `check_repay_v0` call from outside tuktuk (refused), a close whose token destination is not the authority's account (refused), and, under bankrun, creating a DCA through the refill leg, making its task disappear, and closing it through dc-auto-top with the USDC back in `dca_mint_account` — plus a non-authority signer and a foreign token destination, both refused.


programs/tuktuk-dca/src/instructions/check_repay_v0.rs
```diff
@@ -98,43 +108,49 @@ pub fn handler(ctx: Context<CheckRepayV0>, _args: CheckRepayArgsV0) -> Result<Ru
...
+        .and_then(|expo_diff| expo_diff.checked_add(decimal_diff))
+    })
+    .ok_or(ErrorCode::ArithmeticError)?;
+  let scale_factor = 10_u64
+    .checked_pow(scale.unsigned_abs())
+    .ok_or(ErrorCode::ArithmeticError)?;
   let input_amount = dca.swap_input_amount;
 
   // Calculate expected output based on the input amount and oracle prices
   // We multiply by input price first, then divide by output price to avoid integer truncation
```
bryzettler: Do the floor arithmetic in u128: with scale 2 and an 8-decimal price the u64 product `input_amount * 100 * input_price` overflows at about 1.8e9 input minor units, so a larger order fails every run with ArithmeticError.

```suggestion
  let scale_factor = 10_u128
    .checked_pow(scale.unsigned_abs())
    .ok_or(ErrorCode::ArithmeticError)?;
  let input_amount = dca.swap_input_amount;
  let input_amount_u128 = u128::from(input_amount);
  let input_price = u128::from(input_price_with_conf);
  let output_price = u128::from(output_price_with_conf);

  // Calculate expected output based on the input amount and oracle prices
  // We multiply by input price first, then divide by output price to avoid integer truncation
```

programs/tuktuk-dca/src/instructions/check_repay_v0.rs
```diff
@@ -98,43 +108,49 @@ pub fn handler(ctx: Context<CheckRepayV0>, _args: CheckRepayArgsV0) -> Result<Ru
...
-    Ordering::Equal => input_amount
+  let expected_repayment_amount = if scale > 0 {
+    input_amount
+      .checked_mul(scale_factor)
+      .and_then(|amount| amount.checked_mul(input_price_with_conf))
+      .and_then(|amount| amount.checked_div(output_price_with_conf))
+  } else {
+    // `scale_factor` is 1 when the two exponents and the two decimals cancel out.
+    input_amount
       .checked_mul(input_price_with_conf)
```
bryzettler: Second part of the u128 change above (split only because deleted lines sit between the hunks' added lines).

```suggestion
  let expected_repayment_amount = if scale > 0 {
    input_amount_u128
      .checked_mul(scale_factor)
      .and_then(|amount| amount.checked_mul(input_price))
      .and_then(|amount| amount.checked_div(output_price))
  } else {
    // `scale_factor` is 1 when the two exponents and the two decimals cancel out.
    input_amount_u128
      .checked_mul(input_price)
```

programs/tuktuk-dca/src/instructions/check_repay_v0.rs
```diff
@@ -98,43 +108,49 @@ pub fn handler(ctx: Context<CheckRepayV0>, _args: CheckRepayArgsV0) -> Result<Ru
...
-      .unwrap()
-      .checked_div(output_price_with_conf)
-      .unwrap(),
-  };
+      .and_then(|amount| amount.checked_div(output_price_with_conf))
+      .and_then(|amount| amount.checked_div(scale_factor))
+  }
+  .ok_or(ErrorCode::ArithmeticError)?;
```
bryzettler: Third part of the u128 change above.

```suggestion
      .and_then(|amount| amount.checked_div(output_price))
      .and_then(|amount| amount.checked_div(scale_factor))
  }
  .ok_or(ErrorCode::ArithmeticError)?;
  let expected_repayment_amount =
    u64::try_from(expected_repayment_amount).map_err(|_| ErrorCode::ArithmeticError)?;
```

programs/tuktuk-dca/src/instructions/check_repay_v0.rs
```diff
@@ -54,11 +53,22 @@ pub struct CheckRepayV0<'info> {
...
+  /// the supplied Sysvar could be anything else.
+  #[account(address = IX_ID)]
+  pub instruction_sysvar: AccountInfo<'info>,
 }
 
 pub fn handler(ctx: Context<CheckRepayV0>, _args: CheckRepayArgsV0) -> Result<RunTaskReturnV0> {
   let dca_key = ctx.accounts.dca.key();
   let mut dca = ctx.accounts.dca.load_mut()?;
+  // The binding comes before the state check so that a call from outside the DCA's own task is
+  // refused whatever state the DCA is in.
```
bryzettler: Account constraints (has_one, `next_task: Account<TaskV0>`, `queued_at`) still run before the handler, so this ordering only changes which error an outside call sees, not whether the binding is hit first.

```suggestion
  // The binding comes before the state check so that a call from outside the DCA's own task is
  // refused as InvalidCpiContext rather than LendNotCalled.
```

programs/tuktuk-dca/src/instructions/close_dca_v0.rs
```diff
@@ -27,6 +28,7 @@ pub struct CloseDcaV0<'info> {
     has_one = rent_refund,
     has_one = input_mint,
     has_one = input_account,
+    constraint = dca.load()?.is_swapping == 0 @ ErrorCode::SwapInProgress,
```
bryzettler: The DCA authority's signature is never available inside a tuktuk run, so nothing can reach close between lend_v0 and check_repay_v0; a DCA left with `is_swapping = 1` by a run that never reset it would become uncloseable, and close is the only way to recover its input.

```suggestion
```

programs/tuktuk-dca/src/instructions/close_dca_v0.rs
```diff
@@ -12,7 +13,7 @@ use tuktuk_program::{
   TaskQueueAuthorityV0, TaskV0,
 };
 
-use crate::{dca_seeds, queue_authority_seeds, state::*};
+use crate::{dca_seeds, errors::ErrorCode, queue_authority_seeds, state::*};
```
bryzettler: Unused once the close constraint is removed.

```suggestion
use crate::{dca_seeds, queue_authority_seeds, state::*};
```

programs/tuktuk-dca/src/errors.rs
```diff
@@ -28,4 +28,10 @@ pub enum ErrorCode {
 
   #[msg("Slippage from oracle must be less than 100%")]
   InvalidSlippage,
+
+  #[msg("The run must supply the task the next order is queued into")]
+  MissingNextTask,
+
+  #[msg("A swap is in progress")]
+  SwapInProgress,
```
bryzettler: Unused once the close constraint is removed.

```suggestion
```

bryzettler (commented):
Three fixes and three wording changes, all as suggestions. Main points: the new `is_swapping == 0` guard on `close_dca_v0` would leave a DCA that never reached `check_repay_v0` uncloseable, so the suggestion removes it (and the PR body sentence about refusing mid-swap would go with it); the dequeue in close should be bound to the task's `queued_at` like lend and check_repay are; and the `scale > 0` branch of the repay floor overflows u64 for larger orders, so the arithmetic moves to u128. One question: dc-auto-top goes 0.0.14 → 0.0.16 with no 0.0.15 in git; intentional?

programs/dc-auto-top/src/instructions/close_dca_v0.rs
```diff
@@ -0,0 +1,93 @@
...
+    mut,
+    constraint = dca.load()?.authority == auto_top_off.key() @ ErrorCode::InvalidDcaAuthority,
+  )]
+  pub dca: AccountLoader<'info, DcaV0>,
+  pub dca_mint: Box<Account<'info, Mint>>,
+  /// The DCA's own input account, drained into `dca_mint_account` by the CPI below.
+  #[account(mut)]
+  pub dca_input_account: Box<Account<'info, TokenAccount>>,
+  /// The top-off's own input token account: the DCA is its authority's associated account for
+  /// the input mint, which is the only destination tuktuk-dca's close accepts.
```
bryzettler: The sentence reads as if the DCA were the token account.

```suggestion
  /// The top-off's own input token account. tuktuk-dca's close only pays out to the authority's
  /// associated account for the input mint, and this is that account for the top-off.
```

tests/tuktuk-dca.ts
```diff
@@ -543,5 +553,75 @@ describe("tuktuk-dca", () => {
...
+
+      after(() => {
+        slippageBps = 0;
+        setDcaServerRepayBps(10000);
+      });
+
+      it("refuses the repayment", async () => {
+        await new Promise((resolve) => setTimeout(resolve, 2000));
+        // Sent through the provider rather than runAllTasks, which reports a failed run
+        // without the logs the error name is read from.
```
bryzettler: Duplicated comment; the second version says more.

```suggestion
```

.changeset/dca-repay-floor.md
```diff
@@ -0,0 +1,6 @@
+---
+"@helium/idls": patch
+"@helium/helium-admin-cli": patch
+---
+
+Price the DCA repay floor in output minor units. `DcaV0` now records `input_decimals` and `output_decimals` (taken from the two reserved bytes), and `check_repay_v0` scales the oracle price ratio by the mint decimal difference as well as the Pyth exponent difference. `check_repay_v0` also requires that it is running inside the DCA's own tuktuk task. `close_dca_v0` no longer parses the recorded task, so a DCA whose task is gone can still be closed, and it refuses while a swap is in flight. dc-auto-top gains `close_dca_v0` for the DCAs its HNT refill leg creates, reachable from the new `close-auto-top-off-dca` admin command. Adds the `MissingNextTask` and `SwapInProgress` error codes to tuktuk-dca and `InvalidDcaAuthority` to dc-auto-top.
```
bryzettler: Matches the close constraint removal.

```suggestion
Price the DCA repay floor in output minor units. `DcaV0` now records `input_decimals` and `output_decimals` (taken from the two reserved bytes), and `check_repay_v0` scales the oracle price ratio by the mint decimal difference as well as the Pyth exponent difference. `check_repay_v0` also requires that it is running inside the DCA's own tuktuk task. `close_dca_v0` no longer parses the recorded task, so a DCA whose task is gone can still be closed. dc-auto-top gains `close_dca_v0` for the DCAs its HNT refill leg creates, reachable from the new `close-auto-top-off-dca` admin command. Adds the `MissingNextTask` error code to tuktuk-dca and `InvalidDcaAuthority` to dc-auto-top.
```

programs/tuktuk-dca/src/instructions/close_dca_v0.rs
```diff
@@ -108,15 +112,20 @@ pub fn handler(ctx: Context<CloseDcaV0>) -> Result<()> {
...
 
-  // Only dequeue the task if it's not pointing to itself (which means no task scheduled)
-  if ctx.accounts.next_task.key() != ctx.accounts.dca.key() {
+  // The task is only dequeued while one is actually there: the account points at the DCA itself
+  // when nothing was scheduled, and it is empty once the task has been run, dequeued or closed
+  // as stale.
+  if ctx.accounts.next_task.key() != ctx.accounts.dca.key()
+    && !ctx.accounts.next_task.data_is_empty()
+  {
+    let next_task = try_from!(Account<TaskV0>, ctx.accounts.next_task)?;
```
bryzettler, line 120: Bind the dequeue to the task this DCA queued, as lend_v0 and check_repay_v0 do: a task id can be reused after the original is gone, and `try_from!` accepts any live TaskV0 at the recorded address.

```suggestion
    let next_task = try_from!(Account<TaskV0>, ctx.accounts.next_task)?;
    // A task id can be reused once the original is gone, so only a task queued for this DCA is
    // dequeued.
    if next_task.queued_at != ctx.accounts.dca.load()?.queued_at {
      return Ok(());
    }
```

programs/tuktuk-dca/src/instructions/check_repay_v0.rs
madninja: Applied in 0a933e7f9: scale factor, input amount and both prices are u128 now.

programs/tuktuk-dca/src/instructions/check_repay_v0.rs
madninja: Applied.

programs/tuktuk-dca/src/instructions/check_repay_v0.rs
madninja: Applied, with the checked u64::try_from back. tests/tuktuk-dca.ts now runs a single 2,000 USDC order end to end, which fails with ArithmeticError on the u64 version.

programs/tuktuk-dca/src/instructions/check_repay_v0.rs
madninja: Reworded as suggested.

programs/tuktuk-dca/src/instructions/close_dca_v0.rs
madninja: Removed; you're right that it only made an abandoned mid-swap DCA unrecoverable.

programs/tuktuk-dca/src/instructions/close_dca_v0.rs
madninja: Removed.

programs/tuktuk-dca/src/errors.rs
madninja: Removed. It was the last variant on this branch, so no error number moves.

programs/dc-auto-top/src/instructions/close_dca_v0.rs
madninja: Applied.

tests/tuktuk-dca.ts
madninja: Removed the first one.

.changeset/dca-repay-floor.md
madninja: Taken as written.

programs/tuktuk-dca/src/instructions/close_dca_v0.rs
madninja: Applied. dc-auto-topoff-bankrun.ts now has a case where the recorded address holds a live task with a different queued_at, and asserts close leaves it queued.

madninja:
On the version question: 0.0.15 is taken by #1318 (dc-auto-top refill), which lands first, so this branch is 0.0.16 by design. There is no 0.0.15 in git yet because that PR is still open.

bryzettler (approved):

PR #1318, opened 2026-09-15
fix(dc-auto-top): buy what the USDC covers within one slot, clear dequeued tasks, guard the DCA mint and oracle

madninja:
The HNT refill sizes its DCA from the whole gap to `hnt_threshold` and then buys `min(gap, dca_mint balance / swap amount, ceil(seconds to the next slot / order interval))` orders. The slot term is a ceiling because the first order fires at once and the rest one interval apart, so N orders finish `(N-1) * interval` after the run and the largest N that lands before the next run is the ceiling; one order therefore always fits. Two invariants come out of that. A run that cannot pay for the whole gap buys the prefix it can pay for instead of skipping the refill for the day, and no DCA carries an order that is still due when the next run opens its own. Three bankrun tests: a gap ten orders wide with USDC for three produces a DCA of three orders and moves exactly three swap amounts; the same gap against a slot 1500 seconds away at a 300 second interval produces five; and at a 7200 second interval, wider than the slot, it still produces the one order that fires now. A zero order interval is refused at initialize and at update, since it is the divisor in that bound.

Both legs now price HNT against the `HNT_PRICE_ORACLE` constant. The DC leg's compiled task already named it, and `data_credits::mint_data_credits_v0` pins it again off its own state, so `TopOffDcV0` drops `has_one = hnt_price_oracle`. The HNT leg compiled its task against the mutable field instead; `get_task_ix_hnt` now names the constant, `TopOffHntV0` drops the `has_one` and carries `address = HNT_PRICE_ORACLE` on the account, and `UpdateAutoTopOffArgsV0` and the admin CLI's update path drop the field, so a value written at init can no longer decide what either leg reads. The state field stays, unwritten by update, because the layout is frozen. Note that the constant is not devnet-gated; that is pre-existing and now applies to both legs.

`update_auto_top_off_v0` leaves `next_task` and `next_hnt_task` naming the top off itself once it has dequeued them. That is how `schedule_task_v0`, `close_auto_top_off_v0` and both legs already spell "nothing scheduled", so an update no longer leaves either field pointing at a closed PDA that a later task can reoccupy. Tested by asserting both fields after an update and then scheduling again in the same test. The admin CLI passes the same sentinel to the `schedule_task_v0` it builds in that transaction, and reads a leg's rent refund only when a task is actually scheduled, since the sentinel is not a task account and would not decode as one.

The instruction also takes `dca_mint` and `dca_mint_account` as an optional pair and writes them only when both are supplied, so an update that does not carry them leaves the leg pointed where it is rather than repointing it at an empty ATA. A change of mint additionally requires the account the leg spends from today, which must hold nothing, so the balance in it cannot be stranded. Five tests cover the instruction's refusals and its one permitted change: a mint without its account, an account the top off does not own, a zero interval, a mint change while the old account holds a balance, and the same change once that account is empty.

Also: both legs' `verify_running_in_tuktuk` bounds the account list against the index it reads rather than against emptiness, and the HNT leg moves the returned DCA tasks instead of cloning them.


bryzettler (commented):
Pin the HNT leg to the same oracle constant as the DC leg, count the orders that fit a slot with a ceiling, add tests for the new update guards, and skip the task fetch in the CLI when a leg is unscheduled.

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
```diff
@@ -147,11 +144,8 @@ pub fn handler<'info>(
   // The address in `next_hnt_task` is a task queue slot, and a slot is reusable once the task it
   // held is gone. The recorded time is what makes the leg run on its own schedule: it is due once,
   // and the reschedule below moves it to the next slot.
-  require_gte!(
-    Clock::get()?.unix_timestamp,
-    auto_top_off.next_hnt_task_time,
-    ErrorCode::TaskNotDue
-  );
+  let now = Clock::get()?.unix_timestamp;
+  require_gte!(now, auto_top_off.next_hnt_task_time, ErrorCode::TaskNotDue);
```
bryzettler, line 150: `get_task_ix_hnt` still compiles the HNT leg against the stored `hnt_price_oracle` field while the DC leg is already pinned to the `HNT_PRICE_ORACLE` constant, and this PR drops the only way to update that stored field, so a wrong value written at init is permanent for this leg. Pin the HNT leg to the same constant. The changes belong outside this hunk:

```rust
// schedule_task_v0.rs line 23: make the constant visible to the crate (value unchanged)
pub const HNT_PRICE_ORACLE: Pubkey = pubkey!("...");

// schedule_task_v0.rs line 153, in get_task_ix_hnt
hnt_price_oracle: HNT_PRICE_ORACLE,

// top_off_hnt_v0.rs: import it
use super::schedule_task_v0::HNT_PRICE_ORACLE;

// top_off_hnt_v0.rs line 35: drop this line from the auto_top_off constraints
//   has_one = hnt_price_oracle,

// top_off_hnt_v0.rs lines 59-62: pin the account itself instead
#[account(
  address = HNT_PRICE_ORACLE,
  constraint = hnt_price_oracle.verification_level == VerificationLevel::Full @ ErrorCode::PythPriceNotFound,
)]
pub hnt_price_oracle: Account<'info, PriceUpdateV2>,
```

`initialize_auto_top_off_v0.rs` and the state struct stay as they are.

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
```diff
@@ -344,34 +337,40 @@ pub fn handler<'info>(
...
+      let affordable = ctx
+        .accounts
+        .dca_mint_account
+        .amount
+        .checked_div(swap_amount_per_order)
+        .unwrap_or_default();
+      let slot_seconds = u64::try_from(next_time.saturating_sub(now)).unwrap_or_default();
+      let within_slot = slot_seconds
+        .checked_div(interval_seconds)
+        .unwrap_or_default();
```
bryzettler: The first order fires at `now` and the rest every `interval_seconds`, so N orders finish at `(N-1) * interval < slot` and the largest N that fits is the ceiling, not the floor: the floor undercounts by one whenever the slot is not a whole number of intervals, and returns 0 whenever the interval is longer than the slot, which silently stops the leg from buying at all.

```suggestion
      // so the balance is the hard cap; the orders drain one interval apart, so a count whose
      // last order lands after the next run would still be running when that run opens its own
      // DCA.
      let affordable = ctx
        .accounts
        .dca_mint_account
        .amount
        .checked_div(swap_amount_per_order)
        .unwrap_or_default();
      let slot_seconds = u64::try_from(next_time.saturating_sub(now)).unwrap_or_default();
      // The first order fires now and the rest every interval, so the orders that finish
      // before the next run are the ceiling of the ratio, and one always fits.
      let within_slot = if interval_seconds == 0 {
        0
      } else {
        slot_seconds.div_ceil(interval_seconds)
      };
```

tests/dc-auto-topoff-bankrun.ts
```diff
@@ -665,6 +737,201 @@ describe("dc-auto-topoff under bankrun", () => {
...
+        BigInt(task.trigger.timestamp![0].toString()) + 3600n - 1500n,
+      );
+      await crank(hntTask);
+
+      const dca = await dcaAt(autoTopOff, 0);
+      expect(dca.initialNumOrders).to.equal(
+        5,
+        "1500 seconds at 300 per order is five orders",
+      );
+    });
```
bryzettler, line 815: A case where the interval outlasts the slot pins the ceiling: the existing 1500/300 case reads the same under floor and ceiling, so it does not cover the count the change is about.

```suggestion
    });

    it("still buys the order that fires now when the interval outlasts the slot", async () => {
      // Hourly again, but a 7200 second order interval is wider than the 1500 seconds left in
      // the slot, so only the order that fires straight away comes due before the next run.
      const { autoTopOff, hntTask } = await gapOf(
        10,
        {
          schedule: padded("0 0 * * * *", 128),
          dcaIntervalSeconds: new anchor.BN(7200),
        },
        BigInt(swapAmount.muln(10).toString()),
      );
      const task = await tuktukProgram.account.taskV0.fetch(hntTask);
      await warpTo(
        ctx,
        BigInt(task.trigger.timestamp![0].toString()) + 3600n - 1500n,
      );
      await crank(hntTask);

      const dca = await dcaAt(autoTopOff, 0);
      expect(dca.initialNumOrders).to.equal(
        1,
        "the first order fires now, so one order always fits the slot",
      );
    });
```

tests/dc-auto-topoff-bankrun.ts
```diff
@@ -665,6 +737,201 @@ describe("dc-auto-topoff under bankrun", () => {
...
+      );
+      const newMint = await createMint(6);
+      await changeMintTo(autoTopOff, newMint);
+
+      const after = await program.account.autoTopOffV0.fetch(autoTopOff);
+      expect(after.dcaMint.toBase58()).to.equal(newMint.toBase58());
+      expect(after.dcaMintAccount.toBase58()).to.equal(
+        getAssociatedTokenAddressSync(newMint, autoTopOff, true).toBase58(),
+      );
+    });
```
bryzettler, line 1039: The two new guards in `update_auto_top_off_v0` and the associated-token constraint on `dca_mint_account` have no test asserting the error they raise.

```suggestion
    });

    /** The update, with every argument and optional account the caller chooses. */
    async function update(
      autoTopOff: PublicKey,
      args: Record<string, unknown>,
      accounts: Record<string, unknown>,
    ) {
      const state = await program.account.autoTopOffV0.fetch(autoTopOff);
      await program.methods
        .updateAutoTopOffV0({
          schedule: null,
          threshold: null,
          hntThreshold: null,
          dcaSwapAmount: null,
          dcaIntervalSeconds: null,
          dcaInputPriceOracle: null,
          ...args,
        } as any)
        .accounts({
          authority: me,
          payer: me,
          autoTopOff,
          taskQueue,
          nextTask: state.nextTask,
          nextHntTask: state.nextHntTask,
          taskRentRefund: me,
          hntTaskRentRefund: me,
          dcaMint: null,
          dcaMintAccount: null,
          currentDcaMintAccount: null,
          ...accounts,
        })
        .rpc();
    }

    it("refuses an interval of zero, which would come due never", async () => {
      const { autoTopOff } = await autoTopOffWith(50_000_000, 1_000_000_000, 0n);
      await expectError(
        "InvalidDcaInterval",
        update(autoTopOff, { dcaIntervalSeconds: new anchor.BN(0) }, {}),
      );
    });

    it("refuses a mint without the account it is spent from", async () => {
      const { autoTopOff } = await autoTopOffWith(50_000_000, 1_000_000_000, 0n);
      const state = await program.account.autoTopOffV0.fetch(autoTopOff);
      await expectError(
        "IncompleteDcaMintChange",
        update(
          autoTopOff,
          {},
          {
            dcaMint: await createMint(6),
            dcaMintAccount: null,
            currentDcaMintAccount: state.dcaMintAccount,
          },
        ),
      );
    });

    it("refuses an account the top off does not own", async () => {
      const { autoTopOff } = await autoTopOffWith(50_000_000, 1_000_000_000, 0n);
      const state = await program.account.autoTopOffV0.fetch(autoTopOff);
      const newMint = await createMint(6);
      await expectError(
        "ConstraintTokenOwner",
        update(
          autoTopOff,
          {},
          {
            dcaMint: newMint,
            dcaMintAccount: await ataWith(newMint, me, 0n),
            currentDcaMintAccount: state.dcaMintAccount,
          },
        ),
      );
    });
```

packages/helium-admin-cli/src/setup-dc-auto-topoff.ts
```diff
@@ -396,9 +393,19 @@ export async function run(args: any = process.argv) {
         taskRentRefund,
         hntTaskRentRefund,
         authority: autoTopOffAcc.authority,
-        dcaMint: argv.dcaMint
-          ? new PublicKey(argv.dcaMint)
-          : autoTopOffAcc.dcaMint,
+        // The leg keeps the mint it has unless this run is changing it, and a change carries
+        // the account it is spent from today so the program can see that it holds nothing.
+        dcaMint: argv.dcaMint ? new PublicKey(argv.dcaMint) : null,
```
bryzettler, line 398: When a leg is unscheduled the stored task key is the auto top-off's own address, and `tuktukProgram.account.taskV0.fetchNullable` on it hits an account with a different discriminator, so Anchor's decoder throws before the transaction is built. Replace lines 360-368 with a helper that short-circuits on the sentinel:

```ts
    // A leg with nothing scheduled stores the auto top off's own address, which is not a task
    // account, so fetching it there would fail to decode.
    const rentRefundFor = async (task: PublicKey) =>
      task.equals(autoTopOff!)
        ? authority
        : (await tuktukProgram.account.taskV0.fetchNullable(task))?.rentRefund ||
          authority;
    const taskRentRefund = await rentRefundFor(autoTopOffAcc.nextTask);
    const hntTaskRentRefund = await rentRefundFor(autoTopOffAcc.nextHntTask);
```

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
madninja: Applied as sketched: the constant is now pub, get_task_ix_hnt names it, TopOffHntV0 drops the has_one and carries address = HNT_PRICE_ORACLE; state struct and initialize unchanged.

programs/dc-auto-top/src/instructions/top_off_hnt_v0.rs
madninja: Applied: within_slot is now slot_seconds.div_ceil(interval_seconds) behind the zero-interval guard, and the bound in the PR body says ceil.

tests/dc-auto-topoff-bankrun.ts
madninja: Added, expecting 1 order at a 7200 second interval against the 1500 seconds left; reverting div_ceil to / makes exactly this test fail.

tests/dc-auto-topoff-bankrun.ts
madninja: Added the update helper and all three tests; each fails when its guard is removed, and ConstraintTokenOwner is raised by the associated-token constraint as expected.

packages/helium-admin-cli/src/setup-dc-auto-topoff.ts
madninja: Applied: rentRefundFor short-circuits on the sentinel for both legs.

bryzettler (approved):

